PR #39, opened 2023-11-20
Introduce Azure OpenAi extension

geoand:
This extension brings the same friction-less usage of Langchain4j but with Azure AI services.

For this extension to work, 3 properties need to be set:

```properties
quarkus.langchain4j.azure-openai.api-key=
quarkus.langchain4j.azure-openai.resource-name=
quarkus.langchain4j.azure-openai.deployment-id=
```

cc @edeandrea @agoncal


cescoffier (approved):

edeandrea:
Is the azure open ai integration a completely separate extension from the open ai one? What if they are both on the class path at the same time?

For example, the narration service can connect to one or the other (but not both at the same time) via configuration (each integration has its own separate `enabled` property.

Will that be a problem with these 2 extensions coexisting at the same time but with only one being active?

geoand:
> Is the azure open ai integration a completely separate extension from the open ai one? What if they are both on the class path at the same time?

You will get an error.

For the time being, we want things to be as simple as possible - if real use cases arise that necessitate using multiple "providers", we can reconsider the design.

geoand:
Nothing stops you from just picking one :)

edeandrea:
What we have currently is a single flexible library that allows you to connect to any openai provider simply via configuration. That is the current implementation today and unfortunately I can't replace it with something that doesn't provide similar functionality.

geoand:
Your call

edeandrea:
Seriously? That's your answer? You wouldn't think it would be a valid requirement to be able to connect to any openai provider? I think that is short sighted.

I'm sorry but for now I will have to stay with the SK library.

Another question - are these properties build time fixed or overridable at runtime?

geoand:
> Seriously? That's your answer?

Yes it is. 

The reason is very simple, this is something super new and since no one really has the proper patterns figured out (your example sounds totally contrived), I don't want to provide more capabilities at the expense of usability.

> are these properties build time fixed or overridable at runtime?

All the properties are overridable at runtime.

edeandrea:
> (your example sounds totally contrived)

We can agree to disagree then.

maxandersen:
It does make sense to have multiple named providers; like it's possible to have named datasources - but we are starting out with the simplest approach to avoid exposing too much at runtime. This will also evolve.

edeandrea:
I've added this as #56

Issue #43, opened 2023-11-20
Eliminate `BeanXProviderSupplier`


cescoffier:
Discovering this value is hard. Adding the implementation class name might be simpler.

cescoffier:
I believe it would be simpler if the user can just write:

```
@RegisterAiService(retriever=TheClassNameOfMyBean.class)
```

geoand:
I'll have a look at this tomorrow hopefully.

geoand:
I was thinking of this, and I have a new proposal:

What if we keep what we have now, but change the default to something like: `BeanIfExistsChatMemoryProviderSupplier` which would automatically use the bean if it exists, but would not fail it no bean exists and would just not use any memory.

cescoffier:
That works!

The next issue is: what would be the value when you have multiple
retrievers and you want to select one. The retriever is not a supplier,
just a plain implementation of the retriever.

I think Suppliers are used as factories. However in our case, the scope is
what define when to create the instance.

geoand:
> I think Suppliers are used as factories

Yeah, that was my intention. If users want more advanced ways of choosing an implementation, then it's up to them to write the supplier.

geoand:
https://github.com/quarkiverse/quarkus-langchain4j/pull/51 should take care of it

PR #51, opened 2023-11-22
Remove the need to explicitly configure retriever and memory beans for @RegisterAiService

geoand:
Closes: #43


cescoffier (changes requested):
The doc should be updated too.

geoand:
> The doc should be updated too.

Done

PR #60, opened 2023-11-23
Add support for in-process embedding models

cescoffier:
- Add management for in-process models
- Fix native compilation when using in-process models
- Write documentation


cescoffier:
Trying to see if my native fixes for on CI - Failing on my machine with "file too large"

geoand (approved):
Nice!

PR #73, opened 2023-11-28
Allow injecting EmbeddingStore<TextSegment> with the type parameter

jmartisk:


jmartisk:
If this makes sense I'll also apply it in the (upcoming) Pinecone store processor

geoand (approved):

geoand:
Does this also need a docs update?

jmartisk:
The examples in the docs inject the concrete implementation class to avoid ambiguity (well, they will after https://github.com/quarkiverse/quarkus-langchain4j/pull/64) so not necessarily

Issue #46, opened 2023-11-20
Check if pg-vector from langchain4j can be used as is.


geoand:
We would probably however want a dev-service for this in any case

sebastienblanc:
@geoand I would like to help on this one, your idea is to check if we can use directly https://github.com/langchain4j/langchain4j/blob/main/langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java , correct ?

geoand:
Hey @sebastienblanc,

Yeah, exactly!

sebastienblanc:
At a first quick "brain" scan, maybe it can be used as is but it would be nice if we could pass it directly an existing Quarkus PG (eventually Reactive) datasource ...

sboeckelmann:
@sebastienblanc Sorry this may seem a may be a bit off-topic.
As the maintainer of https://github.com/quarkiverse/quarkus-opensearch I am currently investigating on integrating https://github.com/langchain4j/langchain4j/blob/main/langchain4j-opensearch/src/main/java/dev/langchain4j/store/embedding/opensearch/OpenSearchEmbeddingStore.java.
Issue and pull request will follow 😄 

It seems to be a general issue with the langchain4j EmbeddingStore implementations:
They don't support the respective Datasources or Clients to be passed directly.

And rather than copying the code it would make sense to adjust them directly within the langchain4j project.
This would make it easier to support (all of) them in Quarkus.
What's your thought on this?

geoand:
> At a first quick "brain" scan, maybe it can be used as is but it would be nice if we could pass it directly an existing Quarkus PG (eventually Reactive) datasource ...

For sure, we want to reuse the Quarkus PG stuff

geoand:
> @sebastienblanc Sorry this may seem a may be a bit off-topic. As the maintainer of https://github.com/quarkiverse/quarkus-opensearch I am currently investigating on integrating https://github.com/langchain4j/langchain4j/blob/main/langchain4j-opensearch/src/main/java/dev/langchain4j/store/embedding/opensearch/OpenSearchEmbeddingStore.java. Issue and pull request will follow 😄
> 
> It seems to be a general issue with the langchain4j EmbeddingStore implementations: They don't support the respective Datasources or Clients to be passed directly.
> 
[...]

PR #74, opened 2023-11-28
Adding support for Pgvector

sebastienblanc:
Based on this discussion : https://github.com/quarkiverse/quarkus-langchain4j/issues/46

this is still in draft but will make the review and discussions easier


geoand:
cc @jmartisk

geoand:
Thanks a lot @sebastienblanc! 

What do you still want to do on this?

geoand:
I assume this means that when the user configures:

```
quarkus.datasource.username = user
quarkus.datasource.password = pass
quarkus.datasource.jdbc.url = jdbc:postgresql://somehost:5432/some_db
```

the datasource is used to configure PgVector as well?

sebastienblanc:
> I assume this means that when the user configures:
> 
> ```
> quarkus.datasource.username = user
> quarkus.datasource.password = pass
> quarkus.datasource.jdbc.url = jdbc:postgresql://somehost:5432/some_db
> ```
> 
> the datasource is used to configure PgVector as well?

Yes ! It takes the default datasource, named datasources is not something I added yet.

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -0,0 +1,289 @@
...
+                } else {
+                    upsertStmt.setNull(3, Types.VARCHAR);
+                    upsertStmt.setNull(4, Types.OTHER);
+                }
+                upsertStmt.addBatch();
+            }
+
+            upsertStmt.executeBatch();
+
+        } catch (SQLException e) {
```
cescoffier, line 282: should we have a single catch block?

geoand:
Great!

> named datasources is not something I added yet.

We don't need to worry about that for now

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -0,0 +1,289 @@
...
+                PGvector vector = (PGvector) resultSet.getObject("embedding");
+                Embedding embedding = new Embedding(vector.toArray());
+
+                String text = resultSet.getString("text");
+                TextSegment textSegment = null;
+                if (isNotNullOrBlank(text)) {
+                    String metadataJson = Optional.ofNullable(resultSet.getString("metadata")).orElse("{}");
+
+                    // Replace TypeToken with TypeReference in Jackson
+                    TypeReference<Map<String, String>> typeReference = new TypeReference<Map<String, String>>() {
```
cescoffier, line 216: probably a good idea to keep it around in a constant.

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/runtime/PgVectorEmbeddingStoreConfig.java
```diff
@@ -0,0 +1,51 @@
+package io.quarkiverse.langchain4j.pgvector.runtime;
+
+import static io.quarkus.runtime.annotations.ConfigPhase.BUILD_AND_RUN_TIME_FIXED;
+
+import io.quarkus.runtime.annotations.ConfigRoot;
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithDefault;
+
+@ConfigRoot(phase = BUILD_AND_RUN_TIME_FIXED)
```
cescoffier, line 9: Why is is build time? I should be runtime only. No?

pgvector/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,12 @@
+name: Langchain4j pgvector embedding store
+artifact: ${project.groupId}:${project.artifactId}:${project.version}
+description: Provides the pgvector Embedding store for Langchain4j
```
cescoffier, line 3:
```suggestion
description: Provides the pgvector Embedding store for Quarkus Langchain4j
```

pgvector/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,12 @@
+name: Langchain4j pgvector embedding store
```
cescoffier, line 1:
```suggestion
name: Quarkus Langchain4j pgvector embedding store
```

cescoffier (changes requested):

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -0,0 +1,289 @@
...
+
+/**
+ * PGVector EmbeddingStore Implementation
+ * <p>
+ * Only cosine similarity is used.
+ * Only ivfflat index is used.
+ */
+public class PgVectorEmbeddingStore implements EmbeddingStore<TextSegment> {
+
+    ObjectMapper objectMapper = new ObjectMapper();
```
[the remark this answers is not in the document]
geoand: Definitely, the other extensions in this repo use one that contains the proper configuration for ObjectMapper

cescoffier:
For the tests you can reuse the once from Redis.

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -0,0 +1,289 @@
...
+                                "text TEXT NULL, " +
+                                "metadata JSON NULL" +
+                                ")",
+                        table, ensureGreaterThanZero(dimension, "dimension")));
+            }
+
+            if (useIndex) {
+                connection.createStatement().executeUpdate(String.format(
+                        "CREATE INDEX IF NOT EXISTS ON %s " +
+                                "USING ivfflat (embedding vector_cosine_ops) " +
```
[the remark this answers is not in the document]
sebastienblanc: It should IMO, here I just copy/pasted the existing class from the langchain4j project and I did not add any extra feature. (Except making it work with Quarkus)

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/PgVectorEmbeddingStore.java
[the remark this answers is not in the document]
sebastienblanc: Sorry , could you point me to the other extension that uses the one from Quarkus so I can get "inspired" :) ?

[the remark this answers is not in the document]
geoand: Sure! 

https://github.com/quarkiverse/quarkus-langchain4j/blob/main/redis/runtime/src/main/java/io/quarkiverse/langchain4j/redis/RedisEmbeddingStore.java#L162

is one example

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/runtime/PgVectorEmbeddingStoreConfig.java
sebastienblanc: Probably because I copy it from the Redis Config file.

pgvector/deployment/src/main/java/io/quarkiverse/langchain4j/pgvector/deployment/Langchain4jPgvectorProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @BuildStep
+    @Record(ExecutionTime.RUNTIME_INIT)
+    public void createBean(
+            BuildProducer<SyntheticBeanBuildItem> beanProducer,
+            PgVectorEmbeddingStoreRecorder recorder,
+            PgVectorEmbeddingStoreConfig config) {
+        beanProducer.produce(SyntheticBeanBuildItem
+                .configure(PGVECTOR_EMBEDDING_STORE)
+                .types(EmbeddingStore.class)
```
jmartisk, line 39:
```suggestion
                .types(ClassType.create(EmbeddingStore.class),
                        ParameterizedType.create(EmbeddingStore.class, ClassType.create(TextSegment.class)))
```

We recently did this for other stores to allow also injecting `EmbeddingStore<TextSegment>` instead of just the raw type

jmartisk (changes requested):

pgvector/runtime/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.quarkiverse.langchain4j</groupId>
+    <artifactId>quarkus-langchain4j-pgvector-parent</artifactId>
+    <version>999-SNAPSHOT</version>
+  </parent>
+  <artifactId>quarkus-langchain4j-pgvector</artifactId>
+  <name>Quarkus langchain4j-pgvector - Runtime</name>
+  <description>Do something useful.</description>
```
jmartisk, line 12: :)

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/runtime/PgVectorEmbeddingStoreConfig.java
jmartisk: Redis uses build time because it needs to resolve the Redis data source injection at build time, I think here it can be runtime only

jmartisk: Ah, maybe not; if you're going to also allow picking what configured data source it will use

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/PgVectorEmbeddingStore.java
sebastienblanc: Can you elaborate a bit on this ?

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/runtime/PgVectorEmbeddingStoreConfig.java
sebastienblanc: So should leave it as it is ?

pgvector/runtime/pom.xml
cescoffier: Well, it's accurate, no?

pgvector/runtime/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-arc</artifactId>
+    </dependency>
+     <dependency>
+      <groupId>io.quarkiverse.langchain4j</groupId>
+      <artifactId>quarkus-langchain4j-core</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
```
sebastienblanc, line 22: @jmartisk @cescoffier I need some help on this, I have added here this dep but for some reasons in my sample project I still need to add this dep.

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/runtime/PgVectorEmbeddingStoreConfig.java
jmartisk: So I guess the optimal solution would be to have this config class runtime-only, and if you're going to add the option to select the named postgres datasource, then that would go into a separate config class, and that one would be build-time.

pgvector/deployment/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+    <dependency>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-embeddings-all-minilm-l6-v2-q</artifactId>
+        <version>${langchain4j.version}</version>
+        <scope>test</scope>
+    </dependency>
+        <dependency>
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-jdbc-postgresql</artifactId>
+        <scope>test</scope>
```
jmartisk, line 61: This should be a dependency on the deployment module (`quarkus-jdbc-postgresql-deployment`) and should be in compile scope.
And the runtime pgvector module should have a dependency on the runtime artifact (`quarkus-jdbc-postgresql`), also in compile scope

pgvector/runtime/pom.xml
jmartisk: Not sure how you got into this state, but after applying what I suggest in https://github.com/quarkiverse/quarkus-langchain4j/pull/74/files#r1409040930 it works for me, maybe it's related

jmartisk: Then the application only needs to depend on `quarkus-langchain4j-pgvector`, nothing else

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -0,0 +1,293 @@
...
+     */
+    @Override
+    public List<EmbeddingMatch<TextSegment>> findRelevant(Embedding referenceEmbedding, int maxResults, double minScore) {
+        List<EmbeddingMatch<TextSegment>> result = new ArrayList<>();
+        try (Connection connection = setupConnection()) {
+            String referenceVector = Arrays.toString(referenceEmbedding.vector());
+            String query = String.format(
+                    "WITH temp AS (SELECT (2 - (embedding <=> '%s')) / 2 AS score, embedding_id, embedding, text, metadata FROM %s) SELECT * FROM temp WHERE score >= %s ORDER BY score desc LIMIT %s;",
+                    referenceVector, table, minScore, maxResults);
+            PreparedStatement selectStmt = connection.prepareStatement(query);
```
jmartisk, line 207: You have to close the statement, otherwise you get leaks. 
This shows in the log:
```
2023-11-29 11:06:13,666 WARN  [io.agr.pool] (executor-thread-1) Datasource '<default>': JDBC resources leaked: 1 ResultSet(s) and 2 Statement(s)
```

Multiple usages in this class are missing this

jmartisk (changes requested):

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/PgVectorEmbeddingStore.java
jmartisk: turn it into a static final variable :)

pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/PgVectorEmbeddingStore.java
sebastienblanc: Let me check this, and probably the upstream project would need to fix this as well (since I copy pasted :) )

jmartisk: Ah yes, maybe a good opportunity to contribute to that project too :)

jmartisk (approved):
Nice, technically all looks good to me now.
We're missing documentation and the option to choose a non-default datasource

sebastienblanc:
I will add soon some doc. Do we want also a simple sample project ? like we have here https://github.com/quarkiverse/quarkus-langchain4j/blob/main/samples/chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/IngestorExample.java ?

geoand:
> . Do we want also a simple sample project ? like we have here https://github.com/quarkiverse/quarkus-langchain4j/blob/main/samples/chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/IngestorExample.java ?

Yeah, that would be nice.

Also note, that we'll the commits to be squashed and rebased when done.

sebastienblanc:
> > . Do we want also a simple sample project ? like we have here https://github.com/quarkiverse/quarkus-langchain4j/blob/main/samples/chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/IngestorExample.java ?
> 
> Yeah, that would be nice.
> 
> Also note, that we'll the commits to be squashed and rebased when done.

@geoand Will try to add this early next week, I had a busy end of the week. 
For the doc is a page like this generated or do I need to write it manually https://github.com/quarkiverse/quarkus-langchain4j/blob/main/docs/modules/ROOT/pages/includes/quarkus-langchain4j-redis.adoc ? 
*

geoand:
👍

That doc is generated

sebastienblanc:
Yeah I did one this weekend but another is needed (and realized my git rebase skills were a bit rusty :) )

geoand:
For now just a sample manual test would be fine

geoand:
I rebased this onto the latest `main`

PR #76, opened 2023-11-30
Add metrics support to AiServices

geoand:
What this does is add `langchain4j.aiservices.ServiceName.methodName` timer metrics to the service method calls.

I am wondering whether I should do it also for tools, but I decided against it because the tools can potentially be invoked by other code as well.

Relates to: #8


cescoffier (approved):
Maybe a bit of doc would be useful? The rest looks great!

geoand:
> Maybe a bit of doc would be useful? The rest looks great!

Yeah, I plan to add some docs when I add the tracing support as well

PR #77, opened 2023-12-01
Add initial OpenTelemetry support for declarative services

geoand:
For the email-me-a-pom example, the outcome looks like:

Screenshot from 2023-12-01 12-48-22

Note that it contains a span for the service, spans for the REST calls (we get those for free) and a span for the tool invocation

Relates to: #8


cescoffier (approved):
Awesome!

PR #83, opened 2023-12-04
Sketch out API for auditing

geoand:


cescoffier (approved):
That's exactly what I had in mind.

geoand:
I still need to add some documentation entry

geoand:
@jmartisk the pinecone test is failing, do you have any idea?

jmartisk:
Hard to say, maybe some other CI was running in parallel with it?
That test is really hard to maintain because of the shared database :( and because we need that delay after removing items, we currently have 30 seconds, but I'm not sure it's always enough

jmartisk:
I'll try to run the tests for this PR on my laptop

geoand:
One thing I am thinking we should do is introduce the Git Incremental Builder by @famod that we also use in Quarkus, as it an awesome way of only running tests for modules that can be affected by a change

jmartisk:
I see you re-ran the tests and the Pinecone tests passed this time, but there's some other failures this time.... lovely

geoand:
On this current run it seems like pinecode failed again

geoand:
There is something very weird going on with the native tests that requires more investigation

famod:
@geoand 
> One thing I am thinking we should do is introduce the Git Incremental Builder by @famod that we also use in Quarkus, as it an awesome way of only running tests for modules that can be affected by a change

Happy to help! I should have some time for that later this month. Best create an issue with some details and assign it to me. 🙂

geoand:
Will do, thanks!

PR #85, opened 2023-12-04
Fix dependencies shown in quickstart guide

andreas-repp:
The dependencies shown in the quickstart guide don't work. When I tried it, this dependency couldn't be found:

```
<dependency>
    <groupId>io.quarkiverse.langchain4j</groupId>
    <artifactId>quarkus-langchain4j</artifactId>
    <version>{project-version}</version>
</dependency>
```

However, on the readme, the correct examples are shown. So I change the text with the one from the readme.


geoand (approved):
Nicely spotted!

Issue #88, opened 2023-12-04
Qute templating error when using Kotlin


andreas-eberle:
I have a weird bug with the qute template when I try to replicate the example from the [Quickstart](https://docs.quarkiverse.io/quarkus-langchain4j/dev/index.html#_quick_overview) with Kotlin. 

I'm using Quarkus 3.6 with Kotlin & gradle and OpenApi.

I have the following MyAiService. It is basically the one from the example but in Kotlin and I removed the Email sending part to simplify the reproduce (code attached). 

```
@RegisterAiService
interface MyAiService {
    @SystemMessage("You are a professional poet")
    @UserMessage("Write a poem about {topic}. The poem should be {lines} lines long.")
    fun writeAPoem(topic: String, lines: Int): String
}
```

When I try to call the `writeAPoem` method, I get the following error that topic would be missing:

``` 
io.quarkus.qute.TemplateException: Rendering error: Entry "topic" not found in the data map in expression {topic}
	at io.quarkus.qute.TemplateException$Builder.build(TemplateException.java:169)
	at io.quarkus.qute.EvaluatorImpl.propertyNotFound(EvaluatorImpl.java:234)
	at io.quarkus.qute.EvaluatorImpl.resolve(EvaluatorImpl.java:204)
	at io.quarkus.qute.EvaluatorImpl.resolveReference(EvaluatorImpl.java:131)
	at io.quarkus.qute.EvaluatorImpl.evaluate(EvaluatorImpl.java:85)
	at io.quarkus.qute.ResolutionContextImpl.evaluate(ResolutionContextImpl.java:29)
	at io.quarkus.qute.ExpressionNode.resolve(ExpressionNode.java:36)
	at io.quarkus.qute.SectionNode$SectionResolutionContextImpl.execute(SectionNode.java:228)
	at io.quarkus.qute.SectionHelper$SectionResolutionContext.execute(SectionHelper.java:66)
	at io.quarkus.qute.Parser$1.resolve(Parser.java:1288)
	at io.quarkus.qute.SectionNode.resolve(SectionNode.java:53)
	at io.quarkus.qute.SectionNode.resolve(SectionNode.java:58)
	at io.quarkus.qute.TemplateImpl$TemplateInstanceImpl.renderData(TemplateImpl.java:233)
	at io.quarkus.qute.TemplateImpl$TemplateInstanceImpl.renderAsyncNoTimeout(TemplateImpl.java:224)
[...]

PR #89, opened 2023-12-04
Ignore known annotations in parameter binding

geoand:
Fixes: #88

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -560,7 +560,7 @@ public class AiServicesProcessor {
 
         List<TemplateParameterInfo> templateParams = new ArrayList<>();
         for (MethodParameterInfo param : params) {
-            if (param.annotations().isEmpty()) { // if a parameter has no annotations it is considered a template variable
+            if (effectiveParamAnnotations(param).isEmpty()) { // if a parameter has no annotations it is considered a template variable
                 templateParams.add(new TemplateParameterInfo(param.position(), param.name()));
             } else {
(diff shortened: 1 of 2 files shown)


cescoffier (approved):

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -581,6 +581,19 @@ private List<TemplateParameterInfo> gatherTemplateParamInfo(List<MethodParameter
         return templateParams;
     }
 
+    private List<AnnotationInstance> effectiveParamAnnotations(MethodParameterInfo param) {
+        return param.annotations().stream().filter(ai -> {
+            String name = ai.name().toString();
+            if (name.startsWith("kotlin")) {
+                return false;
+            }
+            if (name.endsWith("NotNull")) {
```
gsmet, line 590: I wonder if you should ignore everything under `jakarta.validation.constraints.` instead.

geoand: NotNull are super common - most tools even allow custom ones, so I'll leave the rule here

cescoffier: I was wondering if we shouldn't add CheckReturnValue.

geoand: That is only for return values,  no?

gsmet: Jakarta Validation also has `@NotBlank` which could be useful in this case. Thus my recommendation.

You can keep not null as is but I would also ignore everything under jakarta.validation.constraints.

Issue #95, opened 2023-12-05
ChatMemory is persisted over request boundaries


andreas-eberle:
I was not looking a bit more into the examples from the guides page and especially the chat memory management (https://docs.quarkiverse.io/quarkus-langchain4j/dev/ai-services.html#memory). Unfortunately, I got a bit confused when I tested it as it seems to behave quite different from what I initially expected. Hopefully you can help to clarify things.

In the example code, the ChatMemoryBean is created as a RequestScoped bean. However, when you just use it, its get method is only called once with default as memoryId. At the end of the first request, it's close method is also called (as expected) and the memories map is freed. However, when you do the second request, the bean is not used any more and no new memory is created. Instead, all other requests just use the same memory.

Is this expected? If yes, why is the bean even request scoped?

One thing I noticed is that the get method is called by dev.langchain4j.service.AiServiceContext. And AiServiceContext itself has a map doing basically the same as ChatMemoryBean in the sense that it has a map from memory id to ChatMemory. I think this is where the memory lives longer than the request scope.

see also https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Quarkus.20Langchain.3A.20Chat.20Memory.20Management.20Questions

andreas-eberle:
I think this might also play into #47.

geoand:
Yeah, that's a more general thing we need to address.

geoand:
https://github.com/quarkiverse/quarkus-langchain4j/pull/96 should take care of it

PR #96, opened 2023-12-05
Make @RegisterAiService beans request scoped by default

geoand:
This is done because otherwise the chat memory
does not get cleared properly.

Fixes: #95

Do we need a docs update to go along with this?

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -23,7 +23,6 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
-import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.inject.Instance;
 
 import org.jboss.jandex.AnnotationInstance;
@@ -50,6 +49,7 @@ import io.quarkiverse.langchain4j.runtime.AiServicesRecorder;
 import io.quarkiverse.langchain4j.runtime.aiservice.AiServiceClassCreateInfo;
 import io.quarkiverse.langchain4j.runtime.aiservice.AiServiceMethodCreateInfo;
 import io.quarkiverse.langchain4j.runtime.aiservice.AiServiceMethodImplementationSupport;
+import io.quarkiverse.langchain4j.runtime.aiservice.DeclarativeAiServiceBeanDestroyer;
 import io.quarkiverse.langchain4j.runtime.aiservice.DeclarativeAiServiceCreateInfo;
 import io.quarkiverse.langchain4j.runtime.aiservice.MetricsWrapper;
 import io.quarkiverse.langchain4j.runtime.aiservice.QuarkusAiServiceContext;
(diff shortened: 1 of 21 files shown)


geoand:
@jmartisk pinecone test still failing after merging your PR :)

jmartisk:
That must be because https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/7101672400 and https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/7101669612 were triggered at about the same time. 
It guards against running the matrix together within a single build, but it doesn't guard about somebody triggering two independent builds at once by PRs coming from the main repo...

jmartisk:
We'll need to submit PRs from forks unless absolutely necessary :)

geoand:
Ah gotcha, thanks!

geoand:
I think I'm gonna take out the pinecone test altogether and only enable it in the nightly build.
It's too big of a toll on test parallelism for the sake of just one test

cescoffier:
I’m not sure about that. Typically, rest clients are application-scoped. We should change how we handle memory instead.

jmartisk:
> I think I'm gonna take out the pinecone test altogether and only enable it in the nightly build. It's too big of a toll on test parallelism for the sake of just one test

Yeah I kinda agree. You can disable it by removing the propagation of secrets into workflows. The test runs when it sees a non-empty value of `PINECONE_API_KEY`.

cescoffier:
Are you sure this will continue to work with the web socket? It will create many instances of the AI service, which is what we want to avoid. 

I do not get the issue with the creation part. I would need to have a look. Maybe we need to prioritize the scope thingy.

geoand:
> Are you sure this will continue to work with the web socket?

See the changes to the examples :). Essentially it works if you use context propagation

geoand:
> I do not get the issue with the creation part. I would need to have a look. Maybe we need to prioritize the scope thingy.

What happens is that although the `ChatMemoryProvider` closes and removes all content, the `ChatMemory` of `AiServicesContext` still uses them.
If you try to remove them from there as well, you'll get a bunch of breaking tests that use `AiServices.create`

cescoffier:
What if we use a proxy/facade in-between that uses the ARC API to instantiate the memory in the proper context?

cescoffier:
Also, what about having AiServicesContext in a special scope and not the AIService?

geoand:
I thought of that. The problem is producing it, but I'll look into it

geoand:
@jmartisk the nightly build with the pinecone test is failing very weirdly: https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/7108550795/job/19352035493

jmartisk:
Holy * that test is obnoxious. Looking into it...

jmartisk:
Nice, Pinecone has changed the response format and they now don't include an empty `metadata` object in the response if a vector has no metadata, causing the NPE. I will send a PR to adjust it on our side....

geoand:
Thanks!

geoand:
> Also, what about having AiServicesContext in a special scope and not the AIService?

This is what I am trying now, but so far it has subtle issues, not to mention that the code is a lot more complicated than what I have pushed here. 
I do believe that what I pushed is a fine solution, there are no real downsides of having the client be request scoped.

geoand:
Also, the more I think about it, the more it makes sense to me that the context (which is an entity unknown to the user) should have the same scope as the bean itself.

geoand:
> We need to go back to the whiteboard to find a proper way to deal with this.

I totally agre

PR #107, opened 2023-12-07
Update chat memory example and separate ChatMemoryStore from ChatMemory usage

andreas-repp:
This PR updates the ChatMemory example. The idea is to separate providing the `ChatMemory` creation in the `ChatMemoryProviderBean` and the `ChatMemoryStore` creation via a `ChatMemoryStoreProducer`.


docs/modules/ROOT/examples/io/quarkiverse/langchain4j/samples/ChatMemoryProviderBean.java
```diff
@@ -0,0 +1,24 @@
...
+@ApplicationScoped
+public class ChatMemoryProviderBean implements ChatMemoryProvider {
+
+    @Inject
+    ChatMemoryStore store;
+
+    @Override
+    public ChatMemory get(Object memoryId) {
+        return MessageWindowChatMemory.builder()
+                .id(memoryId)
```
cescoffier, line 19: without eviction or4 cleanup, this will lead to OOM.

cescoffier (changes requested):
I'm not totally sure. I don't see any cleanup mechanism, which is likely going to OOM at some point.

docs/modules/ROOT/examples/io/quarkiverse/langchain4j/samples/ChatMemoryProviderBean.java
geoand: I need to see exactly how this is meant to work.

andreas-eberle: Yes, there is currently no eviction mechanism. However, for a simple example, that would be hard to do. The current idea of always evicting when the request ends makes a chat with API calls impossible.

cescoffier: The documentation should mention or have a more complex example with eviction. 
I'm still unsure about the request scope part. I think, we would need to introduce a "conversation" concept soon.

geoand:
I am looking into a little more. There is a chance that things are simpler than we think

geoand:
I took the suggestion to use `ChatMemoryProvider` and incorporated it into a wider scoped change in https://github.com/quarkiverse/quarkus-langchain4j/pull/112.

Thanks for bringing up this idea!

Issue #93, opened 2023-12-05
PgVector processor needs to produce EmbeddingStoreBuildItem (or find a completely different approach)


jmartisk:
We produce `EmbeddingStoreBuildItem` when a processor creates a bean for an embedding store. This is for the `Langchain4jDevUIProcessor` to be able to tell whether to register the page for working with embedding stores and its JSON-RPC backend. 

Or, perhaps as an improvement, we could use something more akin to `SelectedEmbeddingModelCandidateBuildItem` for embedding stores too, but it would somehow need to be produced earlier. `SelectedEmbeddingModelCandidateBuildItem` is produced after `BeanDiscoveryFinishedBuildItem` exists, which is too late, because we need to react to it by potentially registering the JSON-RPC bean, and that isn't possible if bean discovery is already finished.

PR #110, opened 2023-12-08
Allow injecting non-default datasource for PgVector store, produce an…

jmartisk:
… EmbeddingStoreBuildItem

Fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/93


geoand (approved):
Nice!

geoand:
We probably want something like this for Redis too

Issue #47, opened 2023-11-21
Proper CDI scope for conversation


cescoffier:
Memory is required to keep tracks of the previous messages in a conversation. It's also required when using tools.

This issue is about addressing two issues:

- how to configure the memory size - even if it's just a message count, it may not work as the context would be exceeded (especially when using RAG). Also, eviction is tricky
- how to properly define the boundary of the memory - like the web socket session

Of course, the memory storage must be per conversation / users and not for the application. 

An idea is to define a new CDI scope that would handle the second issue.

geoand:
We need this very very much :)

geoand:
Let's start with a specific implementation of memory for Websockets and see how far that gets us.

agoncal:
What about calling it `@ConversationScoped` ;o)

cescoffier:
@agoncal That's exactly the problem. In addition, 

1) we would need "session" support (Ladislav is working on this).
2) a user may start multiple conversations concurrently, making things slightly more complicated.

mkouba:
I believe that a new well-defined scope would be more appropriate. I do remember the hard times trying to fix all the bugs related to the conversation context in Weld.

geoand:
@mkouba just so we can be aware, what sort of issues where there?
Asking because I am sure similar problems will arise in this context

mkouba:
> @mkouba just so we can be aware, what sort of issues where there? Asking because I am sure similar problems will arise in this context

Well, I don't remember all the details but I can try to walk through the Weld issues/codebase in order to identify the problematic parts. I do remember that many issues were related to concurrent access and lifecycle (such as "an HTTPSession is destroyed in the middle of a request processing"). We also had to initialize the context lazily because of some character encoding issues (see http://weld.cdi-spec.org/documentation/#3). You also need a mechanism to prune long-running conversations that were not explicitly ended (i.e. timed out). Anyway, the conversation context is bound to the Servlet API. So at least it would be very confusing if we support this scope on top of Vert.x.

geoand:
Thanks for the info!

exabrial:
FYI, there is already a CDI Scope called `ConversationScoped`, may want to avoid a name collision

geoand:
Yeah, we definitely don't want more confusion

maxandersen:
> how to properly define the boundary of the memory - like the web socket session

[...]

PR #112, opened 2023-12-08
Overhaul how chat memory is configured

geoand:
The basic ideas are:

* A default `ChatMemoryStore` bean using in memory storage is now provided 
  - This can be overridden by users simply by providing their own bean
* A default `ChatMemoryProvider` bean is now provided 
  - It uses the `ChatMemoryStore` bean under the hood 
  - It also leverages configuration for use either `MessageWindowChatMemory` or `TokenWindowChatMemory`

`ChatMemory` objects continue to be removed when the AI service class goes out of scope, however now there is also now a programmatic way of removing `ChatMemory` by using `ChatMemoryRemover`. This is necessary for when the AI service is not request scoped.


core/runtime/src/main/java/io/quarkiverse/langchain4j/ChatMemoryRemover.java
```diff
@@ -0,0 +1,39 @@
...
+    private static final Object[] EMPTY_OBJECT_ARRAY = new Object[0];
+
+    private ChatMemoryRemover() {
+    }
+
+    /**
+     * @param aiService The bean that implements the AI Service annotated with {@link RegisterAiService}
+     * @param memoryId The object used as memory IDs for which the corresponding {@link ChatMemory} should be removed
+     */
+    public static void remove(Object aiService, Object memoryId) {
```
geoand, line 24: I don't like using two objects here as users could easily swap them, but I don't have a better way of handling it...

maxandersen:
context (no pun intended) For  #47 :)

docs/modules/ROOT/pages/ai-services.adoc
```diff
@@ -175,24 +176,40 @@ public class MyChatModelSupplier implements Supplier<ChatLanguageModel> {
...
+The `chatMemoryProviderSupplier` attribute of the `@RegisterAiService` annotation enables configuring the `dev.langchain4j.memory.chat.ChatMemoryProvider`. The default value of this annotation is `RegisterAiService.BeanIfExistsChatMemoryProviderSupplier.class`
 which means that the `AiService` will use whatever `ChatMemoryProvider` bean is configured by the application, while falling back to no memory if no such bean exists.
-An example of such a bean is:
+
+The extension provides a default implementation of `ChatMemoryProvider` which does two things:
+
+* It uses whatever bean `dev.langchain4j.store.memory.chat.ChatMemoryStore` bean is configured, as the backing store. The default implementation is `dev.langchain4j.store.memory.chat.InMemoryChatMemoryStore`
+** If the application provides its own `ChatMemoryStore` bean, that will be used instead of the default `InMemoryChatMemoryStore`,
+* It leverages the available configuration options under `quarkus.langchain4j.chat-memory` to construct the  `ChatMemoryProvider`.
+** The default configuration values result in the usage of `dev.langchain4j.memory.chat.MessageWindowChatMemory` with a window size of ten
```
maxandersen, line 187: is there a way to make it unbounded window?

if user wanted to keep the window but instead passivate/persist memory that have not been accessed for a while is that where ChatMemoryRemover come into play?

cescoffier (approved):

docs/modules/ROOT/pages/ai-services.adoc
geoand: > is there a way to make it unbounded window?

Nope, Langchain4j doesn't have that - you would need to write you own

> if user wanted to keep the window but instead passivate/persist memory that have not been accessed for a while is that where ChatMemoryRemover come into play?

Nope, see the updates samples that use `ChatMemoryRemover` to see when it's meant to be used

geoand: When we have clear scopes (or when the existing request scope is used), everything works magically 😄.


I mean that the extension already has the logic to remove chat messages from the store

geoand:
I'd like @andreas-eberle to also have a look since he raised the concern about `ChatMemoryStore` yesterday

geoand:
Once this is in, I'll do a `0.4.0` release

docs/modules/ROOT/examples/io/quarkiverse/langchain4j/samples/CustomProvider.java
```diff
@@ -0,0 +1,25 @@
...
+    @Override
+    public ChatMemoryProvider get() {
+        return new ChatMemoryProvider() {
+
+            @Override
+            public ChatMemory get(Object memoryId) {
+                return MessageWindowChatMemory.builder()
+                        .maxMessages(20)
+                        .id(memoryId)
+                        .chatMemoryStore(new InMemoryChatMemoryStore())
```
andreas-eberle, line 20: Does it make sense to create a new store here every time?

andreas-eberle:
Looks good to me. Thanks @geoand !

docs/modules/ROOT/examples/io/quarkiverse/langchain4j/samples/CustomProvider.java
geoand: Good catch! I fixed it

PR #121, opened 2023-12-11
Add context to model config docs

maxandersen:
- pass cause to log to know actual error
- called out the explicit models + some docs/links


maxandersen:
bad merge/branch.

PR #122, opened 2023-12-11
called out the explicit models + some docs/links

maxandersen:
wanted to just add links to the openai model config keys but then saw all the repeated "model" in docs that I was not finding when searching for i.e. "moderation model" so started adding them.

but then realized this will have to be duplicated for all the various models....not sure if any better way but at least with this ctrl+F works for me :)

### DIFF
--- a/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/ChatModelConfig.java
+++ b/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/ChatModelConfig.java
@@ -9,7 +9,9 @@ import io.smallrye.config.WithDefault;
 public interface ChatModelConfig {
 
     /**
-     * Model name to use
+     * Chat model name to use.
+     * 
+     * See https://platform.openai.com/docs/models/overview for a list of available models. 
      */
     @WithDefault("gpt-3.5-turbo")
     String modelName();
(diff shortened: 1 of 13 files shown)


geoand (approved):
Makes sense, thanks!

cescoffier (approved):

geoand:
This needs a formatting fix

openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/EmbeddingModelConfig.java
```diff
@@ -7,7 +7,9 @@
 public interface EmbeddingModelConfig {
 
     /**
-     * Model name to use
+     * Embedding Model name to use
+     * 
+     * See https://platform.openai.com/docs/guides/embeddings/embedding-models for a list of available models.
```
cescoffier, line 12: Should be proper links `(<a href...`)

openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/ModerationModelConfig.java
```diff
@@ -7,7 +7,9 @@
 public interface ModerationModelConfig {
 
     /**
-     * Model name to use
+     * Moderation Model name to use. 
+     *
+     * See https://platform.openai.com/docs/guides/moderation/overview for a list of available models.
```
cescoffier, line 12: Should be proper links `(<a href...`)

openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/ChatModelConfig.java
```diff
@@ -9,7 +9,9 @@
 public interface ChatModelConfig {
 
     /**
-     * Model name to use
+     * Chat model name to use.
+     * 
+     * See https://platform.openai.com/docs/models/overview for a list of available models.
```
cescoffier, line 14: Should be proper links `(<a href...`)

geoand:
Let's close this as it is really out of date

PR #125, opened 2023-12-11
Make the `demo` OpenAI key usable

geoand:
For some reason when using the 'demo' key,
the response comes back as type text/html
but the content is still JSON


geoand:
This was first found by @lordofthejars

cescoffier (approved):

PR #156, opened 2023-12-14
Milvus embedding store + Dev services

jmartisk:
Unfortunately I had to go with the official Milvus Java client in the end, and that contains some crazy dependencies that probably prevent us from compiling in native for now.

I tried a REST client, but Milvus's REST api seems quite unfinished and doesn't seem to allow specifying the schema of a newly being created collection.

But hey, with this approach, it should be relatively low amount of maintenance work for us, because we basically just take what langchain4j has, without changes.

The bad part is that since we're using their gRPC client it's not integrated with the gRPC extension (so no metrics, tracing etc)


geoand:
> The bad part is that since we're using their gRPC client it's not integrated with the gRPC extension (so no metrics, tracing etc)

Hopefully at some point in the future we can use our own which would allow us to reap the many benefits as you mentioned

Issue #148, opened 2023-12-13
Building native binary using openai extension doesn't work


edeandrea:
Building a native binary using the `quarkus-langchain4j-openai` extension does not work:

```
╰─ ./mvnw --version                                                                                                   
Apache Maven 3.9.6 (bc0240f3c744dd6b6ec2920b3cd08dcc295161ae)
Maven home: /Users/edeandre/.m2/wrapper/dists/apache-maven-3.9.6-bin/3311e1d4/apache-maven-3.9.6
Java version: 17.0.9, vendor: Eclipse Adoptium, runtime: /Users/edeandre/.sdkman/candidates/java/17.0.9-tem
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "14.1.2", arch: "aarch64", family: "mac"

╰─ java --version
openjdk 17.0.9 2023-10-17
OpenJDK Runtime Environment Temurin-17.0.9+9 (build 17.0.9+9)
OpenJDK 64-Bit Server VM Temurin-17.0.9+9 (build 17.0.9+9, mixed mode)

╰─ $GRAALVM_HOME/bin/java --version
openjdk 21.0.1 2023-10-17
OpenJDK Runtime Environment GraalVM CE 21.0.1+12.1 (build 21.0.1+12-jvmci-23.1-b19)
OpenJDK 64-Bit Server VM GraalVM CE 21.0.1+12.1 (build 21.0.1+12-jvmci-23.1-b19, mixed mode, sharing)

╰─ ./mvnw clean package -DskipTests -Pnative                                                                    
[INFO] Scanning for projects...
[INFO] 
[INFO] -----------< io.quarkus.sample.super-heroes:rest-narration >------------
[INFO] Building Quarkus Sample :: Super-Heroes :: Narration Microservice 1.0
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- clean:3.2.0:clean (default-clean) @ rest-narration ---
[INFO] Deleting /Users/edeandre/workspaces/quarkus/quarkus-super-heroes/rest-narration/target
[INFO] 
[INFO] --- jacoco:0.8.11:prepare-agent (default-prepare-agent) @ rest-narration ---
[...]

PR #157, opened 2023-12-14
Use previous version of POI extension

geoand:
This is needed because Langchain4j
used POI 5.2.3 which requires the
substitution present in version 2.0.4
of the extension in order to work
on Mac


cescoffier (approved):

edeandrea:
I've confirmed this resolves #148

Issue #164, opened 2023-12-16
OpenAiRestApiJacksonProvider affecting non extension clients


yvesv:
Ok, so I'm quite new to Quarkus and figuring things out. But after adding the extension to my projects. Other clients in my project weren't properly deserializing anymore because the extension 'forced' to use the OpenAiRestApiJacksonProvider. I'm currently searching for a way to work around this issue. But preferably of course the extension only affects the clients of the extension.

geoand:
Very interesting.

Any chance you can attach a sample project that shows this problem?

geoand:
https://github.com/quarkiverse/quarkus-langchain4j/pull/165 should fix it

yvesv:
Wow. You are fast! Thanks.

geoand:
Unfortunately CI is complaining, so I'll have to take a closer look on Monday.

geoand:
See https://github.com/quarkiverse/quarkus-langchain4j/pull/165#issuecomment-1858871062. 

This means that it will take a few days to get a new release out with the fix

yvesv:
Thanks. Did a quick check, and this indeed fixed the issue.

geoand:
Thanks for checking!

PR #165, opened 2023-12-16
Limit scope of OpenAiRestApi providers

geoand:
The previous way they were declared meant
that they would affect other REST Clients
too.

- Fixes: #164


geoand:
Unfortunately this can only be fixed once https://github.com/quarkusio/quarkus/pull/37787 is in Quarkus

geoand:
We now have Quarkus 3.6.4 in the extension so this should now pass

cescoffier (approved):

PR #171, opened 2023-12-20
Minor Dev UI fixes for tools

jmartisk:
- don't cause NPE if `ToolsMetadataBuildItem` is null
- properly show the tool's name if it's not the same as the method name


geoand (approved):
Thanks!

PR #182, opened 2023-12-21
Add DevServices support for Ollama model

lordofthejars:
This PR adds dev services to Ollama, so in dev mode, you don't need to have Ollama running locally; it starts an Ollama container and installs the first time the model you want to use. After that, every run against the same model will reuse the created image.

One of the things in this implementation that is different from other devservices is that it is developed in its module, so you need to add the module to enable it. 

The reason it is this way is that the Ollama container takes a lot of minutes to start the first time, and also when running it and querying locally might take  some seconds, so maybe it's not always the best to have devservices running. I know there is the enable flag, so here we can decide if we want to set the flag enabled to false by default and let the user explicitly set to true (which means that developers might not notice there is a devservice by default) or we explicitly make them add as a dependency enabled to true by default


geoand:
Thanks for this!

@cescoffier WDYT?

lordofthejars:
Tests are not the best, I followed what I've found. I tested locally

cescoffier (approved):
Awesome! Thanks!

lordofthejars:
The build is failing because of modules not related with my change.

geoand:
Indeed, I restarted CI

lordofthejars:
When the PR is merged, I'll update the documentation with DevServices section.

Issue #195, opened 2024-01-02
Dev UI support for chatting


jmartisk:
Inject a chat model from CDI and allow the user to have conversations with it. 
Probably support just one conversation at a time, with the option to start over (delete the memory)

PR #196, opened 2024-01-02
Dev UI chatting capability

jmartisk:
Fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/195

Adds a page for chatting when there's a chat model available .

It assumes that only one chat model registered in CDI is allowed, and that a single chat memory provider will also be always available (are these assumptions ok?)

It allows to have a conversation utilizing a memory and then click a button to restart, dropping the memory

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/devui/Langchain4jDevUIProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/devui/Langchain4jDevUIProcessor.java
@@ -8,6 +8,8 @@ import io.quarkiverse.langchain4j.deployment.DeclarativeAiServiceBuildItem;
 import io.quarkiverse.langchain4j.deployment.EmbeddingModelBuildItem;
 import io.quarkiverse.langchain4j.deployment.EmbeddingStoreBuildItem;
 import io.quarkiverse.langchain4j.deployment.ToolsMetadataBuildItem;
+import io.quarkiverse.langchain4j.deployment.items.ChatModelProviderCandidateBuildItem;
+import io.quarkiverse.langchain4j.runtime.devui.ChatJsonRPCService;
 import io.quarkiverse.langchain4j.runtime.devui.EmbeddingStoreJsonRPCService;
 import io.quarkiverse.langchain4j.runtime.tool.ToolMethodCreateInfo;
 import io.quarkus.deployment.IsDevelopment;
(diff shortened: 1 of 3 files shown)


jmartisk:
@geoand , `NoAiServicesTest` is complaining because it assumes that no AI services are found in an application, but with this PR we always add one that is found during build... We need the AI service to be able to use chat memory... Any suggestion what to do? 
Is there a way to have the AI service found only in dev mode to avoid having it scanned when building for prod/test mode? Add a specific exception to ignore this one based on the class name?

geoand:
This sounds very cool!!!

 I'll have a look tomorrow when I'm back.

What I do think we should add (if it's not already in) is the ability to set a system message. This way users can experiment quickly with different prompts.

As a follow up, I think it would also be nice to be able to tweak the model settings thus giving users even more ease in interacting with the model.
And finally, if we also add some kind of history, it would allow users to see what the different options resulted in.

jmartisk:
> What I do think we should add (if it's not already in) is the ability to set a system message. This way users can experiment quickly with different prompts.

Is it possible to set a system message dynamically at runtime? I can only see you can set it directly in the `@SystemMessage` annotation on the AI service, so it has to be set at compile time.

> 
> As a follow up, I think it would also be nice to be able to tweak the model settings thus giving users even more ease in interacting with the model. 

What exactly do you mean by tweaking the model? We just inject the pre-configured model from CDI.

> And finally, if we also add some kind of history, it would allow users to see what the different options resulted in.

The next step might be to implement "chatting" with AI services, but it will be trickier due to having to deal with (de)serialization of model classes and parameters, as opposed to raw usage of a `ChatLanguageModel`, where the whole prompt and response are simply two strings.
I guess a "history" to compare different conversations might be more relevant to this approach.

geoand:
> Is it possible to set a system message dynamically at runtime? I can only see you can set it directly in the @SystemMessage annotation on the AI service, so it has to be set at compile time.

Yeah, the `ChatLanguageModel` allows you to send any kind of message you like, so we would just add a `SystemMessage` as the first message.

> What exactly do you mean by tweaking the model? We just inject the pre-configured model from CDI.

Right, my point is that we would take that model and copy it, while using whatever model parameters the user configured in this view.

> I guess a "history" to compare different conversations might be more relevant to this approach.

+1

geoand:
> NoAiServicesTest is complaining because it assumes that no AI services are found in an application, but with this PR we always add one that is found during build... We need the AI service to be able to use chat memory... Any suggestion what to do?

Do we really need an AI service for this feature?

jmartisk:
> 
> Do we really need an AI service for this feature?

Right, probably not, I'll rework it, hold on

jmartisk:
I've removed the usage of AI service now

geoand (approved):
Nice!

geoand:
I think we should also do something similar for images now that we have an ImageModel

jmartisk:
Great idea, I've reported https://github.com/quarkiverse/quarkus-langchain4j/issues/202

PR #197, opened 2024-01-02
Allow specifying the organizationId in the configuration

### LINKED ISSUE #131: Allow for specifying the organization id in the configuration
See https://platform.openai.com/docs/api-reference/organization-optional - allow the organizationId to be specified in the configuration.

edeandrea:
See https://platform.openai.com/docs/api-reference/organization-optional - allow the organizationId to be specified in the configuration.

Fixes #131

### DIFF
--- a/integration-tests/openai/src/main/java/org/acme/example/openai/QuarkusRestApiResource.java
+++ b/integration-tests/openai/src/main/java/org/acme/example/openai/QuarkusRestApiResource.java
@@ -16,9 +16,7 @@
 */
 package org.acme.example.openai;
 
-import static org.acme.example.openai.MessageUtil.createChatCompletionRequest;
-import static org.acme.example.openai.MessageUtil.createCompletionRequest;
-import static org.acme.example.openai.MessageUtil.createEmbeddingRequest;
+import static org.acme.example.openai.MessageUtil.*;
 
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -50,6 +48,7 @@ public class QuarkusRestApiResource {
 
     private final OpenAiRestApi restApi;
     private final String token;
+    private final String organizationId;
 
     public QuarkusRestApiResource(Langchain4jOpenAiConfig runtimeConfig)
             throws URISyntaxException {
@@ -58,7 +57,7 @@ public class QuarkusRestApiResource {
                 .build(OpenAiRestApi.class);
         this.token = runtimeConfig.apiKey()
                 .orElseThrow(() -> new IllegalArgumentException("quarkus.langchain4j.openai.api-key must be provided"));
-
+        this.organizationId = runtimeConfig.organizationId().orElse(null);
     }
 
     @GET
@@ -66,7 +65,10 @@ public class QuarkusRestApiResource {
     public String chatSync() {
         return restApi.blockingChatCompletion(
                 createChatCompletionRequest("Write a short 1 paragraph funny poem about segmentation fault"),
-                OpenAiRestApi.ApiMetadata.of(token, null))
+                OpenAiRestApi.ApiMetadata.builder()
+                        .apiKey(token)
+                        .organizationId(organizationId)
+                        .build())
                 .content();
     }
 
@@ -75,7 +77,10 @@ public class QuarkusRestApiResource {
     public Uni<String> chatAsync() {
         return restApi
                 .createChatCompletion(createChatCompletionRequest("Write a short 1 paragraph funny poem about Unicode"),
-                        OpenAiRestApi.ApiMetadata.of(token, null))
+                        OpenAiRestApi.ApiMetadata.builder()
+                                .apiKey(token)
+                                .organizationId(organizationId)
+                                .build())
                 .map(ChatCompletionResponse::content);
     }
 
@@ -85,7 +90,10 @@ public class QuarkusRestApiResource {
     public Multi<String> chatStreaming() {
         return restApi.streamingChatCompletion(
                 createChatCompletionRequest("Write a short 1 paragraph funny poem about Enterprise Java"),
-                OpenAiRestApi.ApiMetadata.of(token, null))
+                OpenAiRestApi.ApiMetadata.builder()
+                        .apiKey(token)
+                        .organizationId(organizationId)
+                        .build())
                 .map(r -> {
                     if (r.choices() != null) {
                         if (r.choices().size() == 1) {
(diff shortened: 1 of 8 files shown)


edeandrea:
Just an implementation note. In `OpenAiRestApi` I didn't add an `@ClientHeaderParam` annotation because then I would have had to modify every method signature to add a `@NotBody String organizationId` parameter to it. I thought instead adding a `ResteasyReactiveClientRequestFilter` that conditionally added the header would be a better/cleaner approach.

geoand:
Thanks for this.

I'll have to think about this, I'm not really a fan of the filter for this case.

This might be an opportunity actually to improve the REST Client... 
But like I said, I have to think about it 🙂

edeandrea:
Yeah I'd much prefer to have just the annotation as well, but I thought having to add an extra method parameter to every method, when a lot of the time the value is `null` wasn't the best option either. This way, the filter is only ever added to the chain if the value is even specified, which is better IMO.

geoand:
The thing is that I don't want to have two different approaches for the parameters if it can be avoided.

I'm thinking about some improvements to the `@BeanParam` support of the REST Client (I think we support it, but I could be mistaken) that could apply in this case

edeandrea:
I thought about making the filter more generic as well, something like a `ClientHeaderFilter` where we could configure it with different headers, rather than a specific filter for each set of headers we might want. But for now I kept it simple.

Some of the other edits (like transforming annonymous class instances to lambdas) we done by IntelliJ automatically. I can turn that off and revert those changes if you'd prefer.

geoand:
Yeah, no lambdas please 🙂.

But in any case, I need to look into what I mentioned above before moving this ahead

edeandrea:
> Yeah, no lambdas please

Ok I'll revert that. Can I ask why?

I feel like

```java
return builder::build;
```

is way cleaner than

```java
return new Supplier<>() {
            @Override
            public Object get() {
                return builder.build();
            }
        };
```

and

```java
public void cleanUp(ShutdownContext shutdown) {
     shutdown.addShutdownTask(QuarkusOpenAiClient::clearCache);
}
```

is way cleaner than

```java
    public void cleanUp(ShutdownContext shutdown) {
        shutdown.addShutdownTask(new Runnable() {
            @Override
            public void run() {
                QuarkusOpenAiClient.clearCache();
            }
        });
    }
```

geoand:
> Ok I'll revert that. Can I ask why?

Certainly! 
The reason is that early on during Quarkus' development, we found that when a larger number of lambdas where bootstrapped during application initialization, startup time and RSS took a small hit.

geoand:
Assuming #198 works fine (it works just fine in whatever test I ran), this should be rebased onto it and it will be much simpler

edeandrea:
I'll take a look at that and rebase onto this. Had somewhat of a pet medical emergency today so I might not get to it until tomorrow.

geoand:
Hopefully your pet will be fine!

openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/OpenAiRestApi.java
```diff
@@ -298,6 +300,19 @@ public void aroundWriteTo(WriterInterceptorContext context) throws IOException,
         }
     }
 
+    class OpenAiOrganizationIdRequestFilter implements ResteasyReactiveClientRequestFilter {
```
geoand, line 303: The filter should no longer be necessary now that I added the ApiMetadata class

edeandrea:
Thanks!

So I started rebasing this and saw what you did with the `ApiMetadata` metadata class and `@BeanParam`. Question - this organizationId header is optional, meaning that it shouldn't always be set/sent, and should never be sent in the Azure OpenAI case.

If I add another `@HeaderParam` value in the class, will that value be sent if the value is `null`?

geoand:
A header (or whatever) will not be sent if the value is null

edeandrea:
> A header (or whatever) will not be sent if the value is null

Thats what I needed to know. I'll work on this new integration now.

Just a thought - instead of adding a new method parameter to `public static ApiMetadata of(String apiKey, String apiVersion)` what if I instead created a builder for the `ApiMetaData`? That way in the future whenever we want to add new things we don't have to touch everywhere in the codebase where we use that specific method?

geoand:
Yeah, we can certainly have a builder, but we need to be careful of how the key translates to the two headers

edeandrea:
> Yeah, we can certainly have a builder, but we need to be careful of how the key translates to the two headers

I'll put something together and you can review.

geoand:
👌

edeandrea:
Just pushed changes. Let me know what you think.

integration-tests/openai/src/main/java/org/acme/example/openai/QuarkusRestApiResource.java
```diff
@@ -16,9 +16,7 @@
 */
 package org.acme.example.openai;
 
-import static org.acme.example.openai.MessageUtil.createChatCompletionRequest;
-import static org.acme.example.openai.MessageUtil.createCompletionRequest;
-import static org.acme.example.openai.MessageUtil.createEmbeddingRequest;
+import static org.acme.example.openai.MessageUtil.*;
```
geoand, line 19: I'd prefer we don't use star imports

geoand (commented):
I added a minor comment.

Also, it would be nice to have the commits squashed

integration-tests/openai/src/main/java/org/acme/example/openai/QuarkusRestApiResource.java
edeandrea: I ran `mvn impsort:sort` from the project root and thats what it produced. If I commit without running that then the build fails saying I need to run it...

geoand (approved):
Thanks a lot!

geoand:
I don't *think* I said it's an antipattern.
I definitely did say that for reasons of runtime footprint we avoid lambdas wherever possible

edeandrea:
> I don't _think_ I said it's an antipattern.

No, I don't think you did. I'm saying it is :)

edeandrea:
Not sure why the native java 21 build is failing. Its timing out when creating the binary. I've seen that happen in other apps for java 21. Takes longer to build the image.

geoand:
That's a bug in GraalVM that we need to address.

Issue #181, opened 2023-12-21
Cover the whole Ollama API


sebastienblanc:
I can see that Quarkus extension only covers the `/api/generate` Ollama endpoint but do we plan to cover everything like here https://github.com/langchain4j/langchain4j/blob/main/langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaApi.java ? 
I'm mostly interested in the `embeddings` endpoint. I can also try to contribute to this.

geoand:
We could certainly use the `embeddings` and `chat` models :)

sebastienblanc:
I was looking at the code and saw this https://github.com/quarkiverse/quarkus-langchain4j/blob/main/ollama/deployment/src/main/java/io/quarkiverse/langchain4j/ollama/deployment/OllamaProcessor.java#L56-L64 , did you start a first attempt for embedding ? Was it not working or just no time to implement it ?

geoand:
I honestly don't remember :)

PR #201, opened 2024-01-03
Add embeddings endpoint support for Ollama

sebastienblanc:
Tested on my sample project and worked, you have to set the chat-model because the default is `latest` which feels weird (maybe based on old ollama docs ? ) 

- Relates to: #181


geoand (approved):
Nice, thanks a lot!

PR #209, opened 2024-01-07
Adding the user config property for openai embedding model

sebastienblanc:
This fixes a regression that appeared with `0.5.1` , the OpenAI Embedding request is failing because an user with `null` value was passed to the request. 
This add the `user` property , I added a default value because that was the only way to get the request passed.


geoand (approved):
Thanks!

PR #222, opened 2024-01-12
Add first version of IBM BAM support

geoand:


andreadimaio:
Hi @geoand , I'm looking at the code that you are writing to implement the `watsonx` module in the `quarkus-langchain4j` extension, and from the defaults in the [Langchain4jWatsonxConfig.java](https://github.com/quarkiverse/quarkus-langchain4j/pull/222/files#diff-795ff47d552cb904f0d310e3bdf13464778f70bb302e62d396ae6f50055772a9) and [ChatModelConfig.java](https://github.com/quarkiverse/quarkus-langchain4j/pull/222/files#diff-74eabeef452ed71bd29a5502dc1019202197ad85c019950d6ef8d9a6b6aab68f) I understand that the **BAM APIs** are your starting point.

This could be a problem because the API endpoint that you are calling in the [WatsonxChatModel.java](https://github.com/quarkiverse/quarkus-langchain4j/pull/222/files#diff-0d1a1ee37a39a5c8ab402f3590f64107995543125037d2be05ef4b6210c22c84) class, the `v2/text/chat`, is not present in **watsonx**. So if you change the base URL in `application.properties` to switch from **BAM** to **watsonx** you will get an exception `[404]`.

The correct url to call should be `ml/v1-beta/generation/text` which is very similar to the [text generation API](https://bam.res.ibm.com/docs/api-reference#text-generation) in **BAM**.

This is what I understood by reading some of the online documentation and trying the free version of the **watsonx** product.
I hope this could be useful for what you are doing  😃

andreadimaio:
Another two differences are:

- To call a REST Api you need to create a **project**, each **project** has a `project_id` and this must be passed as a parameter in the request body.

- [Authentication](https://cloud.ibm.com/docs/account?topic=account-iamtoken_from_apikey&code=curl) requires an intermediate step. From an **API Key** you need to generate an **IAM Token** (valid for 60 minutes) which is used to execute the REST API call.

Maybe using the BAM APIs as a starting point is not so good because from my understanding today there are differences that don't allow your code to be used for both products.

cescoffier:
I guess we need to rename in into bam instead of watsonx.

andreadimaio:
If your idea is to integrate the **BAM APIs** into the `quarkus-langchain4j` extension, I could contribute with my code.
The only thing you need to know is that the **BAM API** is for internal use only, which means that the new module cannot be used by the entire Quarkus community.

geoand:
Thanks for the input @andreadimaio!

So maybe we should just keep this as for us to be able to experiment with BAM is and not merge it?

cescoffier:
I would just rename it and keep it there. We are likely going to use this
for demo purposes.

geoand:
PR updated to replace all occurences of `watsonx` with `bam`

cescoffier (approved):

PR #234, opened 2024-01-22
Added apiVersion parameter to AzureOpenAiEmbeddingModel and AzureOpen…

Yanick-Salzmann:
Currently the implementation of AzureOpenAiRecorder only passes the apiVersion parameter when creating an AzureOpenAiChatModel but not when creating either an AzureOpenAiStreamingChatModel or an AzureOpenAiEmbeddingModel meaning any call I made using those models fails with "404 resource not found".


geoand (approved):
Thanks a lot!

PR #236, opened 2024-01-23
Add watsonx module

andreadimaio:
In this PR I pushed a first version of the `watsonx` module.
For now the only feature is the `generation`, so no `StreamingChatModel` and `EmbeddingModel`.

One of the differences with the other modules is the way the `bearer token` is obtained, because there is an intermediate step to do, a rest api call to https://iam.cloud.ibm.com as described in the [documentation](https://cloud.ibm.com/apidocs/iam-identity-token-api#create-an-iam-access-token-for-a-user-or-service-i). 

The idea to solve this problem is to have a [ClientRequestFilter](https://github.com/andreadimaio/quarkus-langchain4j/blob/main/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/client/filter/BearerRequestFilter.java) attached to the [WatsonRestApi](https://github.com/andreadimaio/quarkus-langchain4j/blob/main/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/client/WatsonRestApi.java) to calculate the token, the business logic is in the [TokenGenerator](https://github.com/andreadimaio/quarkus-langchain4j/blob/main/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/TokenGenerator.java) class that caches the result and uses it until it expires.

Other differences are the `quarkus.langchain4j.watsonx.project-id` and `quarkus.langchain4j.watsonx.base-url` parameters that must be set to use the module.

Hope this module can be useful 😄


watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/client/WatsonRestApi.java
```diff
@@ -0,0 +1,198 @@
...
+            return new WatsonException(joiner.toString(), response.getStatus(), ex);
+
+        } catch (ClientWebApplicationException | ProcessingException e) {
+            return new WatsonException(e.getCause(), response.getStatus());
+        }
+    }
+
+    @ClientObjectMapper
+    static ObjectMapper objectMapper(ObjectMapper defaultObjectMapper) {
+        return QuarkusJsonCodecFactory.SnakeCaseObjectMapperHolder.MAPPER;
```
geoand, line 81: Does WatsonX really use snake case in it's JSON? If not, this should probably use camel case or whatever :)

watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/bean/IAMError.java
```diff
@@ -0,0 +1,15 @@
+package io.quarkiverse.langchain4j.watsonx.bean;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public record IAMError(
+
+        @JsonProperty("errorCode") Code errorCode,
+        @JsonProperty("errorMessage") String errorMessage) {
```
geoand, line 8: These should not be necessary if we configure the `ObjectMapper` to use camel case.

watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/client/IAMRestApi.java
```diff
@@ -0,0 +1,58 @@
...
+            }
+
+        } catch (ClientWebApplicationException | ProcessingException e) {
+            return new WebApplicationException(response);
+        }
+    }
+
+    @ClientObjectMapper
+    static ObjectMapper objectMapper(ObjectMapper defaultObjectMapper) {
+        return QuarkusJsonCodecFactory.SnakeCaseObjectMapperHolder.MAPPER;
```
geoand, line 56: Same question as for the other REST API

geoand (commented):
This is really cool, thanks!

I have added some comments about JSON handling

watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/client/WatsonRestApi.java
andreadimaio: Yes, this is an example of response from watsonx.
```
{
    "model_id": "meta-llama/llama-2-70b-chat",
    "created_at": "2024-01-23T13:12:46.370Z",
    "results": [
        {
            "generated_text": "...",
            "generated_token_count": 9,
            "input_token_count": 50,
            "stop_reason": "eos_token",
            "seed": 588258570
        }
    ],
 ...
 ```

watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/bean/IAMError.java
andreadimaio: The `IAMRestApi` uses the snake case for the "OK" response, but it is not true if there is an error.

geoand: Understood, thanks

andreadimaio:
> When you are done, can you please squash the commits into a single one?

Let's wait for CI. If all is green, I'm going to squash all the commits in `Add watsonx module`.

andreadimaio:
Squash done.

geoand (approved):
Thanks a lot!

Issue #175, opened 2023-12-20
Detect vector dimension mismatches in other embedding stores beyond Redis


jmartisk:
We did this for Redis in https://github.com/quarkiverse/quarkus-langchain4j/pull/172, others might use it too

PR #239, opened 2024-01-24
Various Pinecone improvements

jmartisk:
- Move to some new Pinecone REST endpoints with a slightly different API
- Fix creating a new index automatically (and add a configurable timeout for waiting for it to become ready)
- Make sure error details from HTTP error responses are added into the exception - this kinda resolves https://github.com/quarkiverse/quarkus-langchain4j/issues/175 for the Pinecone store

### DIFF
--- a/pinecone/runtime/src/main/java/io/quarkiverse/langchain4j/pinecone/PineconeEmbeddingStore.java
+++ b/pinecone/runtime/src/main/java/io/quarkiverse/langchain4j/pinecone/PineconeEmbeddingStore.java
@@ -59,7 +59,7 @@ public class PineconeEmbeddingStore implements EmbeddingStore<TextSegment> {
         this.indexName = indexName;
         this.dimension = dimension;
         String baseUrl = "https://" + indexName + "-" + projectId + ".svc." + environment + ".pinecone.io";
-        String baseUrlIndexOperations = "https://controller." + environment + ".pinecone.io";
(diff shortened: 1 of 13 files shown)


pinecone/runtime/src/main/java/io/quarkiverse/langchain4j/pinecone/PineconeEmbeddingStore.java
```diff
@@ -199,4 +206,20 @@ private void addAllInternal(List<String> ids, List<Embedding> embeddings, List<T
         Log.debug("Added embeddings: " + response.getUpsertedCount());
     }
 
+    private void waitForIndexToBecomeReady(String indexName, Duration timeout) {
```
geoand, line 209: Does smallrye fault tolerance have any utils we can use here?

jmartisk: Nothing coming to my mind, it's just a cycle that asks for the status until the status is ready.... 
At most we could use some retry logic on the individual REST client calls (which might be applicable to the whole client api, but we'd have to be very careful with retrying non-idempotent operations...)

geoand: 👌🏼

PR #247, opened 2024-01-25
Update release workflow

gastaldi:
- As needed in https://github.com/quarkiverse/quarkiverse-devops/issues/210

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,11 +30,11 @@ jobs:
 
       - uses: actions/checkout@v4
 
-      - name: Set up JDK 11
+      - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
           distribution: temurin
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -28,24 +28,19 @@ jobs:
...
-          passphrase: ${{ secrets.GPG_PASSPHRASE }}
-
-      - name: Set up JDK 17
-        uses: actions/setup-java@v3
+      - name: Set up JDK 11
+        uses: actions/setup-java@v4
         with:
           distribution: temurin
-          java-version: 17
+          java-version: 11
```
geoand, line 37: We don't want this, as the extension only works for Java 17

gastaldi: Updated

geoand (approved):
Thanks!

PR #251, opened 2024-01-27
Run integration tests in parallel

ppalaga:


ppalaga:
I am curious whether the actions/setup-java cache also covers the snapshots built by the needed jobs. If not, the individual native jobs will fail.

gsmet:
It's not recommended to use the cache to propagate artifacts from one job to another as the cache lookup/save can fail. It is recommended to upload/download an artifact instead. That's what we do in the Quarkus CI: the artifacts built in the Initial JDK 17 build are propagated to the other jobs.

ppalaga:
> It's not recommended to use the cache to propagate artifacts from one job to another as the cache lookup/save can fail. 

Thanks, good to know.

> It is recommended to upload/download an artifact instead. That's what we do in the Quarkus CI: the artifacts built in the Initial JDK 17 build are propagated to the other jobs.

We do the same in CQ and QCXF. Let me do that that way here too.

ppalaga:
14 mins - not bad.

geoand:
Very nice improvement, thanks a lot!

@gsmet can you confirm that this is the recommended way to go?

integration-tests/pom.xml
```diff
@@ -18,11 +18,11 @@
     <module>simple-ollama</module>
     <module>azure-openai</module>
     <module>devui</module>
-    <module>in-process-embedding-all-minilm-l6-v2-q</module>
-    <module>in-process-embedding-all-minilm-l6-v2</module>
-    <module>in-process-embedding-bge-small-en-q</module>
-    <module>in-process-embedding-bge-small-en</module>
-    <module>in-process-embedding-e5-small-v2-q</module>
-    <module>in-process-embedding-e5-small-v2</module>
+    <module>embed-all-minilm-l6-v2-q</module>
```
cescoffier, line 21: Why changing the module names?

cescoffier (approved):
I wonder why did you change the module names?

integration-tests/pom.xml
ppalaga: Just to match the common conventions. Is there a special reason to keep these special?

ppalaga:
Any feedback on this @gsmet? I am open to change anything.

... and I have the BOM PR waiting in the queue.

integration-tests/pom.xml
geoand: We used the names that are upsteam langchain4j uses, but since this is just about ITs, I don't think it matters much

cescoffier: yes, I think it's fine in this case.

.github/workflows/build-pull-request.yml
```diff
@@ -20,29 +20,119 @@ defaults:
...
+        run: |
+          tar -czf ${{ runner.temp }}/maven-repo.tgz -C ~ .m2/repository
+          # Avoid caching our own artifacts
+          rm -Rf ~/.m2/repository/io/quarkiverse/langchain4j
+      - name: Persist the Maven repo
+        uses: actions/upload-artifact@v4
+        with:
+          name: maven-repo
+          path: ${{ runner.temp }}/maven-repo.tgz
+          retention-days: 1
```
gsmet, line 61: I would advise to keep them for 5 days. I used to be very aggressive in Quarkus and then it became very unpractical when you want to rerun a specific subjob after a week-end.

.github/workflows/build-pull-request.yml
```diff
@@ -20,29 +20,119 @@ defaults:
...
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.build-and-test-jvm.outputs.matrix) }}
+    runs-on: ubuntu-latest
+
+    steps:
+
+      - name: Prepare git
+        run: git config --global core.autocrlf false
+        if: startsWith(matrix.os, 'windows')
```
gsmet, line 116: This looks odd given you run all your matrix on Ubuntu?

gsmet (commented):
Looks good in general.

I added one advice and one comment.

.github/workflows/build-pull-request.yml
ppalaga: I copied it from the old main job without too much thinking. Thanks for the point, removed.

.github/workflows/build-pull-request.yml
ppalaga: +1, increased to 5 days

ppalaga:
491a318:
* Rebased
* Addressed the suggestions from @gsmet

ppalaga:
Anybody has an idea what the deadlock in azure-openai test means? I have not seen anything like that before.

geoand:
Unfortunately that is a bug we have encountered in GraalVM since day 1 of this extension. 
@zakkak said he will look into it when time permits, but for now there is nothing we can do.

PR #265, opened 2024-02-02
Only pass Authorization/api-key headers when necessary

### LINKED ISSUE #264: Authorization header is not masked properly
I've only tested with the `openai` and `azure-openai` extensions, so I can't say whether or not this is present in the others, but it seems that if `logRequests` is set to `true`, the `Authorization` header is exposed in the log along with its value. This could leak the users api key.

The `api-key` header is masked. See screenshot (I've not captured the entire key so as to not leak mine :) )

image

edeandrea:
Only send the `Authorization` header for OpenAI requests and only send the `api-key` header on Azure OpenAI requests.

- Fixes #264


geoand (approved):
Thanks

PR #267, opened 2024-02-05
Ensure correct build order (samples AFTER extensions)

jmartisk:
I've seen some failures (during a parallel build) due to wrong build order - the `parent.relativePath` of samples has to point at an exact pom.xml, otherwise it doesn't make sure that all extensions are built before samples


geoand (approved):
Thanks!

jmartisk:
@gsmet I don't get an error from Maven even if I set it to complete nonsense (I do get an error in IntelliJ in both cases though). But yeah adding deployment artifacts is probably a good idea, I'll do it

gsmet:
I'm not sure if you get an error if your parent is in the local repo.

PR #272, opened 2024-02-06
Update Options.java

markusfindenig:
Parameter seed added


markusfindenig:
Parameter seed added

geoand:
Thanks for this!

For the change to be meaningful, `OllamaRecorder` and `Langchain4jOllamaConfig` should also be updated

markusfindenig:
Ok, i have it added

PR #275, opened 2024-02-07
Bump to LangChain4j 0.27.1

geoand:
In the future we'll also have to transition to using `retrievalAugmentor` instead of `retriever` in `@RegisterAiService`, but let's not hold up a new release for that

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,8 +36,8 @@
   </scm>
   <properties>
     <quarkus.version>3.7.2</quarkus.version>
-    <langchain4j.version>0.25.0</langchain4j.version>
-    <langchain4j-embeddings.version>0.25.0</langchain4j-embeddings.version>
+    <langchain4j.version>0.27.1</langchain4j.version>
+    <langchain4j-embeddings.version>0.27.1</langchain4j-embeddings.version>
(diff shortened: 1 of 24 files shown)


pom.xml
```diff
@@ -39,9 +39,9 @@
     <maven.compiler.release>17</maven.compiler.release>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <quarkus.version>3.7.0</quarkus.version>
-    <langchain4j.version>0.25.0</langchain4j.version>
-    <langchain4j-embeddings.version>0.25.0</langchain4j-embeddings.version>
+    <quarkus.version>3.7.1</quarkus.version>
+    <langchain4j.version>0.27.0-SNAPSHOT</langchain4j.version>
```
geoand, line 43: This needs to be changed to `0.26.2` when that release becomes available

pom.xml
```diff
@@ -39,9 +39,9 @@
...
+    <langchain4j-embeddings.version>0.26.1</langchain4j-embeddings.version>
```
geoand, line 44: Same here

PR #277, opened 2024-02-07
Add support for OpenAI response format

### LINKED ISSUE #274: Add support for the OpenAI response_format configuration
Hello folks

I often observed issues with a model not outputting proper json despite the prompt specifically asking for json ("you must return a valid json object, with no control characters" etc...).
This obviously can be mitigated by Retries but it remains a waste of token and API calls.

OpenAI offers a [response_format](https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format) on some of their models, allowing to force the model to output a specific format (mainly json at this stage).

I did some testing and it seems this setting fixes the issue. Langchain4j surfaces this setting (eg https://github.com/langchain4j/langchain4j/blob/b36315056ef6a4f4a021bd4cfcc97ac0d6c1a4de/langchain4j/src/test/java/dev/langchain4j/service/AiServicesIT.java#L249) and I'd like to make it accessible in the `ChatModelConfig` of this project.

I have [a working branch locally](https://github.com/PierreBtz/quarkus-langchain4j/commit/00904db478c6895d04f8430efb5df8e16fa704fe) but before I go ahead, polish it and open a PR I want to make sure with you it's something you want.

PierreBtz:
- Resolves #274

Not sure about a couple of things:

* tests: I looked into enriching the ITs but didn't find anything meaningful to assert...
* response format type: currently I went with a String. Langchain4J uses an enum `ResponseFormatType`. Not sure if I should surface the langchain4j enum to let langchain4j handle the maintenance :) Or use a new enum for this project? Let me know what you prefer.

### DIFF
--- a/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/ChatModelConfig.java
+++ b/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/ChatModelConfig.java
@@ -68,4 +68,11 @@ public interface ChatModelConfig {
      */
     @ConfigDocDefault("false")
     Optional<Boolean> logResponses();
+
+    /**
+     * The response format the model should use.
+     * Some models are not compatible with some response formats, make sure to review OpenAI documentation.
+     */
+    @WithDefault("text")
+    String responseFormat();
 }
--- a/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/ChatModelConfig.java
+++ b/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/ChatModelConfig.java
@@ -73,6 +73,5 @@ public interface ChatModelConfig {
      * The response format the model should use.
      * Some models are not compatible with some response formats, make sure to review OpenAI documentation.
      */
-    @WithDefault("text")
-    String responseFormat();
+    Optional<String> responseFormat();
 }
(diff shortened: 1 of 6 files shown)


openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/ChatModelConfig.java
```diff
@@ -68,4 +68,11 @@ public interface ChatModelConfig {
      */
     @ConfigDocDefault("false")
     Optional<Boolean> logResponses();
+
+    /**
+     * The response format the model should use.
+     * Some models are not compatible with some response formats, make sure to review OpenAI documentation.
+     */
+    @WithDefault("text")
```
geoand, line 76: Shouldn't we leave this unset if the user does not select something?

It seems to me that the default in LangChain4j is to not set a value

PierreBtz: Indeed, I went with the default as indicated by OpenAI doc, LangChain4j doesn't set anything.
How do you handle this scenario? `Optional<String>` then pass the property with something like `chatModelConfig.responseFormat().orElse(null)`? Or is there another pattern I'm not aware of?

geoand: Yeah, exactly that

geoand:
cc @cescoffier you will probably be interested in this as well

geoand (approved):

PR #279, opened 2024-02-08
Field seed in Class Options added

markusfindenig:
I have added the Field seed in the Class Options for Ollama. Because of that i can add a random int or a static int


ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/runtime/config/ChatModelConfig.java
```diff
@@ -47,4 +47,13 @@ public interface ChatModelConfig {
...
     @WithDefault("40")
     Integer topK();
+
+     /**
+     * With a static number the result is always the same. With a random number the result varies
+     * Example:
+     * Random random = new Random();
+     * int x = random.nextInt(Integer.MAX_VALUE);
+     */
+    @WithDefault("42")
```
geoand, line 57: Why this default?

geoand (commented):
Thanks for this, I have added a comment

markusfindenig:
Only a static number is required. The number can also be 0Am 08.02.2024 16:27 schrieb Georgios Andrianakis ***@***.***>:
@geoand commented on this pull request.

Thanks for this, I have added a comment

In ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/runtime/config/ChatModelConfig.java:
> @@ -47,4 +47,13 @@ public interface ChatModelConfig {
      */
     @WithDefault("40")
     Integer topK();
+
+     /**
+     * With a static number the result is always the same. With a random number the result varies
+     * Example:
+     * Random random = new Random();
+     * int x = random.nextInt(Integer.MAX_VALUE);
+     */
+    @WithDefault("42")

Why this default?

markusfindenig:
I think the default value is zero. I didn't find any. But also for the other values ​​such as top_k, the value is zero and must be set

markusfindenig:
I mean, there are no default values, because in langchain4j is only the model "Options" without any values
https://github.com/langchain4j/langchain4j/blob/main/langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/Options.java

geoand:
Cool. In that case we should make the configuration Optional and only set it the corresponding field when the Optional contains a value

markusfindenig:
now the standard value for seed is null

ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/runtime/config/ChatModelConfig.java
```diff
@@ -47,4 +47,12 @@ public interface ChatModelConfig {
...
     @WithDefault("40")
     Integer topK();
+
+     /**
+     * With a static number the result is always the same. With a random number the result varies
+     * Example:
+     * Random random = new Random();
+     * int x = random.nextInt(Integer.MAX_VALUE);
+     */
+    Integer seed();
```
geoand, line 57: Actually, this needs to be `Optional<Integer>` othewire the property will be required

geoand (commented):
Thanks for the update.

I added one more comment.

When that is addressed, we'll also need to squash the commits into a single one.

markusfindenig:
I hope this is correct now

ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/runtime/OllamaRecorder.java
```diff
@@ -17,11 +17,18 @@ public class OllamaRecorder {
     public Supplier<?> chatModel(Langchain4jOllamaConfig runtimeConfig, String modelName) {
         Langchain4jOllamaConfig.OllamaConfig ollamaConfig = correspondingOllamaConfig(runtimeConfig, modelName);
         ChatModelConfig chatModelConfig = ollamaConfig.chatModel();
+
+        int seed=null;
+        if(null!=chatModelConfig.seed()){
```
geoand, line 22: As `chatModelConfig.seed` will not always be an optional, it needs to be handled accordingly :)

geoand:
Great, can you please squash the commits and force push?

markusfindenig (commented):
ok

geoand:
We'll a rebase and squash, not a merge of main into this branch.

Do you want me to do it?

markusfindenig:
Yes, please do it. Am 11.02.2024 13:02 schrieb Georgios Andrianakis ***@***.***>:
We'll a rebase and squash, not a merge of main into this branch.
Do you want me to do it?

Issue #103, opened 2023-12-06
Infinispan as embedding storage support


karesti:
Infinispan is implementing the Vector Search capabilities. 
Tagging the issue here

geoand:
@karesti I am sure what your timeline for this is, but I just wanted to let you know that this extension releases much quicker than the uptream langchain4j project.

karesti:
@geoand we need 15.0.0.Final version in the infinispan-client upstream quarkus as well

karesti:
@geoand somewhere in january will be releasing 15.0.0.Final, then with the latest quarkus release containing the version, we can integrate all here. for now i'm working locally with snapshot and dev versions to fill all the gaps for the integration in langchain upstream and here

geoand:
Sounds good

PR #288, opened 2024-02-09
Infinispan embedding store

karesti:
we introduced a new method in 15 api that allows simply calling search, I can't use 14 client for that reason
On top of that, the jakarta dependency does not exist in 15 anymore, but in 14
However, when overriding all, the test and module I added works and builds (at least in JVM, did not try more for now)

Closes https://github.com/quarkiverse/quarkus-langchain4j/issues/103


karesti:
@maxandersen @geoand

geoand:
I'ld like @jmartisk to review this as well

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/runtime/InfinispanEmbeddingStoreConfig.java
```diff
@@ -0,0 +1,30 @@
...
+     * the embedding model that you use. For example, AllMiniLmL6V2QuantizedEmbeddingModel produces vectors of dimension 384.
+     * OpenAI's text-embedding-ada-002 produces vectors of dimension 1536.
+     */
+    Long dimension();
+
+    /**
+     * Name of the cache that will be used in Infinispan when searching for related embeddings.
+     * If this cache doesn't exist, it will be created.
+     */
+    @WithDefault("lanchain-cache")
```
cescoffier, line 27:
```suggestion
    @WithDefault("langchain-cache")
```

I'm actually wondering if langchain should be part of the name.

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/runtime/InfinispanEmbeddingStoreConfig.java
```diff
@@ -0,0 +1,30 @@
...
+@ConfigRoot(phase = RUN_TIME)
+@ConfigMapping(prefix = "quarkus.langchain4j.infinispan")
+public interface InfinispanEmbeddingStoreConfig {
+
+    /**
+     * The dimension of the embedding vectors. This has to be the same as the dimension of vectors produced by
+     * the embedding model that you use. For example, AllMiniLmL6V2QuantizedEmbeddingModel produces vectors of dimension 384.
+     * OpenAI's text-embedding-ada-002 produces vectors of dimension 1536.
+     */
+    Long dimension();
```
cescoffier, line 21: Open question: do we really need this? 

I found it super confusing when using Redis.

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/runtime/LanchainItemMarshaller.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.List;
+
+import org.infinispan.protostream.MessageMarshaller;
+
+import io.quarkiverse.langchain4j.infinispan.SchemaAndMarshallerProducer;
+
+/**
+ * Marshaller to read and write embeddings to Infinispan
+ */
+public class LanchainItemMarshaller implements MessageMarshaller<LangchainInfinispanItem> {
```
cescoffier, line 14:
```suggestion
public class LangchainItemMarshaller implements MessageMarshaller<LangchainInfinispanItem> {
```

infinispan/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,12 @@
+name: Langchain4j Infinispan embedding store
+artifact: ${project.groupId}:${project.artifactId}:${project.version}
+description: Provides the Infinispan Embedding store for Langchain4j
+metadata:
+  keywords:
+    - ai
+    - langchain4j
+    - openai
```
cescoffier, line 8: I would add `embedding`.

cescoffier (changes requested):

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/SchemaAndMarshallerProducer.java
```diff
@@ -0,0 +1,39 @@
...
+
+import io.quarkiverse.langchain4j.infinispan.runtime.InfinispanEmbeddingStoreConfig;
+import io.quarkiverse.langchain4j.infinispan.runtime.LanchainItemMarshaller;
+
+@ApplicationScoped
+public class SchemaAndMarshallerProducer {
+
+    public static final String LANCHAIN_ITEM = "LanchainItem";
+
+    private static final String PROTO = "syntax = \"proto2\";\n" + "\n" + "/**\n" + " * @Indexed\n" + " */\n"
```
jmartisk, line 19: Some other embedding stores allow configuring the name of the fields (in this case you have hardcoded `id`, `floatVector` and `text`). Wondering if it's actually useful and we should have it here too, maybe not.

In any case, we should also have a piece of documentation that describes the requirements on the schema of the cache - for the cases where the cache already exists on startup.

infinispan/pom.xml
```diff
@@ -0,0 +1,21 @@
...
+        <groupId>io.quarkiverse.langchain4j</groupId>
+        <artifactId>quarkus-langchain4j-parent</artifactId>
+        <version>999-SNAPSHOT</version>
+    </parent>
+    <artifactId>quarkus-langchain4j-infinispan-parent</artifactId>
+    <name>Quarkus LangChain4j - Infinispan embedding store - Parent</name>
+    <packaging>pom</packaging>
+
+    <properties>
+        <infinispan.version>15.0.0.Dev07</infinispan.version>
```
jmartisk, line 14: I understand it's temporary, but let's still put it into the root `pom.xml` along with all other version declarations?

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/InfinispanEmbeddingStore.java
```diff
@@ -0,0 +1,176 @@
...
+        }
+        // blocking call
+        remoteCache.putAll(elements);
+    }
+
+    @Override
+    public List<EmbeddingMatch<TextSegment>> findRelevant(Embedding referenceEmbedding, int maxResults,
+            double minScore) {
+        Query<Object[]> query = remoteCache.query("select i, score(i) from " + "LanchainItem" + schema.getDimension().toString()
+                + " i where i.floatVector <-> " + Arrays.toString(referenceEmbedding.vector()) + "~3");
```
jmartisk, line 129: What does the `~3` mean? Maximum vector distance or something? Should it be configurable rather than hardcoded?

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/runtime/InfinispanEmbeddingStoreConfig.java
karesti: Yes, the size of the vector is needed to define the index, and depending on the type of data/vector of your embeddings, the size can be diff. This is configurable in every embedding store in langchain

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/runtime/InfinispanEmbeddingStoreConfig.java
karesti: I'll rename to vector-cache

karesti: or embeddings-cache

karesti:
@jmartisk @cescoffier I reported the changes

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/InfinispanEmbeddingStore.java
```diff
@@ -0,0 +1,176 @@
...
+import io.quarkiverse.langchain4j.infinispan.runtime.LangchainInfinispanItem;
+
+public class InfinispanEmbeddingStore implements EmbeddingStore<TextSegment> {
+
+    private final RemoteCache<String, LangchainInfinispanItem> remoteCache;
+    private final InfinispanSchema schema;
+    private static final String DEFAULT_CACHE_CONFIG = "<distributed-cache name=\"CACHE_NAME\">\n"
+            + "<indexing storage=\"local-heap\">\n"
+            + "<indexed-entities>\n"
+            + "<indexed-entity>LANCHAINITEM</indexed-entity>\n"
```
jmartisk, line 36: G missing

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/runtime/InfinispanEmbeddingStoreConfig.java
jmartisk: But please make it `Optional<Long>` so it's not required in cases when the schema already exists. If then, during boot, you find out that the cache needs to be created, and the dimension wasn't specified, then you throw an error.

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/runtime/InfinispanEmbeddingStoreConfig.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Name of the cache that will be used in Infinispan when searching for related embeddings.
+     * If this cache doesn't exist, it will be created.
+     */
+    @WithDefault("embeddings-cache")
+    String cacheName();
+
+    /**
+     * The distance between the
```
jmartisk, line 31: Missing part of the description :)

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/runtime/InfinispanEmbeddingStoreConfig.java
karesti: ok, let's do that

karesti: the dimension is also used to know which index type is needed. it's a but more complex than that, I need to grab the cache configuration and infer the indexed entity name.
I decided the name for this use case, that could be parametrable

karesti: ok, I can't do this easily. Grabbing the configuration of a cache, for security reasons, is not an available method from the client. This action is only for "admins" and the user of this api might not have those rights.
What I can do is either specifying the vector size, either the indexed entity name

karesti: ok, right now dimension is mandatory. Schemas need to exist in the server and the client. I can't make it "optional" for this use case.

karesti: IMO, at least for now, schemas for vectors need to follow an specific convection in Infinispan. Caches can exist or not. The dimension is on the schema.

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/SchemaAndMarshallerProducer.java
karesti: I did nothing on documentation yet, yes this must be explained.

karesti: @jmartisk concerning the naming, we can adapt the naming but I don't see the benefit right now.

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/runtime/InfinispanEmbeddingStoreConfig.java
jmartisk: Ah, so you're saying that the client requires to have its own copy of the schema definition (and it can't ask the server to provide it for an existing cache) and that's why the client always needs to have the dimension configured?
What happens if the cache exists on the server, but the client has a WRONG dimension configured?

karesti: The serialization is done with protostream (protobuf) and hibernate search.
If the cache exists with LangchainItem384 (schema on client and server)
But then I try the existing cache with items on 384 with another dimension (383)
The server will upload a new schema for that size, register on the client the serialization context for that dimension and then Hibernate Search will say: 

```org.infinispan.client.hotrod.exceptions.HotRodClientException:Request for messageId=15 returned server error (status=0x85): org.hibernate.search.util.common.SearchException: ISPN014505: Unknown entity name: 'LangchainItem383'.```

Comment (karesti) [reply infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/runtime/InfinispanEmbeddingStoreConfig.java:21]: this is mandatory in the cache config for this use case
```json
"indexing": {
        "enabled": true,
        "storage": "local-heap",
        "indexed-entities": [
          "LangchainItem384"
        ]
      },
```
They need to match

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/SchemaAndMarshallerProducer.java
```diff
@@ -0,0 +1,39 @@
...
+import org.infinispan.protostream.FileDescriptorSource;
+import org.infinispan.protostream.MessageMarshaller;
+
+import io.quarkiverse.langchain4j.infinispan.runtime.InfinispanEmbeddingStoreConfig;
+import io.quarkiverse.langchain4j.infinispan.runtime.LangchainItemMarshaller;
+
+@ApplicationScoped
+public class SchemaAndMarshallerProducer {
+
+    public static final String LANCHAIN_ITEM = "LangchainItem";
```
jmartisk, line 17: There's still a missing G :)

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/runtime/InfinispanEmbeddingStoreConfig.java
jmartisk: Ok thanks for explaining, then it makes sense to me

karesti: We can improve the error message to be more clear on this particular integration, however things can evolve and get even better from our side. for 15 a bunch of things are coming like protobuf 3 integration and some API improvements and capabilities. The code I did here, will still work, but will be improved

infinispan/runtime/src/main/java/io/quarkiverse/langchain4j/infinispan/SchemaAndMarshallerProducer.java
karesti: forgot to push that change :) crazy G xD

jmartisk (approved):

cescoffier (approved):

PR #291, opened 2024-02-12
Ability to disable integrations

### LINKED ISSUE #130: Ability to serve a default response?
Since integrating with OpenAI costs real money, should there be a way that I can serve a default response if I want to "turn off" the integration at runtime? Maybe when I'm running in dev mode or if I'm doing a quick demo I don't want it to make "real" calls to an openai provider.

edeandrea:
This provides configuration around langchain4j/langchain4j#549 which will create disabled versions of the models based on configuration.

- Closes: #130

### DIFF
--- a/bam/runtime/pom.xml
+++ b/bam/runtime/pom.xml
@@ -22,6 +22,22 @@
             <artifactId>quarkus-langchain4j-core</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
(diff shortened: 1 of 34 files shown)


geoand:
This will need a rebase and revert of the docs changes as they have been updated by another PR

edeandrea:
Should the docs not get updated at all? Or just re-updated via `mvn install` after a rebase?

geoand:
This time they should not get updated as I already updated them

edeandrea:
@geoand I've rebased. This should be good to go once the checks run.

geoand:
How about some docs on why/when to use it?

edeandrea:
> How about some docs on why/when to use it?

Since this is cross-cutting across all the modules, where do you think the best place for that kind of doc is? Something new under advanced topics?

geoand:
> Something new under advanced topics?

Sounds good

edeandrea:
I've added some documentation. Let me know what you think.

geoand (approved):
Lovely!

PR #292, opened 2024-02-13
Add user-agent to Azure AI requests

geoand:
This is the same as upstream LangChain4j does and was requested my MS in order to be able to determine the use of LangChain4j


cescoffier (approved):
Should it be configurable or it needs to be a constant?

geoand:
I was told it should be constant.

PR #293, opened 2024-02-13
Unify on using EmbeddingStoreIT as test base for embedding stores

jmartisk:
Reuse the `EmbeddingStoreIT` class that kinda serves as a TCK for embedding stores.
This removes a lot of code duplication and brings us a bit closer to the upstream langchain4j project.

I also had to add the `deleteAll` functionality for the Chroma store, because the testcase requires that each test starts with a clean store.

### DIFF
--- a/chroma/deployment/pom.xml
+++ b/chroma/deployment/pom.xml
@@ -36,6 +36,14 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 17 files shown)


geoand (approved):
Nice!

I was thinking of doing something similar with the Wiremock stuff we have, but that's for another day...

PR #300, opened 2024-02-14
Include Infinispan configuration docs

karesti:
I realized that the wrong import is in place


karesti:
@geoand I also changed the main title that skipped my eyes

PR #302, opened 2024-02-15
Automatically set the image name for embedding stores with dev services

geoand:


geoand:
Also cc @karesti

docs/modules/ROOT/pages/redis-store.adoc
```diff
@@ -19,8 +19,6 @@ To utilize the Redis document store, you'll need to include the following depend
 
 This extension relies on the Quarkus Redis client. Ensure the default Redis datasource is configured appropriately. For detailed guidance, refer to the link:https://quarkus.io/guides/redis[Quarkus Redis Quickstart] and the link:https://quarkus.io/guides/redis-reference[Quarkus Redis Reference].
 
-IMPORTANT: The Redis document store's functionality is built on the Redis JSON and Redis Search modules. Ensure these modules are installed, or consider using the Redis Stack. If you want to use the Quarkus Dev Services for Redis, make sure to configure `quarkus.redis.devservices.image-name=redis/redis-stack:latest` in your `application.properties` file.
```
jmartisk, line 22: I think at least the first ~half of this note is still relevant to avoid surprises.
And the presence of the `langchain4j` extension now changes stuff from the Redis extension (overrides the property), so perhaps this should be mentioned, along with saying which property controls it.

geoand: Good point. 

I updated the PR, WDYT?

jmartisk (approved):

PR #317, opened 2024-02-21
Don't show chat card unless a model is present

geoand:


jmartisk:
I'm a bit confused, what is the difference between `ChatModelProviderCandidateBuildItem` and `SelectedChatModelProviderBuildItem`? "Selected" would hint at being a single model that is selected, but then why is it a `MultiBuildItem`...

Issue #315, opened 2024-02-21
Improve the error message in the DevUI when the call times out


geoand:
When for example an image generation request from the DevUI times out, we get this error:

![Screenshot from 2024-02-21 11-46-05](https://github.com/quarkiverse/quarkus-langchain4j/assets/4374975/63937de3-7fef-4b68-a45f-f82a82d04c8b)

The exception thrown is:

```posh
2024-02-21 11:44:10,150 ERROR [io.qua.dev.run.jso.JsonRpcCodec] (executor-thread-1) Error in JsonRPC Call: java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
        at io.smallrye.mutiny.unchecked.UncheckedSupplier.lambda$toSupplier$0(UncheckedSupplier.java:45)
        at io.smallrye.context.impl.wrappers.SlowContextualSupplier.get(SlowContextualSupplier.java:21)
        at io.smallrye.mutiny.operators.uni.builders.UniCreateFromItemSupplier.subscribe(UniCreateFromItemSupplier.java:28)
        at io.smallrye.mutiny.operators.AbstractUni.subscribe(AbstractUni.java:36)
        at io.smallrye.mutiny.operators.uni.UniRunSubscribeOn.lambda$subscribe$0(UniRunSubscribeOn.java:27)
        at io.quarkus.vertx.core.runtime.VertxCoreRecorder$14.runWith(VertxCoreRecorder.java:582)
        at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2513)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1538)
        at org.jboss.threads.DelegatingRunnable.run(DelegatingRunnable.java:29)
        at org.jboss.threads.ThreadLocalResettingRunnable.run(ThreadLocalResettingRunnable.java:29)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:118)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
[...]

PR #318, opened 2024-02-21
Devui chat error handling

jmartisk:
- Show the exception's message if the chat request fails
- To be safer in avoiding inconsistencies after errors, the backend now always sends the full current chat history to the frontend (before this change, the frontend "maintains" its own copy of the history, but it might diverge from the correct one due to errors or race conditions)

If this makes sense, I'll see where else to also apply similar fixes (the embedding store, image and moderation pages..)

 - Fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/315


geoand (approved):
Thanks!

jmartisk:
Btw, the reason why originally we lost the actual exception message is this: 
https://github.com/quarkusio/quarkus/pull/38953
(This PR won't really change it now, because I've reworked the Langchain4j service to pass the message as a custom field in the regular response)

geoand:
🙏🏼

Issue #322, opened 2024-02-22
Require`quarkus.langchain4j.openai.api-key` when starting the app


gastaldi:
I created an application using `quarkus create app --extension=langchain4j-openai` then I ran `quarkus dev` and browsed to http://localhost:8080/q/dev-ui/io.quarkiverse.langchain4j.quarkus-langchain4j-openai/images.

Because I forgot to set the `quarkus.langchain4j.openai.api-key`, I get this strange error in the UI:

<img width="2262" alt="Screenshot 2024-02-22 at 13 38 35" src="https://github.com/quarkiverse/quarkus-langchain4j/assets/54133/99d2899e-7b5b-4412-b617-6ca22c180e51">

I see this in the logs: 
```
Caused by: dev.ai4j.openai4j.OpenAiHttpException: {
  "error": {
    "code": "invalid_api_key",
    "message": "Incorrect API key provided: dummy. You can find your API key at https://platform.openai.com/account/api-keys.",
    "param": null,
    "type": "invalid_request_error"
  }
} 
```

PS: This is also shown after I set my key (because I reached my billing limit):

```
Caused by: dev.ai4j.openai4j.OpenAiHttpException: {
  "error": {
    "code": "billing_hard_limit_reached",
    "message": "Billing hard limit has been reached",
    "param": null,
    "type": "invalid_request_error"
  }
}
```

geoand:
Ah, we need to surface that error earlier

jmartisk:
By the way, the useless error message (`InvocationTargetException`... stuff) will be improved by https://github.com/quarkusio/quarkus/pull/38953, the error will contain the actual exception class and message.

gastaldi:
@jmartisk that's possible, I have zero classes in my `src/main/java`

geoand:
> But I wonder why I get this error right during boot, while you don't:
> 
> ```
> SRCFG00014: The config property quarkus.langchain4j.openai.api-key is required but it could not be found in any config source
> ```
> 
> I guess it's because my app actually directly uses (injects) a chat model so it gets initialized eagerly

That's what I thought should have happened

jmartisk:
Aha, this is the problem
The property has a default value of "dummy": https://github.com/quarkiverse/quarkus-langchain4j/blob/0.8.3/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/Langchain4jOpenAiConfig.java#L50

[...]

PR #330, opened 2024-02-27
Adapt OpenAI Dev UI pages to multiple models

jmartisk:
- Fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/322 : for the moderation and image pages, when using the default model and the API key is missing, we show a nicer error now. For other pages, it's not necessary because they inject the model directly from CDI and thus if the API key is missing, it will have failed earlier anyway.
- Moderation and image pages now support selecting the named OpenAI model configuration, which will serve as the "base", on top of which you can override parameters using dropdowns
- I'd like to extend the embedding store and chat pages to support multiple models too, but haven't figured out yet how to do it nicely, will revisit

CC @gastaldi


geoand (approved):
Thanks!

Issue #326, opened 2024-02-26
Retriever is deprecated in favor of ContentRetriever


karesti:
Use ContentRetriever instead of Retriever

```
/**
 * Interface for retrieving relevant items.
 * This class is deprecated, use {@link ContentRetriever} instead.
 *
 * @param <T> the type of the items.
 */
@Deprecated
public interface Retriever<T> {

```

Change in favor of ContentRetriever
```
public @interface RegisterAiService {
```

geoand:
See https://github.com/quarkiverse/quarkus-langchain4j/pull/275#issue-2123310352

karesti:
@geoand working on it

geoand:
@karesti see https://github.com/quarkiverse/quarkus-langchain4j/pull/338 by @jmartisk

karesti:
oh ok xD

PR #338, opened 2024-02-29
Advanced RAG

jmartisk:
- deprecates usage of `retriever` for AI Services and adds `retrievalAugmentor` instead
- adds a new integration test module for testing RAG-related stuff
- updates RAG-related samples to use a `retrievalAugmentor`
- updates the docs a little, but there's probably no point in repeating all details about advanced RAG strategies when they are in Langchain4j docs and examples?!

* Fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/332

CC @langchain4j


core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -230,6 +231,29 @@ public void findDeclarativeServices(CombinedIndexBuildItem indexBuildItem,
...
+                if (NO_RETRIEVAL_AUGMENTOR.equals(retrieverClassDotName)) {
+                    retrievalAugmentorSupplierClassDotName = null;
+                } else {
+                    BuiltinScope declaredScope = BuiltinScope
+                            .from(index.getClassByName(retrievalAugmentorSupplierClassDotName));
+                    if (declaredScope == null) {
+                        throw new IllegalConfigurationException(
+                                "Unable to determine the CDI scope of " + retrievalAugmentorSupplierClassDotName
+                                        + ". Is it a CDI bean?");
+                    }
```
geoand, line 247: Wouldn't it be better to use the pattern we have for other types like `BeanChatMemoryProviderSupplier` in order to not have to force users to create a bean of type `Supplier<RetrievalAugmentor>`, but what they actually need which is a `RetrievalAugmentor`?

core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
```diff
@@ -88,11 +89,21 @@
...
      */
+    @Deprecated(forRemoval = true)
     Class<? extends Retriever<TextSegment>> retriever() default NoRetriever.class;
 
+    /**
+     * Configures the way to obtain the {@link RetrievalAugmentor} to use (when using RAG).
+     * The Supplier is expected to be a CDI bean.
+     * By default, no retrieval augmentor is used.
+     */
+    Class<? extends Supplier<RetrievalAugmentor>> retrievalAugmentor() default NoRetrievalAugmentor.class;
```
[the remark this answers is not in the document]
jmartisk: For `Retriever`, the typical usage was creating a custom implementation of the `Retriever` interface, which is nicely doable by having the user provide a CDI bean that implements `Retriever`. 

But for `RetrievalAugmentor`, the typical usage is to use the builder of `DefaultRetrievalAugmentor` to build an appropriately configured instance of it. We don't want the user to create a class that extends `RetrievalAugmentor`, we want them to use the builder to build an instance of `DefaultRetrievalAugmentor`.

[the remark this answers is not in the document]
geoand: Then in that case I would much prefer to use the approach we have with `moderationModelSupplier` (and others).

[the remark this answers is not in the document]
jmartisk: That is pretty much what I did here, no? Or what exactly should be different?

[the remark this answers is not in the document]
geoand: `moderationModelSupplier` is more flexible because:

It provides `BeanModerationModelSupplier` which essentially looks up the bean, but it also allows for an arbitraty supplier that does not need to be a bean (for advanced cases)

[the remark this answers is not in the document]
jmartisk: Hmm, I'm still not completely convinced. You mean that 
- if the supplier class a CDI bean, then inject it from CDI
- if it's not a CDI bean, then call the no-arg constructor?

In 99% of cases, you need the supplier to be a bean, because you can't build a reasonable augmentor instance without injecting an embedding store and model. That way we would open up a doorway to user errors where they have injection points in their `Supplier<RetrievalAugmentor>`, but we call the constructor instead of using it as a CDI bean.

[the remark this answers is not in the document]
geoand: The is the following:

* `BeanRetrievalAugmentorSupplier` would automatically look use whatever `BeanRetrievalAugmentor` CDI then is.
* Any other class would be a customer supplier would allow users to return whatever `BeanRetrievalAugmentor` they like from `get`. This class may or may not be a CDI bean.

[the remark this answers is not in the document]
jmartisk: Maybe we're misunderstanding each other, I see no reason why the supplier implementation should be allowed to NOT be a bean :)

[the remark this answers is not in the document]
jmartisk: Anyway I've still pushed an update where it is allowed to not be a bean, so please check it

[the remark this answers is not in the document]
geoand: Thanks, I'll have a look tomorrow

docs/modules/ROOT/examples/io/quarkiverse/langchain4j/samples/RetrievalAugmentorExample.java
```diff
@@ -0,0 +1,35 @@
...
+
+import jakarta.enterprise.context.Dependent;
+
+import dev.langchain4j.model.embedding.EmbeddingModel;
+import dev.langchain4j.rag.DefaultRetrievalAugmentor;
+import dev.langchain4j.rag.RetrievalAugmentor;
+import dev.langchain4j.rag.content.retriever.EmbeddingStoreContentRetriever;
+import io.quarkiverse.langchain4j.redis.RedisEmbeddingStore;
+
+@Dependent
```
geoand, line 13: Is there a reason why this is `@Dependent`? Wouldn't `@Singleton` be just fine?

core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
```diff
@@ -88,11 +89,22 @@
...
+    @Deprecated(forRemoval = true)
     Class<? extends Retriever<TextSegment>> retriever() default NoRetriever.class;
 
+    /**
+     * Configures the way to obtain the {@link RetrievalAugmentor} to use (when using RAG).
+     * The Supplier is expected to be a CDI bean.
+     * If unspecified, Quarkus will attempt to locate a CDI bean that implements {@link RetrievalAugmentor}
+     * and use it if one exists.
+     */
+    Class<? extends Supplier<RetrievalAugmentor>> retrievalAugmentor() default BeanIfExistsRetrievalAugmentorSupplier.class;
```
geoand, line 106: 🙏🏼

integration-tests/rag/src/main/java/org/acme/example/AiServiceWithQueryTransformer.java
```diff
@@ -0,0 +1,46 @@
...
+public interface AiServiceWithQueryTransformer {
+
+    String chat(String message);
+
+    /**
+     * Contains a query transformer that transforms the query text to
+     * lowercase. If the transformed worked properly, the content retriever
+     * will append content saying "The transformer works!".
+     */
+    @Dependent
```
geoand, line 26: Same question as earlier

integration-tests/rag/src/main/java/org/acme/example/AiServiceWithSpecifiedRetrievalAugmentor.java
```diff
@@ -0,0 +1,43 @@
...
+import dev.langchain4j.rag.content.retriever.EmbeddingStoreContentRetriever;
+import dev.langchain4j.store.embedding.inmemory.InMemoryEmbeddingStore;
+import io.quarkiverse.langchain4j.RegisterAiService;
+
+@RegisterAiService(retrievalAugmentor = AiServiceWithSpecifiedRetrievalAugmentor.NaiveRagAugmentor.class)
+public interface AiServiceWithSpecifiedRetrievalAugmentor {
+
+    String chat(String message);
+
+    @Dependent
```
geoand, line 22: And here

geoand:
Very nice @jmartisk! 

@cescoffier @langchain4j could you please have a look at the docs and examples?

Thanks

karesti:
@jmartisk why aren't we using contentRetrieval as well ?

karesti:
I opened this issue, wondering if should be closed then https://github.com/quarkiverse/quarkus-langchain4j/issues/326

jmartisk:
@karesti Retrieval augmentors typically contain content retrievers (among other things) so I think it's not necessary to also support standalone content retrievers. Or should we? I'm not sure

jmartisk:
Hmm, OK, given that langchain4j's AI service implementation also supports specifying a standalone content retriever (and it wraps it into a retrieval augmentor automatically) I think we can add that on our side too. I'll look into it

karesti:
@jmartisk I have a work in progress locally that replaces simply retrieval by content retrieval (deprecating retrieval)

karesti:
@jmartisk https://github.com/karesti/quarkus-langchain4j/tree/content-retrieval

jmartisk:
@karesti Ok then I guess I can just leave it to you and you'll send a follow-up PR :)

Please consider whether we should have the bean providing the content retriever as a `Supplier<ContentRetriever>` instead of just `ContentRetriever`, because the typical usage is, just like with `RetrievalAugmentor`, to create a configured instance of a class that implements `ContentRetriever`. I see that in your code, you basically have to "wrap" an instance of the actual retriever inside a custom object that also implements it (https://github.com/karesti/quarkus-langchain4j/commit/ad7df344f73c13526bba6d7de79d3b06e6e7768d#diff-0e4e6053292ca1aed882a051f674e71ee8f249a33278838fd6a404f528872210R17) which is a bit clumsy IMO, maybe it should follow the same pattern that I have for `Supplier<RetrievalAugmentor>` here. But let's have @cescoffier and @langchain4j tell us their opinion

karesti:
@jmartisk my branch is very simple now since I was just doing the deprecate code replacement. I'll follow the standard

docs/modules/ROOT/examples/io/quarkiverse/langchain4j/samples/RetrievalAugmentorExample.java
```diff
@@ -0,0 +1,35 @@
...
+@Singleton
+public class RetrievalAugmentorExample implements Supplier<RetrievalAugmentor> {
+
+    private final RetrievalAugmentor augmentor;
+
+    RetrievalAugmentorExample(RedisEmbeddingStore store, EmbeddingModel model) {
+        EmbeddingStoreContentRetriever contentRetriever = EmbeddingStoreContentRetriever.builder()
+                .embeddingModel(model)
+                .embeddingStore(store)
+                .maxResults(20)
```
cescoffier, line 22: That's a lot of document - this will exceed context size. Can you change it to something more realistic like 3.

cescoffier (approved):

geoand:
Merging, we can certainly improve with this in

docs/modules/ROOT/pages/retrievers.adoc
```diff
@@ -56,18 +56,36 @@ For instance, for CSV files with document representation separated by `\n`, `new
...
+
+The example above is a CDI bean that implements
+`Supplier<RetrievalAugmentor>`. An alternative way to wire things up is to
+create a CDI bean that directly implements `RetrievalAugmentor`, for example
+via a CDI producer, and letting Quarkus auto-discover it (by not specifying
+the `retrievalAugmentor` parameter of the `@RegisterAiService` annotation).
+
+The EmbeddingStoreContentRetriever necessitates a configured embedding store
+(Redis, Chroma, etc.) and an embedding model. Configure the maximum number
+of documents to retrieve (e.g., 20 in the example) and set the minimum
```
langchain4j, line 84: Should it be 3 instead of 20?

docs/modules/ROOT/pages/retrievers.adoc
```diff
@@ -56,18 +56,36 @@ For instance, for CSV files with document representation separated by `\n`, `new
...
+relevance score if required.
 
 Make sure that the number of documents is not too high (or document too large).
 More document you have, more data you are adding to the LLM context, and you may exceed the limit.
```
langchain4j, line 88: To address this concern, I was thinking to extend existing `ContentAggregator` implementations to (optionally) accept a `Tokenizer` and something like `maxTokens`. In this case it will be able to pass only most relevant documents/segments up to that `maxTokens` value. This feature can also be added in `EmbeddingStoreContentRetriever`, but if multiple `ContentRetriever`s are used, then it makes more sense to have it in the aggregator, WDYT?

samples/chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/AugmentorExample.java
```diff
@@ -0,0 +1,32 @@
...
+        retriever = EmbeddingStoreContentRetriever.builder()
+                .embeddingModel(model)
+                .embeddingStore(store)
+                .maxResults(20)
+                .build();
+    }
+
+    @Override
+    public RetrievalAugmentor get() {
+        return DefaultRetrievalAugmentor.builder()
```
langchain4j, line 28: This was probably already proposed, but would it make sense to allow user to define just a `ContentRetriever` bean and it will be automatically wrapped into `DefaultRetrievalAugmentor` if there are no `RetrievalAugmentor` beans anywhere? Not sure it fits the Quarkus style?

docs/modules/ROOT/pages/retrievers.adoc
jmartisk: Ah, good catch, I'll update that.

samples/chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/AugmentorExample.java
jmartisk: Yes this was discussed with @karesti  in the comments in this PR, she's going to send a PR for that

docs/modules/ROOT/pages/retrievers.adoc
jmartisk: Yes, that might be useful when there's a lot of content retrievers

samples/chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/AugmentorExample.java
karesti: I don't know if I'm understanding well enough, I'll pr a draft to check

Issue #331, opened 2024-02-27
Concurrent issue with InMemoryChatMemoryStore


andreadimaio:
I found a problem when I run two concurrent API calls due to the use of the `InMemoryChatMemoryStore` with the memory id `default` in the [AiServiceMethodImplementationSupport](https://github.com/quarkiverse/quarkus-langchain4j/blob/7dae65702297764444ff6f35ca082560e883b52e/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java#L100) class.

```java
Object memoryId = memoryId(createInfo, methodArgs).orElse("default");
```

From my initial analysis, the problem occurs when two threads execute the code that goes [from line 114 to line 129](https://github.com/quarkiverse/quarkus-langchain4j/blob/7dae65702297764444ff6f35ca082560e883b52e/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java#L114-L129).

```java
if (context.hasChatMemory()) {
    ChatMemory chatMemory = context.chatMemory(memoryId);
    if (systemMessage.isPresent()) {
        chatMemory.add(systemMessage.get());
    }
    chatMemory.add(userMessage);
}

List<ChatMessage> messages;
if (context.hasChatMemory()) {
    messages = context.chatMemory(memoryId).messages();
} else {
    messages = new ArrayList<>();
    systemMessage.ifPresent(messages::add);
    messages.add(userMessage);
}
```

Using the same memory id (`default`) can lead to anomalous behavior in ChatMessage list objects, like the one I show you below.
I have an interface called `WatsonxService` annotated with `@RegisterAiService`:

```java
@SystemMessage("You are a poet")
@UserMessage("Write a poem of 5 lines about {topic}")
public String poemSync(String topic);
```

This service is exposed as endpoint at `/poem`:

```java
@GET
@Path("/")
public String watsonxPoemSync(@QueryParam("el") String el) {
     return watsonxService.poemSync(el);              
}
```

Now, when I run two curl commands in parallel
```
> curl http://localhost:8080/poem?el=dog
> curl http://localhost:8080/poem?el=cat
```
These are the results from the request logs:

```
2024-02-27 08:55:08,431 INFO  [io.qua.lan.wat.cli.WatsonxRestApi$WatsonClientLogger] (vert.x-eventloop-thread-3) Request:
- method: POST
[...]

PR #339, opened 2024-02-29
Provide a smarter default as memoryId when possible

geoand:
The idea here is to try and use the request context as the default in order to avoid getting messages mixed up when concurrent requests are made and no explicit memoryId is being used

Fixes: #331


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -309,11 +312,22 @@ private static Object transformTemplateParamValue(Object value) {
...
+            return methodArgs[createInfo.getMemoryIdParamPosition().get()];
         }
-        return Optional.empty();
+        if (hasChatMemoryProvider) {
+            // first we try to use the current context in order to make sure that we don't interleave chat messages of concurrent requests
+            ArcContainer container = Arc.container();
+            if (container != null) {
+                ManagedContext requestContext = container.requestContext();
+                if (requestContext.isActive()) {
+                    return requestContext;
```
gsmet, line 325: Stupid question: what's your contract for the memory id? Does it use equals/hashCode? If so does the requestContext implement it properly?

Do you store the memoryId somewhere? Would it be a problem to keep the request context around if so?

Probably questions you already answered but I couldn't help thinking about them when reading the patch.

geoand: > what's your contract for the memory id?

It can be any object

 > If so does the requestContext implement it properly?

It does not implement them at all

> Do you store the memoryId somewhere? Would it be a problem to keep the request context around if so?

Those exact questions plagued us for a while, but they have since been sorted out with the proper integration between the scope of the AiService and the lifecycle of the request.

> Probably questions you already answered but I couldn't help thinking about them when reading the patch.

👍🏼

PR #346, opened 2024-03-04
Update extension metadata to use name consistent with LangChain4j

holly-cummins:
I notice that https://quarkus.io/extensions/io.quarkiverse.langchain4j/quarkus-langchain4j-azure-openai/ (and others) refers to 'Langchain4j' but https://github.com/langchain4j/langchain4j calls it 'LangChain4j'. 

We should probably be consistent, so I've updated the metadata. Our `pom.xml`s already use 'LangChain4j', so I didn't need to change those. 

All of the Java code uses `Langchain4j`, but changing that is more invasive, and could affect external APIs, so I've left it alone for now.

### DIFF
--- a/.github/workflows/build-against-langchain4j.yml
+++ b/.github/workflows/build-against-langchain4j.yml
@@ -1,4 +1,4 @@
-name: Build against Langchain4j main
+name: Build against LangChain4j main
 
 on:
   schedule:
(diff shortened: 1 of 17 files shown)


geoand (approved):
Thanks!

holly-cummins:
@geoand do you have a thought on changing things like `io.quarkiverse.langchain4j.deployment.Langchain4jDotNames` and `Langchain4jHuggingFaceConfig`?

geoand:
Go ahead, no real reason not to change them :)

PR #348, opened 2024-03-05
Rework csv-chatbot to use a SQL database

jmartisk:
Currently, it uses Hibernate for inserting the data, I'm not completely sure we want that... But the code then looks nicer than raw JDBC usage, so perhaps keep it?


geoand:
> But the code then looks nicer than raw JDBC usage, so perhaps keep it?

:cough: StatelessSession :cough: :)

jmartisk:
Uhmm, swapped `EntityManager` for `StatelessSession` :)

samples/csv-chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/MovieDatabaseContentRetriever.java
```diff
@@ -0,0 +1,61 @@
...
+                        results.add(Content.from(json.toString()));
+                    }
+                }
+            }
+        } catch (SQLException e) {
+            throw new RuntimeException(e);
+        }
+        return results;
+    }
+}
```
geoand, line 61: Can't we use the `StatelessSession` here? It should cut down on the code

jmartisk: I don't think so, the problem is that we don't have that much control over the SQL query that the LLM creates.  We don't know which columns of the table will be in the SELECT clause and the result potentially might not be mappable via ORM to a `Movie` instance at all.

jmartisk: So instead of mapping it via ORM, we just take the whole ResultSet, along with column names from its metadata, and transform it blindly into a JSON object.

geoand: Gotcha

samples/csv-chatbot/src/main/resources/META-INF/resources/components/demo-title.js
```diff
@@ -64,7 +64,8 @@ export class DemoTitle extends LitElement {
             <div class="explanation">
                 <ol>
                     <li>The user send a question.</li>
-                    <li>The application looks for relevant data in the store.</li>
+                    <li>The application asks the LLM to create a SQL query that retrieves the relevant data from the database.</li>
+                    <li>The application looks executes the SQL query.</li>
```
geoand, line 68:
```suggestion
                    <li>The application executes the SQL query.</li>
```

samples/csv-chatbot/src/main/resources/META-INF/resources/components/demo-title.js
```diff
@@ -54,7 +54,7 @@ export class DemoTitle extends LitElement {
                 This demo shows how to build a chatbot powered by GPT 3.5 and retrieval augmented generation.
```
geoand, line 54:
```suggestion
                This demo shows how to build a chatbot powered by GPT 3.5 and advanced retrieval augmented generation.
```

jmartisk:
Amazing. I can even ask questions like 
```
What's the average rating of movies in the database?
```
and it executes
```
SELECT AVG(imdb_rating) as average_rating
FROM movie;
```
and gives the correct answer. This wouldn't be possible with embeddings at all.

geoand:
That's awesome!

To be honest I am not completely susprised because from what I have been reading, LLMs have been trained on a ton of SQL

PR #353, opened 2024-03-05
Content Retrieval

karesti:
I don't understand this well enough

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
@@ -15,6 +15,7 @@ import dev.langchain4j.memory.chat.MessageWindowChatMemory;
 import dev.langchain4j.model.chat.ChatLanguageModel;
 import dev.langchain4j.model.moderation.ModerationModel;
 import dev.langchain4j.rag.RetrievalAugmentor;
+import dev.langchain4j.rag.content.retriever.ContentRetriever;
 import dev.langchain4j.retriever.Retriever;
 import dev.langchain4j.service.AiServices;
 import dev.langchain4j.store.memory.chat.ChatMemoryStore;
@@ -98,6 +99,18 @@ public @interface RegisterAiService {
     Class<? extends Retriever<TextSegment>> retriever() default NoRetriever.class;
 
     /**
+     * Configures the way to obtain the {@link ContentRetriever}.
+     * The Supplier may or may not be a CDI bean (but most
+     * typically it will, so consider adding a bean-defining annotation to
+     * it). If it is not a CDI bean, Quarkus will create an instance
+     * by calling its no-arg constructor.
+     *
+     * If unspecified, Quarkus will attempt to locate a CDI bean that
+     * implements {@link ContentRetriever} and use it if one exists.
+     */
+    Class<? extends Supplier<ContentRetriever>> contentRetriever() default BeanIfExistsContentRetrieverSupplier.class;
+
+    /**
      * Configures the way to obtain the {@link RetrievalAugmentor} to use
      * (when using RAG). The Supplier may or may not be a CDI bean (but most
      * typically it will, so consider adding a bean-defining annotation to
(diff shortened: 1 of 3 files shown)


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/AiServicesRecorder.java
```diff
@@ -170,6 +171,29 @@ public T apply(SyntheticCreationalContext<T> creationalContext) {
                                 quarkusAiServices.retrievalAugmentor(supplier.get());
                             }
                         }
+                    } else if (info.getContentRetrieverSupplierClassName() != null) {
```
karesti, line 174: I don't understand what we mean by "automatically wrap into a default" mentioned by @langchain4j

langchain4j: What I meant was: if there is a `ContentRetriever` and no `RetrievalAugmentor`, `DefaultRetrievalAugmentor` should be created wrapping an existing `RetrievalAugmentor`:
```java
return DefaultRetrievalAugmentor.builder()
    .contentRetriever(retriever)
                .build();
```

jmartisk: AFAIU Langchain4j wraps it automatically when you call `aiServices.contentRetriever()`: https://github.com/langchain4j/langchain4j/blob/0.27.1/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java#L333

core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
```diff
@@ -97,6 +98,18 @@
...
     Class<? extends Retriever<TextSegment>> retriever() default NoRetriever.class;
 
+    /**
+     * Configures the way to obtain the {@link ContentRetriever}.
+     * The Supplier may or may not be a CDI bean (but most
+     * typically it will, so consider adding a bean-defining annotation to
+     * it). If it is not a CDI bean, Quarkus will create an instance
+     * by calling its no-arg constructor.
+     *
+     * If unspecified, Quarkus will attempt to locate a CDI bean that
```
jmartisk, line 108: One problem here is that we have exactly the same approach for the `retrievalAugmentor` (that it can be located automatically), but an AI service can only have one of these (not both). And also the same for the classic old `retriever`. An AI service can have at most one out of these **three** components. So we need to decide how to detect and handle potential conflicts. 

The `AiServicesProcessor` already [throws an error](https://github.com/quarkiverse/quarkus-langchain4j/blob/0.9.0/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java#L256)  if both `retriever` and `retrievalAugmentor` are set. 

But the fact that we have auto-detection makes it a bit harder to detect when an error should be thrown.

Maybe the `AiServicesRecorder` should keep track of what `Retriever`/`ContentRetriever`/`RetrievalAugmentor` instances it creates and if more than one of these exists, throw an error? I think that would be best.

core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
```diff
@@ -178,6 +191,18 @@ public RetrievalAugmentor get() {
         }
     }
 
+    /**
+     * Marker that is used to tell Quarkus to use the {@link ContentRetriever} that the user has configured as a CDI bean.
+     * If no such bean exists, then no content retriever will be used.
+     */
+    final class BeanIfExistsContentRetrieverSupplier implements Supplier<ContentRetriever> {
```
jmartisk, line 198: There should probably also be  a `NoContentRetrieverSupplier`  class, just like we have `NoRetrievalAugmentorSupplier`, which will mean "don't attempt to auto-detect a content retriever supplier"

core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
jmartisk: Another possible would be to move the conflict detection to Langchain4j and let the `dev.langchain4j.service.AiServices` builder throw an error any two out of `retriever`/`contentRetriever`/`retrievalAugmentor`  get called (instead of silently overriding the previous call, it would throw an error - @langchain4j WDYT?)

langchain4j: @jmartisk good point, it makes sense to check this in `AiServices`. BTW, old `retriver` is deprecated now and will be removed soon

jmartisk: I've submitted https://github.com/langchain4j/langchain4j/pull/710 so if it's merged, we won't have to be that careful about checking it on the Quarkus side

karesti:
@jmartisk @langchain4j I'm going to close this because I feel I'm not getting all the info, missing stuff and making things slower here

Issue #336, opened 2024-02-28
Integration for scoring models


jmartisk:
LangChain4j supports scoring models, currently with a single implementation, using Cohere: https://github.com/langchain4j/langchain4j/blob/main/langchain4j-cohere/src/main/java/dev/langchain4j/model/cohere/CohereScoringModel.java

This is useful for:
- Further filtering of retrieved content documents, for example when using embeddings - in many cases, it may be necessary to retrieve too many embeddings to get to the relevant data, but we can't slap them all into the query, because providing the LLM too much data makes hallucinate more (and also uses more tokens), so reranking through a scoring model can be beneficial
- When retrieving content by a combination of multiple methods (embeddings, full-text search, etc) that have mutually incomparable scoring, then after fusing the contents into a single list using Reciprocal Rank Fusion, the resulting list may also be sub-optimal and contain a lot of irrelevant content, in which case it is also useful to run it through a rerank then

PR #355, opened 2024-03-07
Support Scoring models with Cohere to allow Reranking

jmartisk:
Fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/336

The test the that actually uses Cohere (`CohereScoringModelTest`): I've added `COHERE_API_KEY` to the repository secrets, so it should, just like the Pinecone tests, run nightly. It's a free key so hopefully we won't get locked out for using it too much.

Not sure if we want to also supply multiple models here? Right now it allows configuring one Cohere scoring model. But I can support multiple.


jmartisk:
Also CC @langchain4j

geoand:
> Not sure if we want to also supply multiple models here? Right now it allows configuring one Cohere scoring model. But I can support multiple.

I think we can do that later

geoand (approved):

langchain4j:
@jmartisk which other rerankers do you have in mind? I plan to at least add integration with jina (also for embeddings), support [cross-encoders in in-process embedding models](https://github.com/langchain4j/langchain4j-embeddings/compare/main...cross_encoders) and potentially add ColBERT, but that will require much more time and is not clear yet whether it will be useful.

PR #372, opened 2024-03-13
Enable ModerationModel for the BAM module

andreadimaio:
The idea of this PR is to enable the `ModerationModel` for the BAM and move the watsonx apis from `v1-beta` to `v1`.
The PR is a draft because I have some considerations about how the moderation part is handled today by the `AiServiceMethodImplementationSupport` and `AiServicesProcessor` classes. In the next few hours I will write another comment to talk about this, but in the meantime this is the code realized so far.


andreadimaio:
I have two points I would like to discuss.

The first point is related to the use of the `ModerationModel` interface.
Today, to work with moderation, you have to annotate a method with the `@Moderate` annotation and it is also mandatory to have the `moderationModelSupplier` property set for the `@RegisterAiService`.
```java
@RegisterAiService(moderationModelSupplier = RegisterAiService.BeanModerationModelSupplier.class)
    interface ChatWithModeration {

        @Moderate
        String chat(String message);
    }
```
It might be more useful to have something "smarter", very similar to what was done for the `StreamingChatModel` with `Multi`, such as: _"If there is a method annotated with `@Moderate`, use the CDI implementation (or the supplier) if present, otherwise throw an exception"._ 

----------------------------------

The second point is about when the call to the `ModerationModel` API is made by the [AiServiceMethodImplementationSupport ](https://github.com/quarkiverse/quarkus-langchain4j/blob/3bd3a16ea45f975af038eb072d8029460e516b08/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java#L157-L169)class. Today, the two API calls are asynchronous (one to the LLM and one to the moderation), which can be an advantage in terms of response time if the Moderation API call detects no hate speech in the input provided, but what about if the API detects hate?

In the best case scenario, we have made a useless call to the LLM API, but what happens when we use the tools and some methods are called? Wouldn't it be easier to make everything synchronous or give to the developer the choice?

geoand:
> It might be more useful to have something "smarter", very similar to what was done for the StreamingChatModel with Multi, such as: "If there is a method annotated with @Moderate, use the CDI implementation (or the supplier) if present, otherwise throw an exception".

The point of why it's done like this (this is how upstream LangChain4j does it) is that your AiService could have multiple methods and perhaps not all methods need it.

geoand:
The PR looks good to me by the way :)

andreadimaio:
> > It might be more useful to have something "smarter", very similar to what was done for the StreamingChatModel with Multi, such as: "If there is a method annotated with @moderate, use the CDI implementation (or the supplier) if present, otherwise throw an exception".
> 
> The point of why it's done like this (this is how upstream LangChain4j does it) is that your AiService could have multiple methods and perhaps not all methods need it.

I don't understand why the fact that an AiService can have multiple methods forces me to write this `@RegisterAiService(moderationModelSupplier = RegisterAiService.BeanModerationModelSupplier.class)`. 

The `@Moderate` annotation is not enough to understand that we need to inject a `ModerationModel` interface?

geoand:
> The @Moderate annotation is not enough to understand that we need to inject a ModerationModel interface?

This is a good point.

geoand:
I'll think about it some more.

geoand:
> The watsonx module supports moderation, but in the same API call.

What do the request and responses look like in this case?

andreadimaio:
> What do the request and responses look like in this case?

If you add the `moderations` field to the request, the text will be checked:
```json
{
    "model_id": "ibm/granite-13b-chat-v2",
    "input": "I want to kill you!",
    "project_id": "xxx",
    "moderations": {
        "hap": {
            "input": {
                "enabled": true,
                "threshold": 0.5,
                "mask": {
                "remove_entity_value": true
                }
            }
        }
    }
}
```
```json
{
    "results": [
        {
            "generated_text": "",
            "generated_token_count": 0,
            "input_token_count": 0,
            "stop_reason": "",
            "moderations": {
                "hap": [
                    {
                        "score": 0.9873412251472473,
                        "input": true,
                        "position": {
                            "start": 0,
                            "end": 19
                        },
                        "entity": "has_HAP",
                        "word": "I want to kill you!"
                    }
                ]
            }
        }
    ]
}
```

andreadimaio:
> I'll think about it some more.

I tried to implement the smart inject of the `ModerationModel`, you can find the details in the commit [Improve ModerateModel injection](https://github.com/quarkiverse/quarkus-langchain4j/pull/372/commits/70db975223a61a3104bfd03154e4ce3575b7f8a7)

geoand:
I'll have a close look at your PR soon!

geoand:
Really nice work (as always)!

I do agree this approach is better than the current state

geoand:
Feel free to make it ready for review when it's good to go

geoand (approved):

andreadimaio:
One note, today the `ModerationModel` doesn't work (is not called) if you use the `StreamingChatModel`.
Maybe we can discuss this in another PR.

andreadimaio:
@geoand, I've been thinking that the changes I made in the **"Improve ModerationModel injection"** commit are not backwards compatible (I didn't use the `@Deprecated`).

geoand:
They are not indeed. 

Which breakage specifically you have in mind as being a problem?

andreadimaio:
When someone uses `@RegisterAiService(moderationModelSupplier = RegisterAiService.BeanModerationModelSupplier.class)`

geoand:
I think we can live with that at this point of the project's life :)

andreadimaio:
You can proceed 👍

geoand (approved):
Thanks again!

PR #386, opened 2024-03-18
RAG in the Dev UI chat page

jmartisk:
- If a retrieval augmentor is available in the CDI container, automatically pick it up by the chat page
- When a user submits a new message, at first the UI shows the original question as submitted, but then, when the response arrives, it gets rewritten to show the full augmented message

- It also contains some adjustments to the page itself, this is mainly for @phillip-kruger to look at:
  - Fixes an issue where each message resets the memory and starts a new conversation
  - Instead of waiting for an actual chat message to happen, make the current conversation disappear as soon as the user enters a new system message to start a new conversation 
  - After a new AI response arrives, we don't just append it to the history, instead we now rewrite the complete history from the history received in the response (from JsonRPC). This is necessary for two reasons - 1) to replace the user message with the augmented one; 2) to avoid potential desynchronization between the history in the UI and the memory on the server, due to potential race conditions and errors.

The next step will be to allow the usage of tools...


jmartisk:
Hint for reviewers: use https://github.com/quarkiverse/quarkus-langchain4j/tree/main/samples/csv-chatbot for reviewing, open the Dev UI and ask stuff like 

> What's the rating of Inception?

geoand:
@jmartisk Can you also upload a screencast of this in action?

Asking because it would be nice for posterity to know what this introduced.
Furthermore I would urge you to reuse the screencast in a social media post 😉

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/ChatJsonRPCService.java
```diff
@@ -27,19 +30,44 @@ public class ChatJsonRPCService {
...
+        this.retrievalAugmentor = null;
+        for (Supplier<RetrievalAugmentor> supplier : retrievalAugmentorSuppliers) {
+            this.retrievalAugmentor = supplier.get();
+            if (this.retrievalAugmentor != null) {
+                break;
+            }
+        }
+        if (this.retrievalAugmentor == null) {
+            for (RetrievalAugmentor retrievalAugmentor1 : retrievalAugmentors) {
+                this.retrievalAugmentor = retrievalAugmentor1;
```
geoand, line 52: Tiny complaint: I can't say I like `retrievalAugmentor1` as a name :)

geoand (approved):

jmartisk:
I've added a GIF demonstrating this :) 
Yeah I plan to make an even better app (maybe it could, for effect, even write to the database) and then use it for promoting on twitter etc

PR #387, opened 2024-03-19
Remove deployment dependency in samples

phillip-kruger:


geoand (commented):
I think they are only needed to ensure that the maven modules are built in the correct order

jmartisk:
Yeah this is most likely necessary to ensure the correct build order, otherwise when you're building the whole repo, the samples might be built before the relevant deployment modules. See https://github.com/quarkiverse/quarkus-langchain4j/pull/267

phillip-kruger:
There is a high probability that users will use these examples as starting points for their own app, coping this into their own poms. Maybe the samples should not be part of the parent and be a separate app all together ? If not, at least add a comment in the pom explaining that this is "normal", and the `-deployment` module should not be included in a user app.

geoand:
> If not, at least add a comment in the pom explaining that this is "normal", and the -deployment module should not be included in a user app

Yeah, we at least need this.

jmartisk:
Ok +1 for a comment

jmartisk:
Cool

PR #369, opened 2024-03-12
Update to LangChain4j 0.28.0

jmartisk:

PR #396, opened 2024-03-22
Add Qdrant vector store

### LINKED ISSUE #392: Qdrant integration
LangChain4j supports Qdrant vector database https://github.com/langchain4j/langchain4j/blob/main/langchain4j-qdrant, it would be nice to have a related extension

lburgazzoli:
Fixes #392


qdrant/deployment/src/test/java/io/quarkiverse/langchain4j/qdrant/QdrantEmbeddingStoreTest.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    protected EmbeddingModel embeddingModel() {
+        return embeddingModel;
+    }
+
+    private static JavaArchive archive() {
+        Asset properties = new StringAsset(String.join("\n",
+                "quarkus.langchain4j.qdrant.devservices.collection.name=" + COLLECTION_NAME,
+                // TODO: as today the dev service leverages the service name as a name of the collection, find a better way
+                "quarkus.langchain4j.qdrant.devservices.service-name=" + COLLECTION_NAME,
```
lburgazzoli, line 51: The `QdrantEmbeddingStore` does not auto create the collection hence at this stage I set it to be the same as the service name. Ideally it should use the same value as `quarkus.langchain4j.qdrant.devservices.collection.name` but I haven't found a way to get such value at build time (long time since I did a quarkus extension)

qdrant/runtime/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+            <artifactId>quarkus-arc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.langchain4j</groupId>
+            <artifactId>quarkus-langchain4j-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-qdrant</artifactId>
```
geoand, line 23: Ideally we would want to avoid this altogether as it forces us use a different network stack altogether. 

I suppose we could use it for now as a first version, but I would really like us to move to our own client (which has a bunch of benefits)

qdrant/deployment/src/test/java/io/quarkiverse/langchain4j/qdrant/QdrantEmbeddingStoreTest.java
geoand: You can't :)

lburgazzoli: ok :) so you are ok to use the service name ? or the creation of the collection should be moved to runtime ?

geoand: > ok :) so you are ok to use the service name

Sure, we do can do that.

lburgazzoli: ok, so apart from the client discussion, this is likely ready for review

qdrant/runtime/pom.xml
jmartisk: Yeah... I do agree, but...
But the thing is that if we implement our own client, we bring upon ourselves a huge maintenance burden. :/
With the upstream impl, we simply get future fixes "for free". I can't imagine us maintaining separate clients for everything that langchain4j supports.

For Milvus, we currently also use the upstream (gRPC-based) client. I tried creating our own, but it turned out to be quite a nightmare.

geoand: In this case it should be much easier as there is also a REST API.

But I guess we can leave it for another time...

cescoffier: If we use the upstream client we would not have metrics, tracing, security, context propagation and so on. Basically, we would bring nothing.

cescoffier: You may even create too many event loops.

lburgazzoli: I don't have any objection about using our own client, but I don't have enough knowledge of the underlying langchain4j implementation to asses how complex and maintainable it would be.

qdrant/deployment/src/main/java/io/quarkiverse/langchain4j/qdrant/QdrantBuildConfig.java
```diff
@@ -0,0 +1,78 @@
...
+             * The vector parameters.
+             */
+            VectorParamsConfig vectorParams();
+        }
+
+        interface VectorParamsConfig {
+            /**
+             * Distance function used for comparing vectors
+             */
+            io.qdrant.client.grpc.Collections.Distance distance();
```
jmartisk, line 70: I'm not completely sure it's appropriate for us to use an enum from upstream code in our config
- If we migrate away towards our own client (as we discussed), this might need a breaking change
- If the upstream code introduces changes, this might mean a breaking change on our side too
- If they add any funky stuff in the enum, it might break bytecode recording

I'd say we could rather create a copy of this. @geoand WDYT?

qdrant/deployment/src/main/java/io/quarkiverse/langchain4j/qdrant/QdrantDevServicesProcessor.java
```diff
@@ -0,0 +1,231 @@
...
+        final Supplier<DevServicesResultBuildItem.RunningDevService> defaultQdrantSupplier = () -> {
+
+            // Starting Qdrant
+            timeout.ifPresent(container::withStartupTimeout);
+            container.start();
+
+            if (config.vectorCfg() != null) {
+                try (QdrantClient client = client(container)) {
+                    client
+                            .createCollectionAsync(
```
jmartisk, line 161: I can only see the creation of collections here in the Dev Services code. Do we support automatically creating the collection when connecting to a non-devservices instance? (Most other embedding stores can do that) 
From looking at the code, the upstream implementation doesn't have that, so with this we would only get it for dev services, am I right?

jmartisk:
I see `java.lang.ExceptionInInitializerError: Exception java.lang.UnsatisfiedLinkError: Native Library /home/runner/.djl.ai/tokenizers/0.15.0-0.26.0-linux-x86_64/libtokenizers.so already loaded in another classloader [in thread "main"]` 

That is caused by https://github.com/langchain4j/langchain4j/issues/776. For now you'll need to work around it in the `QdrantEmbeddingStoreTest` like I did in other tests, see  https://github.com/quarkiverse/quarkus-langchain4j/pull/369/files#diff-7f790b81296e9b9dc4ad5e99b7e3c1a79ed9337fea3d41b6a3f3861d329b30e3R29-R40 - we need to make sure that the embedding model constructor is not called twice under different class loaders

qdrant/deployment/src/main/java/io/quarkiverse/langchain4j/qdrant/QdrantBuildConfig.java
geoand: I'm off today, so I'll have to think about this tomorrow

lburgazzoli:
> I see `java.lang.ExceptionInInitializerError: Exception java.lang.UnsatisfiedLinkError: Native Library /home/runner/.djl.ai/tokenizers/0.15.0-0.26.0-linux-x86_64/libtokenizers.so already loaded in another classloader [in thread "main"]`
> 
> That is caused by [langchain4j/langchain4j#776](https://github.com/langchain4j/langchain4j/issues/776). For now you'll need to work around it in the `QdrantEmbeddingStoreTest` like I did in other tests, see https://github.com/quarkiverse/quarkus-langchain4j/pull/369/files#diff-7f790b81296e9b9dc4ad5e99b7e3c1a79ed9337fea3d41b6a3f3861d329b30e3R29-R40 - we need to make sure that the embedding model constructor is not called twice under different class loaders

done

qdrant/deployment/src/main/java/io/quarkiverse/langchain4j/qdrant/QdrantDevServicesProcessor.java
jmartisk: Ok, since we're also using the upstream client for now, I think this is something that should rather be implemented upstream, if at all. I agree it is not trivial to do this in a robust way.

qdrant/deployment/src/main/java/io/quarkiverse/langchain4j/qdrant/QdrantBuildConfig.java
lburgazzoli: would be ok to define a wrapper enum ? like 

```java
enum Distance {
    Cosine(io.qdrant.client.grpc.Collections.Distance.Cosine);
    
    Distance(io.qdrant.client.grpc.Collections.Distance) {
     ...
    }
}
```

jmartisk: That would still depend on the upstream class :( 
I'd say we should have a copy with all possible (string-based) values and some kind of adapter that maps from our enum  to the upstream enum

lburgazzoli: done

jmartisk (approved):

PR #402, opened 2024-03-25
Use correct memory setting and rename java running script for clarity

dandreadis:
Use a memory setting that avoids sending past translated paragraphs.


samples/cli-translator/src/main/resources/application.properties
```diff
@@ -1,5 +1,9 @@
+quarkus.langchain4j.chat-memory.memory-window.max-messages=2
```
jmartisk, line 1: This is very awkward, isn't there a configuration to simply turn off memory altogether? I don't think there is. Perhaps this is something we should implement, @geoand?

dandreadis: I didn't find one to turn it off, so I've restricted it to 2, to let system prompt and user prompt go through.

geoand: There is: you can do it on `@RegisterAiService` with the `NoMemory...`  in `chatMemory...`

jmartisk: Ah yeah. I mean this should be achievable by configuration properties.

geoand: Yeah, I guess, but I'm generally in favor of having one way of doing things if we can

jmartisk: Ok, so for now I'd suggest changing the AI service annotation to `@RegisterAiService(chatMemoryProviderSupplier = RegisterAiService.NoChatMemoryProviderSupplier.class)`, is that ok @dandreadis ? Feels cleaner than the config property.

samples/cli-translator/src/main/resources/application.properties
```diff
@@ -1,5 +1,9 @@
...
 quarkus.langchain4j.openai.api-key=${OPENAI_API_KEY}
 quarkus.langchain4j.openai.chat-model.temperature=0
 quarkus.langchain4j.openai.timeout=60s
 quarkus.banner.enabled=false
-quarkus.log.level=WARN
\ No newline at end of file
+quarkus.log.level=WARN
+# if you want to log the requests and responses that go to OpenAI:
```
jmartisk, line 7: Maybe the comment should say that in that case you also need to set `quarkus.log.level` to `INFO`, otherwise they won't really get logged

PR #416, opened 2024-03-27
Fix Mistral build cycle issue

geoand:


jmartisk:
Yeah this EmbeddingModelBuildItem is clumsy, I need to find a better general solution. The reason why the Dev UI processor isn't simply looking into the list of available beans is that it also needs to register JsonRPC providers, which are also beans, so it can't depend on `SynthesisFinishedBuildItem` or so.

jmartisk (approved):

Issue #418, opened 2024-03-27
EasyRAG sample doesn't work with tesseract


geoand:
It seems like when Tesseract is installed, Tika seems to lose some context or something, as we get:

```
Caused by: org.apache.tika.exception.TikaException: TesseractOCRParser bad exit value 1 err msg: Tesseract Open Source OCR Engine v4.1.1 with Leptonica
Error in fopenReadStream: file not found
Error in pixRead: image file not found: The "Standard savings account" offers a first-time savers and those with low balances a hassle-free way to save.
Image file The "Standard savings account" offers a first-time savers and those with low balances a hassle-free way to save. cannot be read!
Error during processing.
```

This indicates that the text file (named `standard-saving-account.txt`) is not being parsed correctly

geoand:
@jmartisk @sberyozkin any idea what's going on?

cc @dandreadis who first found this issue

jmartisk:
I was just about to report the same.
This is super mega weird, I could swear that it worked for me yesterday, and then suddenly it stopped working.

dandreadis:
I haven't quite understood how is this supposed to work: when installed, does it take over for all types of document parsing, or just specific ones? In my system, with that in place it parsed 2 out of the 4 files. (also produced multi line warnings).

jmartisk:
Yeah I now also get just 2 files parsed successfully. I absolutely don't understand why it worked for me yesterday and what has changed since then. 
Even weirder, we have `EasyRagTest` in place that depends on being able to parse a PDF file, and the test passes, but with the `chatbot-easy-rag` application, PDF fails.

geoand:
My guess is that it works properly only if the image is parsed last

jmartisk:
Seems like it has problems with png images, when I convert it to jpg, it seems to work.

jmartisk:
Let's just convert the `retirement-money-market.png` file to jpg for now, hm?

geoand:
Sure yeah

dandreadis:
When I tried it manually (using the command line) it correctly extracted the text from the .png.

PR #420, opened 2024-03-27
Convert retirement-money-market.png to jpg

jmartisk:
To be sure, @geoand could you try and check that it fixes it for you?

Fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/418


geoand:
Same problem for me unfortunately

jmartisk:
Uhm, great, and it started working with png for me again.

jmartisk:
Then let's try and just get rid of pictures, if that helps, PDF and ODT should be enough to demonstrate the capabilities...

jmartisk:
But generally I think langchain4j (and by extension, Quarkus), should offer some options to configure Tika (disable/enable OCR), which I don't think is possible right now.

geoand:
> But generally I think langchain4j (and by extension, Quarkus), should offer some options to configure Tika (disable/enable OCR), which I don't think is possible right now.

cc @langchain4j

geoand (approved):

langchain4j:
Tika loader can be configured: https://github.com/langchain4j/langchain4j/blob/6fd8fbcade91f9726489a6e40328cecc3b9ef660/document-parsers/langchain4j-document-parser-apache-tika/src/main/java/dev/langchain4j/data/document/parser/apache/tika/ApacheTikaDocumentParser.java#L49

jmartisk:
Aha ok, that will need some thinking on our side, whether we want to generally add configuration properties for different types of document parser (and manage the parsers in CDI), or just something simple in the easy rag extension specifically for Tika (and assume that in other cases, it will be up to the user to build an appropriate document parser)

PR #422, opened 2024-03-27
Updates to avoid code duplication

karesti:
This Pr is related to this one 

https://github.com/langchain4j/langchain4j/pull/845


karesti:
I created a PR in langchain4j to avoid duplication of code

What do you think about this ?

geoand:
You're in charge of the Infinispan stuff, so whatever you think is best :)

karesti:
@geoand I was also thinking overall strategy concerning other integrations, since the code is duplicated and the effort seems high

geoand:
Sure, but we'll see tackle it on a per case badis

Issue #434, opened 2024-04-02
Add support for `fromResource` in `@SystemMessage` and `@UserMessage`


geoand:
@langchain4j I assume those attributes are stable and won't change, correct?

If so, this is a good time to add support for them :)

langchain4j:
@geoand seems like it is "stable", but I am not sure this brings much value now, as we are thinking about adding a way to dynamically supply system/user message templates as lambdas/suppliers.

langchain4j:
Can't find the original discussion, but here are some relevant ones:
https://github.com/langchain4j/langchain4j/pull/862
https://github.com/langchain4j/langchain4j/issues/785

geoand:
🙏🏼

PR #435, opened 2024-04-02
Support fromResource for user and system messages

geoand:
- Closes: #434


edeandrea:
Yes! This is awesome!

Issue #446, opened 2024-04-05
Investigate use of quarkus-wiremock


geoand:
It would be nice to be able to cut down on a lot of the code duplication we have in tests

geoand:
https://github.com/quarkiverse/quarkus-langchain4j/pull/448 laid the groundwork but the real challenge will be updating the OpenAI tests to leverage this

geoand:
https://github.com/quarkiverse/quarkus-langchain4j/pull/455 will close this

PR #452, opened 2024-04-08
Require an key for OpenAI only when default URL is used

geoand:


jmartisk:
Does this mean we can now remove the dummy keys from various places in tests and docs samples?

jmartisk:
Probably just tests that use a mock. We don't override the URL in docs samples.

geoand:
Actually I'll leave out fixing the tests for when I get around to https://github.com/quarkiverse/quarkus-langchain4j/issues/446

maxandersen:
Remind me why this is needed ?

Shouldn't we just not do anything for dummy key and have users set that value explicitly ?

geoand:
The point is that the OpenAI extension can be used with other model serving services besided OpenAI and these services may not require an API key

PR #467, opened 2024-04-17
Add support for Azure AD Tokens

csotiriou:
Adds support for Azure AD tokens. 

When using the Azure OpenAI models, there is also the case where a user needs to authenticate using an Azure AD token, obtained elsewhere. The Python version already supports this like this:

```
client = AzureOpenAI(
    azure_ad_token=token_to_use,
    azure_deployment="...",
    api_version="...",
    azure_endpoint="https://..."
)
```

by allowing to provide the Azure AD token directly. 

This PR adds a configuration `quarkus.langchain4j.azure-openai.azure-ad-token` to allow. the token to be passed.


geoand:
Thanks for this!

@edeandrea as you've used Azure, mind checking this?

edeandrea:
I've never used the Azure AD token before. So long as its passed in the `Authorization` header as a `Bearer` token then this looks fine.

I think making `quarkus.langchain4j.azure-openai.api-key` optional is fine, but there needs to be something in the recorder that enforces that either the api token or the Azure AD token needs to be present.

csotiriou:
@edeandrea I made some changes, and I also changed the error message that is being thrown. I have a remark, I saw a comment ( I believe it was from @geoand ) about a Dummy default value in the api key with a note that the Smallrye config couldn't accept the `Optional` value there. 

I tested it with Quarkus 3.9.1 and it seems to work with the `Optional`, just trying to make sure...

geoand:
> I tested it with Quarkus 3.9.1 and it seems to work with the Optional, just trying to make sure...

I had this problem with the API key, but that might not be a problem anymore (I have not checked) and it also might not be a problem for other fields.

edeandrea:
I also had the problem recently when I added the Antropic module. Maybe it's been fixed?

geoand:
Could be, I'll have to check next week

csotiriou:
I have just seen that the default implementation (the default "dummy" text being passed as an argument) was also there for initializing the OpenAI rest client (since it can't be instantiated without passing the API key). Therefore, I had to override the instantiation behaviour a bit. I hope i didn't break anything by doing so.

geoand:
Thanks, I'll have a close look on Monday!

openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/config/LangChain4jAzureOpenAiConfig.java
```diff
@@ -11,10 +11,7 @@
 import io.quarkus.runtime.annotations.ConfigDocSection;
 import io.quarkus.runtime.annotations.ConfigGroup;
 import io.quarkus.runtime.annotations.ConfigRoot;
-import io.smallrye.config.ConfigMapping;
-import io.smallrye.config.WithDefault;
-import io.smallrye.config.WithDefaults;
-import io.smallrye.config.WithParentName;
+import io.smallrye.config.*;
```
geoand, line 14: Please revert this change - I know you did not do this yourself but it was done by the IDE, but we generally try to stay away from star imports

openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/config/LangChain4jAzureOpenAiConfig.java
```diff
@@ -77,8 +81,7 @@ interface AzureAiConfig {
         /**
          * Azure OpenAI API key
          */
-        @WithDefault("dummy") // TODO: this should be optional but Smallrye Config doesn't like it..
-        String apiKey();
+        Optional<String> apiKey();
```
geoand, line 84: Have you tested this with multiple (named) models? That is where it stared failing for me a while back.

geoand (commented):
Very nice!

I just added a small comment and one more important question

openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/config/LangChain4jAzureOpenAiConfig.java
csotiriou: > Have you tested this with multiple (named) models? That is where it stared failing for me a while back.

I have just tested it with the following configuration.

```
#quarkus.langchain4j.azure-openai.api-key=token1
quarkus.langchain4j.azure-openai.openai1.resource-name=<<myname>>
quarkus.langchain4j.azure-openai.openai1.deployment-name=gpt-35-turbo
quarkus.langchain4j.azure-openai.openai1.api-version=2023-07-01-preview

quarkus.langchain4j.azure-openai.openai2.azure-ad-token=token2
quarkus.langchain4j.azure-openai.openai2.resource-name=<<myname>>
quarkus.langchain4j.azure-openai.openai2.deployment-name=gpt-35-turbo
quarkus.langchain4j.azure-openai.openai2.api-version=2023-07-01-preview


quarkus.langchain4j.embedding-model.provider=dev.langchain4j.model.embedding.AllMiniLmL6V2QuantizedEmbeddingModel
quarkus.langchain4j.redis.dimension=384
rag.path=src/main/resources/catalog

quarkus.log.level=INFO
```

Here are my observations.

In the `AzureOpenAiRecorder` there are 3 configurations that are created. 
- `<default>` --no matter what you do, this is always created
- `openai1`
- `openai2`

The problem lies with the Default configuration. In the above configuraiton, since it's always created, the default configuration will result in an error inside the code when checking the validity of the configuration, thus crashing the rest of the application. This will be hidden from the user. It seems that smallrye works somewhat fine in this scenario, it's the assumptions in the code that make it unusable.

Temporary resolution: 
Put `quarkus.langchain4j.azure-openai.enable-integration=false` in application.properties of the main application, thus making the recorder not check the default configuration at all.

Proposal for permanent solution: 
Make assumptions as to how a configuration is valid. A configuration should be valid if at least one of the tokens is present, otherwise it should not be checked.

I can take a look if you want, I will just need some advice on the intended usage...

geoand: It turns out that `<default>` was being requested because of an AiService being used

geoand: I have also opened #477  to improve the test coverage because this case did get me wondering whether we were missing something.

csotiriou: Ah, yes, it was entirely my fault, sorry about that.

geoand: No problem :)

csotiriou:
I have ran the unit tests, I have tested with a configuration with 2 named configs and verified that the Recorder is called for both of them, and it seems to be working correctly.

openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/config/LangChain4jAzureOpenAiConfig.java
csotiriou: Done

geoand:
> @geoand do you believe that there should be a documentation update in this PR? If yes, I can do it.

If you want to do that now, then by all means :).
If you want to do it in a follow up PR, that's fine too

geoand:
Mind squashing the commits please?

Thanks

geoand (approved):
Thanks!

PR #480, opened 2024-04-22
Remove embedding model build item

jmartisk:
This was kinda confusing, I've reworked the way the `LangChain4jDevUIProcessor` detects whether the embedding store page should be present


geoand (approved):
Thanks!

PR #483, opened 2024-04-22
Added responseFormat to Azure OpenAi Chat Model

UnvirtualHH:


openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/AzureOpenAiChatModel.java
```diff
@@ -116,7 +120,8 @@ private Response<AiMessage> generate(List<ChatMessage> messages,
                 .topP(topP)
                 .maxTokens(maxTokens)
                 .presencePenalty(presencePenalty)
-                .frequencyPenalty(frequencyPenalty);
+                .frequencyPenalty(frequencyPenalty)
+                .responseFormat(responseFormat);
```
Plawn, line 124: Looking at the Lang4Chain documentation, I see

```java
AzureOpenAiChatModel.builder()
        ...
        .responseFormat(new ChatCompletionsJsonResponseFormat())
        .build();
```

I don't see, how the correct item is inject in the builder here, is it actually working as expected ?

openai/azure-openai/runtime/src/test/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorderEndpointTests.java
```diff
@@ -205,6 +205,11 @@ public Double frequencyPenalty() {
                     return null;
                 }
 
+                @Override
```
Plawn, line 208: Shouldn't a test case for the JSON use be added as well ?

openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/AzureOpenAiChatModel.java
UnvirtualHH: Well, yes, it works as Expected. ChatCompletionsJsonResponseFormat is also possible, but you lose the flexibility of changing it over the application properties. It is now implemented the same way as for the open ai part directly.

Plawn: Thanks !

openai/azure-openai/runtime/src/test/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorderEndpointTests.java
UnvirtualHH: Testing is for beginners, expert just do it ;)  Well, without joking, for sure a test could be added, but I needed a quick working version and I'm pretty sure that the impact from this change is minimal to the rest of the code.

geoand (approved):
Thanks!

openai/azure-openai/runtime/src/test/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorderEndpointTests.java
Plawn: Yes, I'm sure it won't break the rest of the code :)
Just in case, this changes, the code won't be protected :)
If I have the time, I'll try to add a test

PR #487, opened 2024-04-23
Add integration module with quarkus-websockets-next

geoand:
This module allows for automatic use of memoryId when no object is specified

TODO: add tests and fix for native image


jmartisk:
Are you also going to update the samples that use websockets?

docs/modules/ROOT/pages/websockets.adoc
```diff
@@ -0,0 +1,70 @@
...
+    }
+}
+----
+
+Two things are important to note in the snippets above:
+
+* There is no `@MemoryId` field being used in the AI service. Quarkus will automatically all the WebSocket connection ID as the memory ID.
+This ensures that each WebSocket session has its own chat memory.
+* The use of `@SessionScoped` is important as the scope of the AI service is tied to the scope of the WebSocket endpoint.
+This allows Quarkus to automatically clear chat memory when the WebSocket connection is closed for whatever reason.
```
jmartisk, line 70: So if one needs to be able to "resume" an existing conversation after the websocket was closed (crashed, or the conversation was inactive for a longer time) they have to implement a custom mechanism? I'm just thinking how important that may be. It will probably be a little tricky to implement.

geoand: Yes, they would have to do something custom. We can certainly think of doing something more advanced in the future if the needs arise.

jmartisk:
Hm, one inconvenience that I now noticed is that by making the AI service session-scoped, I can no longer call it from a REST endpoint. I don't quite understand why we need to make it session-scoped? Isn't the service in itself stateless and only controlled by what the memory contains?

geoand:
> I don't quite understand why we need to make it session-scoped? 

By making it `@SessionsScoped` we are able to automatically remove all memory when the websocket session closes.

> Isn't the service in itself stateless and only controlled by what the memory contains?

It is, but it integrates with memory and can thus be used for automatic memory management.

> I can no longer call it from a REST endpoint

I suggest using different AI services.

Issue #491, opened 2024-04-24
ClassFormatError: Duplicate method name when registering AI Service


iocanel:
This is an edge case, but I still manage to hit that.

`AiServiceProcessor`  tries to implement the methods of the interface that `@RegisterAiService` is added two.
The list of interfaces however, may contain methods with the same signature.
An example case is when having interfaces redefine parent methods.

Either-way, we should guard against this case, as it leads to `ClassForatError` in the generated class.

PR #492, opened 2024-04-24
Fix duplicate method exception when registering ai services

iocanel:
Resolves: #491

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/MethodUtil.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/MethodUtil.java
@@ -1,0 +1,53 @@
+package io.quarkiverse.langchain4j.deployment;
+
+import org.jboss.jandex.MethodInfo;
+
+public final class MethodUtil {
+
+    private MethodUtil() {
+        //Utility class
+    }
+
+    /**
+     * Returns the signature of a method
+     *
+     * @param method the method to get the signature from
+     * @return the signature of the method
+     */
+    public static String signature(MethodInfo method) {
+        StringBuilder signature = new StringBuilder();
+        signature.append(method.name());
+        signature.append("(");
+        for (int i = 0; i < method.parameters().size(); i++) {
+            signature.append(method.parameters().get(i).type());
+            if (i < method.parameters().size() - 1) {
+                signature.append(", ");
+            }
+        }
+        signature.append(")");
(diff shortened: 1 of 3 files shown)


core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/MethodUtil.java
```diff
@@ -0,0 +1,51 @@
+package io.quarkiverse.langchain4j.deployment;
+
+import org.jboss.jandex.MethodInfo;
+
+public final class MethodUtil {
```
geoand, line 5: I am wondering if there is already a util like this in Quarkus

geoand (commented):
Thanks for this!

I added a comment

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/MethodUtil.java
iocanel: Maybe `Methods` has something like it. Looking.

iocanel: Didn't find any.

geoand: I'll get back to you in a few minutes

iocanel: Ok, meanwhile I added a couple of tests:
- one the verifies the problem
- one the verifies the correctness of the util

geoand: > I'll get back to you in a few minutes

Does `io.quarkus.gizmo.SignatureBuilder` do the trick?

iocanel: Let me check

iocanel: No, I don't think we can use that.

They don't seem to provide the name and that makes it of limited use to our use case: https://github.com/quarkusio/gizmo/blob/main/src/main/java/io/quarkus/gizmo/MethodSignatureBuilderImpl.java

geoand (approved):
Thanks!

Issue #493, opened 2024-04-24
Building quarkus-langchain4j on MacOS with Podman fails at Antora Container Start InternalServerErrorException: Status 500: {"cause":"permission denied",


myfear:
Trying to build quarkus-langchain4j on my M2 Pro with. Opening this after briefly talking with @maxandersen . 

podman version
```
Client:       Podman Engine
Version:      5.0.1
API Version:  5.0.1
Go Version:   go1.22.1
Git Commit:   946d055df324e4ed6c1e806b561af4740db4fea9
Built:        Mon Apr  1 14:28:31 2024
OS/Arch:      darwin/arm64

Server:       Podman Engine
Version:      5.0.2
API Version:  5.0.2
Go Version:   go1.21.9
Built:        Wed Apr 17 02:00:00 2024
OS/Arch:      linux/arm64
```

- Docker compatibility mode is enabled.
-  vmtype: applehv
-  "Rootful": true

 mvn --version
```
Apache Maven 3.9.6 (bc0240f3c744dd6b6ec2920b3cd08dcc295161ae)
Maven home: /opt/homebrew/Cellar/maven/3.9.6/libexec
Java version: 21.0.2, vendor: Homebrew, runtime: /opt/homebrew/Cellar/openjdk/21.0.2/libexec/openjdk.jdk/Contents/Home
Default locale: en_DE, platform encoding: UTF-8
OS name: "mac os x", version: "14.4.1", arch: "aarch64", family: "mac"
```

Build Error:

```posh
[INFO] Creating container for image: antora/antora:3.0.1
[WARNING] Image name alpine:3.16 was substituted by configuration to registry.access.redhat.com/ubi8/ubi-minimal:8.4. This approach is deprecated and will be removed in the future
[INFO] Using registry.access.redhat.com/ubi8/ubi-minimal:8.4 as a substitute image for alpine:3.16 (using image substitutor: DefaultImageNameSubstitutor (composite of 'ConfigurationFileImageNameSubstitutor' and 'PrefixingImageNameSubstitutor'))
[WARNING] Unable to mount a file from test host into a running container. This may be a misconfiguration or limitation of your Docker environment. Some features might not work.
[INFO] Container antora/antora:3.0.1 is starting: 302c8f5f77645f5e0a30a32651f980efa7f0a706d946288ae9b5f1e76b08bafc
[ERROR] Could not start container
[...]

PR #496, opened 2024-04-24
disable antora until having proper working solution

maxandersen:
fixing https://github.com/quarkiverse/quarkus-langchain4j/issues/493 and what is [related failures](https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/langchain4j.20build.20troubles) like missing classes in test and conflicting 8081 ports.

by removing antora all builds works again on OSX/podman.

My theory is that the antora devservice is assuming volume mounts/permissions can be adjusted from inside container in a portable way - it can't (unfortunately).

Thus it fails somewhat silently - causing the AntoraTest to not shutdown, causing 8081 failures ...why it then result in devui not being able to find classes is beyond me - but that error goes away with this code commented out.

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -76,11 +76,13 @@
             <artifactId>quarkus-langchain4j-watsonx</artifactId>
             <version>${project.version}</version>
         </dependency>
-        <dependency>
+        <!-- remove as does not work on podman machine 
+            <dependency>
             <groupId>io.quarkiverse.antora</groupId>
             <artifactId>quarkus-antora</artifactId>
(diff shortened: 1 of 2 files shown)


jmartisk:
Could we, instead of removing, move it into a profile that isn't enabled by default?

cescoffier (approved):

maxandersen:
We probably can make it a profile. I just did simplest thing that works on my machine. 

I suggest we merge this so build/samples works. 

Then someone who are able to run it can make pr with profile

PR #500, opened 2024-04-24
Improve user experience for samples

geoand:
* Convert samples to standalone applications
* Vastly improve the chatbot samples


samples/chatbot-easy-rag/pom.xml
```diff
@@ -2,15 +2,36 @@
...
     <name>Quarkus LangChain4j - Sample - Chatbot with Easy RAG</name>
+    <version>1.0-SNAPSHOT</version>
+
+    <properties>
+        <compiler-plugin.version>3.12.1</compiler-plugin.version>
+        <maven.compiler.parameters>true</maven.compiler.parameters>
+        <maven.compiler.release>17</maven.compiler.release>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
```
maxandersen, line 16: Any reason to not have it use the platform as is recommended for user apps?

geoand: The only reason is that I can easily switch to snapshot Quarkus when needed 😎

jmartisk: +1, I tackle this problem often, having to also build a snapshot of the platform would be painful

jmartisk (approved):
This is great!

samples/chatbot-easy-rag/pom.xml
maxandersen: In Quarkus quick start it's a property so easy to override. Snapshot/Quarkus can be in a profile. 

Just that it would be nice the samples works as defacto standard Quarkus app and allow quarkus ext add work like every other default app.

PR #503, opened 2024-04-25
Override the maven-compiler-plugin version in the samples

jmartisk:
This makes the samples work OOTB with Maven 3.8


geoand (approved):
Thanks!

jmartisk:
@geoand the failure in `io.quarkiverse.langchain4j.websockets.next.tests.WebSocketsNextTest.webSocketRequest`, have you seen that before? Looks like it might be a flaky test?

geoand:
I was afraid that the test would be flaky as it has a bunch of async intersections...

PR #507, opened 2024-04-25
Turn off built-in fault tolerance by default

jmartisk:
I've also added some basic validation checks where possible (models that are implemented on our side), but I'm not completely sure about this. We could potentially do that in the recorders, making it more like validating the configuration?!


geoand (approved):
Thanks

PR #511, opened 2024-04-25
Fix wording problem

holly-cummins:
I made a guess at the intended meaning, feel free to adjust if I got it wrong!


geoand (approved):
Thanks a lot!

geoand:
Oops, I just saw that @gastaldi had fixed the same issue!

gastaldi:
`assign` sounds nicer than `use`, so it's up to you to decide which term to use 😉

holly-cummins:
I had to think about it, but I *think* I would say either 
- Quarkus will automatically use the WebSocket connection ID as the memory ID
- Quarkus will automatically assign the WebSocket connection ID as the memory ID
- Quarkus will automatically assign the memory ID to the WebSocket connection ID

... but I agree with @gastaldi that the middle one is a bit awkward and maybe not proper English. And I think I actually like 'use' over 'assign', because it captures a nice little internal thought process for the extension code: 'oh man, I need a web socket id, what can I use, what can I use, umm ... ooh! I found a memory id! cool, I'll use that!'

gastaldi:
@holly-cummins Right, I felt `use` natural while reading, probably because the previous sentence also uses the same verb, so I went with that ;)

cescoffier:
Can you fix the conflict?

holly-cummins:
@cescoffier I've done that so that this is mergeable if we need it ... but I actually think we should close this and go with @gastaldi's version. So I've done that, too.

PR #520, opened 2024-04-26
Move BAM APIs version to 2024-04-15

andreadimaio:
Changes:

-  The moderation API now takes the **hap** and **social_bias** values as parameters.
-  The embedding API can now work with multiple **inputs**.


geoand (approved):
Thanks!

PR #539, opened 2024-05-03
Add secure-fraud-detection demo

sberyozkin:
Draft PR for a secure-fraud-detection demo.

The startup issue described here earlier was resolved as advised by Georgios


geoand:
Nice!

I am guessing that the error you are seeing is due to maven not being configured with to make `javac` retain method parameter names.

sberyozkin:
Thanks @geoand, yes, I was wondering how exactly does the translate demo works which has 2 parameters, and now I've found https://github.com/quarkiverse/quarkus-langchain4j/blob/main/samples/cli-translator/pom.xml#L12 :-)

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/TransactionRepository.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.List;
+
+import dev.langchain4j.agent.tool.Tool;
+import io.quarkus.hibernate.orm.panache.PanacheRepository;
+import jakarta.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+public class TransactionRepository implements PanacheRepository<Transaction> {
+
+    @Tool("Get the transaction for a given customer for the last 15 minutes")
```
jmartisk, line 13:
```suggestion
    @Tool("Get the transactions for a given customer for the last 15 minutes")
```

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/ChatMemoryBean.java
```diff
@@ -0,0 +1,30 @@
...
+
+import jakarta.annotation.PreDestroy;
+import jakarta.enterprise.context.RequestScoped;
+
+import dev.langchain4j.memory.ChatMemory;
+import dev.langchain4j.memory.chat.ChatMemoryProvider;
+import dev.langchain4j.memory.chat.MessageWindowChatMemory;
+
+@RequestScoped
+public class ChatMemoryBean implements ChatMemoryProvider {
```
jmartisk, line 14: Isn't this implemented exactly the same as the default memory provider? Why is this needed then?

sberyozkin: @jmartisk It was only a copy and paste, I think it was added in the original fraud-detection demo just to should how it can be customized. I'm not sure yet it will be needed for the secure variation of the demo

sberyozkin: @jmartisk @geoand I've dropped this custom chat memory provider from the demo as a new memory id is provided on every request, so indeed it is not clear to me how it can be used in this specific demo, but a custom memoryId provider will definitely be of interest here as I'm having an authenticated session, so I can use exactly the same memory id to correlate several fraud amount and distance requests.
So I'll try to add it here with another commit and ping you

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/ProvidersConfigSource.java
```diff
@@ -0,0 +1,29 @@
+package io.quarkiverse.langchain4j.sample;
+
+import java.util.Map;
+import java.util.Set;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+public class ProvidersConfigSource implements ConfigSource {
```
geoand, line 8: Can you explain what this does and why it's needed?

samples/secure-fraud-detection/README.md
```diff
@@ -0,0 +1,127 @@
...
+- [TransactionRepository.java](./src/main/java/io/quarkiverse/langchain4j/samples/TransactionRepository.java)
+
+### AI Service
+
+This demo leverages the AI service abstraction, with the interaction between the LLM and the application handled through the AIService interface.
+
+The interface uses specific annotations to define the LLM and the tools to be used:
+
+```java
+@RegisterAiService(chatMemorySupplier = AiConfig.MemoryProvider.class,
```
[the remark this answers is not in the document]
sberyozkin: @geoand Sure, I'll need to work on README, I'm still thinking how to present to the user without WS

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/ProvidersConfigSource.java
sberyozkin: Sure, this is more to support the message the whole setup is about security (google client id, secret and open ai key are not in the open application.properties), I can drop it if you'd like

geoand: We don't need to drop it necessarily, but if we do keep it I'd like to have some Javadoc on what this does and why it's necessary

samples/secure-fraud-detection/README.md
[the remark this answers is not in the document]
geoand: Thanks

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/ProvidersConfigSource.java
sberyozkin: Sure

sberyozkin: Hi @geoand I'll drop it in favor of proposing to use env properties as shown in the cli translate demo, the fewer resources the better :-), and it will be the same recommendation across at least 2 demos

sberyozkin:
@geoand @jmartisk 
I'm getting closer to the finalizing the demo flow, a uses accesses the login page where a Google authentication option is offered, user logs in and is redirected to the fraud detection entry page where the user can choose to check it by amount or distance. 
This in itself is useful enough to show how to use the authentication token, but I think the most interesting part is actually how to utilize the fact the OIDC session is available for correlating multiple fraud queries to get more interesting responses each time the same authenticated user tries to check the fraud status, so I'll play around next with the custom memory id idea.
Once we are all happy with the demo flow, I'll fix README

sberyozkin:
@geoand I prototyped a custom `DefaultMemoryIdProvider` and in this demo I'd like to do something simple, since no WebSockets are involved,  like adding a boolean property indicating if the customer has already issued a given query before (not 100% sure how to do it yet :-), right now I'm just hoping the existing chat history can be correctly detected), and then later in the secure variation of the csv chat box demo, use the security identity info in the memory id to impact the SQL query.

But the following error is becoming quite persistent:
```
Resulted in: java.lang.RuntimeException: dev.ai4j.openai4j.OpenAiHttpException: {
  "error": {
    "message": "Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'.",
    "type": "invalid_request_error",
    "param": "messages.[1].role",
    "code": null
  }
}
	at dev.langchain4j.internal.RetryUtils$RetryPolicy.withRetry(RetryUtils.java:195)
	at dev.langchain4j.internal.RetryUtils.withRetry(RetryUtils.java:229)
	at dev.langchain4j.model.openai.OpenAiChatModel.generate(OpenAiChatModel.java:153)
	at dev.langchain4j.model.openai.OpenAiChatModel.generate(OpenAiChatModel.java:118)
	at dev.langchain4j.model.chat.ChatLanguageModel_XNMsOaekknG7BdNZ5YSUkjh1SqE_Synthetic_ClientProxy.generate(Unknown Source)
	at io.quarkiverse.langchain4j.runtime.aiservice.AiServiceMethodImplementationSupport.doImplement(AiServiceMethodImplementationSupport.java:228)
	at io.quarkiverse.langchain4j.runtime.aiservice.AiServiceMethodImplementationSupport.implement(AiServiceMethodImplementationSupport.java:97)
	at io.quarkiverse.langchain4j.sample.FraudDetectionAi$$QuarkusImpl.detectAmountFraudForCustomer(Unknown Source)
	at io.quarkiverse.langchain4j.sample.FraudDetectionAi$$QuarkusImpl_Subclass.detectAmountFraudForCustomer$$superforward(Unknown Source)
	at io.quarkiverse.langchain4j.sample.FraudDetectionAi$$QuarkusImpl_Subclass$$function$$1.apply(Unknown Source)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:73)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:62)
	at io.smallrye.faulttolerance.FaultToleranceInterceptor.lambda$syncFlow$3(FaultToleranceInterceptor.java:253)
	at io.smallrye.faulttolerance.core.InvocationContext.call(InvocationContext.java:20)
	at io.smallrye.faulttolerance.core.Invocation.apply(Invocation.java:29)
	at io.smallrye.faulttolerance.core.timeout.Timeout.doApply(Timeout.java:55)
	at io.smallrye.faulttolerance.core.timeout.Timeout.apply(Timeout.java:30)
	at io.smallrye.faulttolerance.FaultToleranceInterceptor.syncFlow(FaultToleranceInterceptor.java:255)
	at io.smallrye.faulttolerance.FaultToleranceInterceptor.intercept(FaultToleranceInterceptor.java:182)
	at io.smallrye.faulttolerance.FaultToleranceInterceptor_Bean.intercept(Unknown Source)
	at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:42)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.perform(AroundInvokeInvocationContext.java:30)
	at io.quarkus.arc.impl.InvocationContexts.performAroundInvoke(InvocationContexts.java:27)
	at io.quarkiverse.langchain4j.sample.FraudDetectionAi$$QuarkusImpl_Subclass.detectAmountFraudForCustomer(Unknown Source)
	at io.quarkiverse.langchain4j.sample.FraudDetectionAi$$QuarkusImpl_ClientProxy.detectAmountFraudForCustomer(Unknown Source)
	at io.quarkiverse.langchain4j.sample.FraudDetectionResource.detectBaseOnAmount(FraudDetectionResource.java:39)
	at io.quarkiverse.langchain4j.sample.FraudDetectionResource_Subclass.detectBaseOnAmount$$superforward(Unknown Source)
	at io.quarkiverse.langchain4j.sample.FraudDetectionResource_Subclass$$function$$2.apply(Unknown Source)
```

Can you please, when you have a few mins, have a quick look at `FraudDetectionAi` and both `TransactionRepository` and `CustomerRepository`, does something look wrong ? I've seen this error on the web, but I'm not sure yet where to start digging, thanks

sberyozkin:
It is transient though, after some delay, I've restarted and I'm getting an interesting response, `There are no recent transactions for the customer John Doe with the email john.doe@example.com.`, and then later I got a normal response.
Probably related to the OpenAi session left hanging after some restarts, so may be some extra `close` is needed.

sberyozkin:
In any case, I'm moving away from using tools in favor of ContentRetriever, yet to be implemented similarly to how it is done in the csv chatbot demo, since it gives a nice option to use `Query.getMetadata().getMemoryId`, but I'm keeping it as a separate commit in case I get stuck :-)

jmartisk:
Gonna look into this one in a bit!

samples/secure-fraud-detection/README.md
```diff
@@ -0,0 +1,127 @@
+# Secure Fraud Detection Demo
+
+This demo showcases the implementation of a simple fraud detection system using LLMs (GPT-4 in this case) for users authenticated with Google.
```
jmartisk, line 3: AFAIU it's configured to use `gpt-3.5-turbo` (the default, unless you override `quarkus.langchain4j.openai.chat-model.model-name`)

geoand: Correct

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/Setup.java
```diff
@@ -0,0 +1,64 @@
...
+            var customer2 = new Customer();
+            customer2.name = "Georgios Andrianakis";
+            customer2.email = "geoand@mail.com";
+            customer2.transactionLimit = 1000;
+            customers.persist(customer2);
+
+            var customer3 = new Customer();
+            customer3.name = "Sergey Beryozkin";
+            customer3.email = "sberyozkin@gmail.com";
+            customer1.transactionLimit = 500;
```
jmartisk, line 42: You wanted `customer3` here?

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/Setup.java
```diff
@@ -0,0 +1,64 @@
...
+    public static String getARandomCity() {
+        return CITIES.get(new Random().nextInt(CITIES.size()));
+    }
+
+    @Transactional
+    public void init(@Observes StartupEvent ev, CustomerRepository customers, TransactionRepository transactions) {
+        Random random = new Random();
+        if (customers.count() == 0) {
+            var customer1 = new Customer();
+            customer1.name = "Clement Escofier";
```
jmartisk, line 28: Missing one f in Escoffier :laughing: unless the name similarity is purely coincidental, of course!

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/Setup.java
```diff
@@ -0,0 +1,64 @@
...
+            customers.persist(customer2);
+
+            var customer3 = new Customer();
+            customer3.name = "Sergey Beryozkin";
+            customer3.email = "sberyozkin@gmail.com";
+            customer1.transactionLimit = 500;
+            customers.persist(customer3);
+        }
+
+        transactions.deleteAll(); // Delete all transactions
```
jmartisk, line 46: If we delete transactions, for consistency should we delete all customers too?

samples/secure-fraud-detection/README.md
```diff
@@ -0,0 +1,127 @@
...
+To enable fraud detection, we provide the LLM with access to customer and transaction data through two Panache repositories:
+
+- [CustomerRepository.java](./src/main/java/io/quarkiverse/langchain4j/samples/CustomerRepository.java)
+- [TransactionRepository.java](./src/main/java/io/quarkiverse/langchain4j/samples/TransactionRepository.java)
+
+### AI Service
+
+This demo leverages the AI service abstraction, with the interaction between the LLM and the application handled through the AIService interface.
+
+The interface uses specific annotations to define the LLM and the tools to be used:
```
jmartisk, line 30: We should mention which interface we're talking about (`io.quarkiverse.langchain4j.sample.FraudDetectionAi`) and the snippets below look different from the actual code, so they need updating

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/FraudDetectionContentRetriever.java
```diff
@@ -0,0 +1,30 @@
...
+
+	@Inject
+	TransactionRepository transactionRepository;
+
+	@Inject
+	CustomerRepository customerRepository;
+
+    @Override
+    public List<Content> retrieve(Query query) {
+        JsonWebToken idToken = (JsonWebToken)query.metadata().chatMemoryId();
```
jmartisk, line 26: I'm puzzled as to what the content retriever should actually do. .With this, I just get `java.lang.ClassCastException: class java.lang.String cannot be cast to class org.eclipse.microprofile.jwt.JsonWebToken (java.lang.String is in module java.base of loader 'bootstrap'; org.eclipse.microprofile.jwt.JsonWebToken is in unnamed module of loader io.quarkus.bootstrap.classloading.QuarkusClassLoader @78b1cc93)` on this line

samples/secure-fraud-detection/README.md
sberyozkin: I haven't started working on README yet, thanks for pointing it out though, will keep it in mind

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/Setup.java
sberyozkin: Man, that was a bad mistake, hope Clement did not see it :-)

samples/secure-fraud-detection/README.md
```diff
@@ -0,0 +1,127 @@
+# Secure Fraud Detection Demo
+
+This demo showcases the implementation of a simple fraud detection system using LLMs (GPT-4 in this case) for users authenticated with Google.
+
+## The Demo
+
+### Setup
```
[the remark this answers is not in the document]
sberyozkin: Definitely, README is just a copy and paste right now from the original fraud detection demo

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/FraudDetectionContentRetriever.java
sberyozkin: It is converted to String somewhere in transit, looking into it

sberyozkin: Hi @jmartisk @geoand, OK, I'm just doing a simple  User name/email conversion from String to the `UserNameAndEmail` bean and it is good for now. Working on Readme next and it should be ready for review

sberyozkin:
The custom `ContentRetriever` has a request context activated to be able to read from Panache repositories.

I was hoping that may be I can add `@Authenticated` there as well and get `SecurityIdentity` injected, instead of using a custom memory id to capture a minimum session state in the form of the authenticated user's name and email, like
```
@ActivateRequestContext
@Authenticated
List<Content> retrieve(Query query) {}
```

But it fails with 

```
jakarta.enterprise.context.ContextNotActiveException: RequestScoped context was not active when trying to obtain a bean instance for a client proxy of CLASS bean [class=io.quarkus.vertx.http.runtime.CurrentVertxRequest, id=0_6n6EmChCiiDdd8HelptG_A0AE]
	- you can activate the request context for a specific method using the @ActivateRequestContext interceptor binding
	at io.quarkus.arc.impl.ClientProxies.notActive(ClientProxies.java:70)
	at io.quarkus.arc.impl.ClientProxies.getSingleContextDelegate(ClientProxies.java:30)
	at io.quarkus.vertx.http.runtime.CurrentVertxRequest_ClientProxy.arc$delegate(Unknown Source)
	at io.quarkus.vertx.http.runtime.CurrentVertxRequest_ClientProxy.getOtherHttpContextObject(Unknown Source)
	at io.quarkus.resteasy.reactive.server.runtime.QuarkusCurrentRequest.get(QuarkusCurrentRequest.java:21)
	at org.jboss.resteasy.reactive.server.core.CurrentRequestManager.get(CurrentRequestManager.java:8)
	at io.quarkus.resteasy.reactive.server.runtime.StandardSecurityCheckInterceptor.intercept(StandardSecurityCheckInterceptor.java:39)
	at io.quarkus.resteasy.reactive.server.runtime.StandardSecurityCheckInterceptor_AuthenticatedInterceptor_Bean.intercept(Unknown Source)
	at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:42)
	at io.quarkus.arc.impl.AroundInvokeInvocationContext.perform(AroundInvokeInvocationContext.java:30)
	at io.quarkus.arc.impl.InvocationContexts.performAroundInvoke(InvocationContexts.java:27)
	at io.quarkiverse.langchain4j.sample.FraudDetectionContentRetriever_Subclass.retrieve(Unknown Source)
	at io.quarkiverse.langchain4j.sample.FraudDetectionContentRetriever_ClientProxy.retrieve(Unknown Source)
```

@michalvavrik, does it ring any bell to you ?

It is strange because `@ActivateRequestContext` is there, it looks like it is lost during the context propagation...

So I guess for now, using a custom memory id is OK, in fact, it can make sense to combine eventually both `@Authenticated` and the custom memory id to compare the injected identity with the info from this memory id to assert within the retriever that the AI conversation is essentially bound to the current identity to avoid the retriever is being  called outside of the current security context. So I'd say the above exception is not a big problem for now as `@Authenticated` is already protecting the entry  `Login` and `FraudDetectionResource`  resources.

michalvavrik:
> It is strange because @ActivateRequestContext is there, it looks like it is lost during the context propagation...

`StandardSecurityCheckInterceptor` has priority `PLATFORM_BEFORE` while ActivateRequestContextInterceptor has `PLATFORM_BEFORE + 100`. Quarkus Security extension standard security interceptors have `LIBRARY_BEFORE` as long as you fix this bug with anything lesser than `LIBRARY_BEFORE`, it will be fine. Sorry, I caused this (quite a long time ago now). You want to skip Quarkus Security interceptors so you need to run the interceptors in Quarkus REST before them.

michalvavrik:
> So I guess for now, using a custom memory id is OK, in fact, it can make sense to combine eventually both @Authenticated and the custom memory id to compare the injected identity with the info from this memory id to assert within the retriever that the AI conversation is essentially bound to the current identity to avoid any chances of the retriever being called outside of the current security context. So I'd say the above exception is not a big problem for now as @Authenticated is already protecting the entry Login and FraudDetectionResource resources.

I know practically nothing about this project and code, but let me note that you will not be able to use standard security annotations on the `FraudDetectionContentRetriever` or down the stream because the exception you are getting means that CDI request context has not been activated by Quarkus REST (e.g. at that point, it's already deactivated for some reason and `ActivateRequestContext` will activate brand new context). Which means no `SecurityIdentity` unless I am missing something custom you do there.

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/FraudDetectionContentRetriever.java
jmartisk: @sberyozkin I am still getting the same error and the method, even if the error were fixed, would return an empty list anyway. Do I need to change something to get it working?

sberyozkin:
Thanks @michalvavrik for the feedback, getting to checking comments only now... OK, I'll open a Quarkus issue to consider what can be done with the priorities.
I've only tried it to consider optimizing the way the custom identity representation is passed through the various Quarkus LangChain4j beans, for now using a custom memory id only is fine IMHO.

samples/secure-fraud-detection/src/main/resources/application.properties
```diff
@@ -0,0 +1,12 @@
...
+quarkus.langchain4j.openai.chat-model.temperature=0
+quarkus.langchain4j.openai.api-key=${OPENAI_API_KEY}
+
+quarkus.oidc.provider=google
+quarkus.oidc.client-id=${GOOGLE_CLIENT_ID}
+quarkus.oidc.credentials.secret=${GOOGLE_CLIENT_SECRET}
+quarkus.oidc.authentication.redirect-path=/login
+
+#quarkus.langchain4j.openai.log-requests=true
+#quarkus.langchain4j.openai.log-responses=true
```
jmartisk, line 11: I'd suggest uncommenting these two properties, they are generally quite useful for experimenting

jmartisk:
Ok I can get it working now.  I'm not super happy about users having to manually update the `Setup` class to add a user that exactly matches their Google account.  If one logs in as some other user, I get a NPE in `CustomerRepository.getTransactionLimit` because the user doesn't exist in the database (this error should definitely be handled better, it should show an error in the UI or something, or at least re-throw the exception with a meaningful message). 

I'm thinking how to make it easier.. Could we maybe allow the user to specify an email address (and name) that should be registered during the Setup, as a config property?

sberyozkin:
Thanks @jmartisk, makes sense to make it easier for users to register emails,  may be with a system property,  will have a look next week

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/CustomerRepository.java
```diff
@@ -0,0 +1,16 @@
...
+import io.quarkus.hibernate.orm.panache.PanacheRepository;
+import jakarta.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+public class CustomerRepository implements PanacheRepository<Customer> {
+    /*
+     * Transaction limit for the customer.
+     */
+	public int getTransactionLimit(String customerName, String customerEmail) {
+		return find("name = ?1 and email = ?2",
```
jmartisk, line 12: Depending on both name AND email can only lead to additional problems without adding any value, email should be enough, no?

sberyozkin: @jmartisk This is meant to show how several pieces from the current Security Identity can be passed around and used for something, like formulating a query. It is not meant as some concrete prod-quality recommendation (and I'm not convinced that there is 0 value in combining both, but I don't really mind, name can be dropped if you prefer).

What having both values does show though is that having a custom memory id transformed to String at the source makes it more difficult to restore though values, which is why I meant to ask @geoand, once it is approved otherwise, if some workaround can be found, but it can be discussed later for sure

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/SecureMemoryIdProvider.java
```diff
@@ -0,0 +1,75 @@
...
+        return null;
+    }
+
+    public static class UserNameAndEmail {
+        private final String name;
+        private final String email;
+        UserNameAndEmail(String name, String email) {
+            this.name = Objects.requireNonNull(name);
+            this.email = Objects.requireNonNull(email);
+		}
```
geoand, line 42: This definitely needs some formatting

samples/secure-fraud-detection/README.md
```diff
@@ -0,0 +1,119 @@
+# Secure Fraud Detection Demo
+
+This demo showcases the implementation of a secure fraud detection system whuich is available only to users authenticated with Google.
```
geoand, line 3:
```suggestion
This demo showcases the implementation of a secure fraud detection system which is available only to users authenticated with Google.
```

geoand (commented):
I've been thinking about this some more. 

IIUC, the intent here is to show that the AiService can access the logged in user when retreving data, correct?
If so, I think that using the `MemoryIdProvider` is the not best abstraction. 

Would it not be far easier to inject the logged in user in `FraudDetectionRetrievalAugmentor`?

sberyozkin:
Thanks @geoand,

>Would it not be far easier to inject the logged in user in FraudDetectionRetrievalAugmentor?

Yeah, good point, I tried it earlier, to inject it directly to `FraudDetectionContentRetriever` but RequestContext carrying the security identity was not available.

I can try to inject it into a custom retrieval augmentor, but if that works, how would I pass it to the content retriever ?

samples/secure-fraud-detection/README.md
sberyozkin: Thanks @geoand, I spotted it earlier and modified README alongside other updates locally so far, so this suggestion won't apply cleanly, let me resolve it

geoand:
> Yeah, good point, I tried it earlier, to inject it directly to FraudDetectionContentRetriever but RequestContext carrying the security identity was not available.

You are right, I see why this is the case... 

Unfortunately I don't have any good solutions for you  at the moment, but I do see how we add something to LangChain4j to make  this possible.

Essentially what I have in mind is to enhance `dev.langchain4j.rag.query.Metadata` with a `Map<String, Object>` where custom data can be added. 
Then we would also need a new concept like `MetadataEnhancer` which we could then use in Quarkus to put in whatever we want, in this case the user id or email or whatever. 

Then your `ContentRetriever` could retrieve the information from `query.getMetadata()`;

What I don't like about this is that it would force user code (the `ContentRetriever` to know specific keys). 

An alternative approach would be for us to write our own version of `DefaultRetrievalAugmentor` which would not suffer from the problem of not passing the context (as it take steps to have that done properly).

WDYT @jmartisk @sberyozkin ?

sberyozkin:
>Essentially what I have in mind is to enhance dev.langchain4j.rag.query.Metadata with a Map<String, Object> where custom data can be added.

Sure, that can definitely do it, some well known key properties like `io.quarkus.security.SecurityIdentity` can be doc-ed

>An alternative approach would be for us to write our own version of DefaultRetrievalAugmentor which would not suffer from the problem of not passing the context (as it take steps to have that done properly).

I'm not sure yet as I don't know much about the mechanics involved, on the basic level having an extra Metadata Map looks like it can work better as the identity may be needed across several pieces where Query is available, not only at the RAG level.

I was actually thinking about some workarounds on the way to the office :-) I was still thinking in terms of a custom memory id since the authenticated/logged in user's session lifetime matches the secure memory id lifetime.
So what I had in mind is that the project introduces `CompositeMemoryIdBean` bean and now, the core code, before converting a provided memory id to String, it checks if has a marker like `CompositeMemoryBeanExpected` (or may be simply if it an instance of String or not - if not - it is something more involved than String). And if the composite memory id bean is expected, then CompositeMemoryIdBean will wrap and carry the original custom memory id object around, its `toString()` method will be done the way it is done now in the core (memory id object string rep plus an AI service class/method name), and then whichever provider needs it, it casts a memory id object to `CompositeMemoryIdBean` after an instanceof check,  and gets an original memoryId object as `getMemoryId`

May be it is a bit hacky, not sure, though it would probably fit closely enough the expectation that the memory id can be any object. I think I also like the idea of the extra metadata map.
This

geoand:
> I was actually thinking about some workarounds on the way to the office :-) I was still thinking in terms of a custom memory id since the authenticated/logged in user's session lifetime matches the secure memory id lifetime.

This is the one solution I don't want because MemoryId is really a different concept. It's fine to populate it automatically, but it's not fine to use it's value to do something else.

sberyozkin:
@geoand I'm happy enough to pursue adding extra Map to `Metadata`, I can probably suggest a PR to langchain4j, it looks it can be of general interest... How would we set the extra bits into this Map though at the quarkus-langchain4 level ?

geoand:
> it looks it can be of general interest.

Yeah, I think so to, but there also needs to be a way to add metadata, something like the `MetadataEnhanced` I suggested above.

sberyozkin:
@geoand OK, then, it will certainly cover the RAG flow. Can you clarify please, when you get a sec, why RequestContext is not available in the custom augmentor or contemt retriever, like you said, that would definitely be a simplest solution, if it were available. (I'm also thinking about making the identity visible in other places like custom tools and other custom non-RAG related providers).
I'm pretty sure your idea of adding an extra map to Metadata can be generally useful though, starting from it would be great, so I'll look into opening a PR in langchain4j

geoand:
>  Can you clarify please, when you get a sec, why RequestContext is not available in the custom augmentor or content retrieve

Because the implementation in upstream LangChain4j uses a `CompletableFuture` and it's own executor in order to do calls in parallel. 
Now that I think about it, it probably makes sense to see updating `DefaultRetrievalAugmentor` to use our own  Executor which we would feed with the one from context propagation works. Do you want to give that a shot?

samples/secure-fraud-detection/src/main/java/io/quarkiverse/langchain4j/sample/CustomerRepository.java
sberyozkin: Let me resolve it for now, happy to revisit a bit later once other issues are sorted out, with PR still being open :-)

geoand:
> Now that I think about it, it probably makes sense to see updating DefaultRetrievalAugmentor to use our own Executor which we would feed with the one from context propagation works. Do you want to give that a shot?

Actually, it should be possible to do this even now, no changes to LangChain4j needed.

sberyozkin:
@geoand Now I'm really happy, the custom memory id provider has gone :-), secure authenticated access to the content retriever is done, thanks for the ManagedExecutor idea. I think it can probably make sense to doc it separately, I can work a bit later on a doc dedicated to securing various AI service parts...

I think this PR is now quite ready to go once yourself and @jmartisk happy enough with it in general

geoand:
Very nice!

+1 on getting this in - but let's wait for @jmartisk

samples/secure-fraud-detection/README.md
```diff
@@ -0,0 +1,118 @@
...
+All you need to do is to register an application with Google, follow steps listed in the [Quarkus Google](https://quarkus.io/guides/security-openid-connect-providers#google) section.
+Name your Google application as `Quarkus LangChain4j AI`, and make sure an allowed callback URL is set to `http://localhost:8080/login`.
+Google will generate a client id and secret, use them to set `quarkus.oidc.client-id` and `quarkus.oidc.credentials.secret` properties.
+
+## Running the Demo
+
+To run the demo, use the following command:
+
+```shell
+mvn quarkus:dev -Dname="Firstname Familyname" -Demail=someuser@gmail.com
```
jmartisk, line 112: Can we be even a bit more explicit that the `name` and `email` have to exactly match the user's google account (maybe say that this is used to automatically add a user entry into the app's database)?

jmartisk:
Guys I see you were discussing passing some extra metadata around a RAG pipeline, please check https://github.com/langchain4j/langchain4j/issues/1122, that is related

samples/secure-fraud-detection/README.md
sberyozkin: @jmartisk This is all mentioned in the Setup section above, starting from line 10:

```
The demo requires that your Google account's full name and email are configured.
You can use system or env properties, see `Running the Demo` section below.
```

following the info that transactions are generated.

I think users will understand the exact match is required. Can you suggest some further edits please there ?

Note, if they made a mistake during the registration, the error page will show them what exactly they should register by listing a full name and email, I added that page yesterday.

sberyozkin:
@jmartisk 

>Guys I see you were discussing passing some extra metadata around a RAG pipeline, please check https://github.com/langchain4j/langchain4j/issues/1122, that is related

Thanks for the link, it is interesting, luckily, at least for the security identity, it can be injected RAG related parts, but I'm not sure  yet if it can be injected into Tools, something I can try in the next demo for the secure web socket based sql chatbot one, so that is indeed a very related discussion, thanks

jmartisk (approved):

PR #546, opened 2024-05-06
Move the parsers out of the core module

geoand:
The reason we don't want these in the core module
is because not every application needs the doc
parsing functionality (which adversely affects the application size).
For example, a simple AiService that uses OpenAI  has at least 30MB 
added to binary size by just including the POI and BouncyCastle dependencies.

We might need to make this module (or some derivatives of it) more explicit
in the docs and / or by using conditional dependencies.


cescoffier (approved):
should the commit be squashed?

geoand:
The second commit was actually a result of older changes so it's pretty much independent of the main change here

PR #548, opened 2024-05-07
Neo4j embedding store

jmartisk:


jmartisk:
CC @adriens if you want to give it a look already. You can set `quarkus-neo4j.version` to 3.9.0 and it will compile and work, but then to use dev services, you'll need to add  `quarkus.neo4j.devservices.image-name=neo4j:5` in `application.properties`. I'd like to wait for a quarkus-neo4j version that already has 5 as the default version and avoid overriding it on our side

neo4j/deployment/src/test/java/io/quarkiverse/langchain4j/neo4j/Neo4jEmbeddingStoreTest.java
```diff
@@ -0,0 +1,60 @@
...
+    // Normally, we would just inject the model into the test
+    @BeforeAll
+    public static void initEmbeddingModel() {
+        embeddingModel = new AllMiniLmL6V2QuantizedEmbeddingModel();
+    }
+
+    @Override
+    protected void clearStore() {
+        try (Session session = driver.session()) {
+            session.run("MATCH (n) DETACH DELETE n");
```
michael-simons, line 47:
```suggestion
            session.run("MATCH (n) CALL { WITH n DETACH DELETE n } IN TRANSACTIONS").consume();
```

Depending on the store size, the default detach delete might be slow.
Also as a good advice, just consume the `run` result, to make sure things actually are in the state you want.

neo4j/runtime/src/main/java/io/quarkiverse/langchain4j/neo4j/runtime/Neo4jEmbeddingStoreRecorder.java
```diff
@@ -0,0 +1,39 @@
...
+@Recorder
+public class Neo4jEmbeddingStoreRecorder {
+
+    public Function<SyntheticCreationalContext<Neo4jEmbeddingStore>, Neo4jEmbeddingStore> embeddingStoreFunction(
+            Neo4jRuntimeConfig config) {
+        return new Function<>() {
+            @Override
+            public Neo4jEmbeddingStore apply(SyntheticCreationalContext<Neo4jEmbeddingStore> context) {
+                Neo4jEmbeddingStore.Neo4jEmbeddingStoreBuilder builder = Neo4jEmbeddingStore.builder();
+                Driver driver = context.getInjectedReference(Driver.class, new Default.Literal());
```
michael-simons, line 23: Did not know that trick, good stuff.

neo4j/deployment/src/test/java/io/quarkiverse/langchain4j/neo4j/Neo4jEmbeddingStoreTest.java
jmartisk: Cool, thanks

neo4j/runtime/src/main/java/io/quarkiverse/langchain4j/neo4j/runtime/Neo4jRuntimeConfig.java
```diff
@@ -0,0 +1,78 @@
...
+     * The query to use when retrieving embeddings. This query has to return the following columns:
+     * <ul>
+     * <li>metadata</li>
+     * <li>score</li>
+     * <li>column of the same name as the 'id-property' value</li>
+     * <li>column of the same name as the 'text-property' value</li>
+     * <li>column of the same name as the 'embedding-property' value</li>
+     * </ul>
+     */
+    @WithDefault("RETURN properties(node) AS metadata, " +
```
geoand, line 71: I assume we can use Java's multi line string here, it will look much better

geoand (commented):
I added a couple small comments

geoand (approved):

PR #556, opened 2024-05-09
Automatically update dependencies for samples during releases

jmartisk:


geoand (approved):
Nice!

geoand:
This [almost worked](https://github.com/quarkiverse/quarkus-langchain4j/commit/eba307d544eaf41383dcc745cd1f77e5f2e34c3e).

Instead of updating the quarkus langchain4j version it updated the project version itself

iocanel:
I think that the issue is that it uses `-pl` (that works with artifactId) instead of `-f` (that works with path).

iocanel:
Let me try to fix that.

Issue #555, opened 2024-05-09
vertex-ai-gemini bad configuration error message


exaV:
Hi, I was eagerly awaiting the release of `quarkus-langchain4j-vertex-ai-gemini`, so props for the release 👍 .

Unfortunately I get a configuration error. I added a log excerpt below. The key "quarkus.langchain4j.vertexai.devopi.location" is logged as missing during startup and later also mentioned in the error message as required. I later realized that the config key had to be `quarkus.langchain4j.vertexai.gemini.devopi.location` but the error message did not help with that.

```
2024-05-09 10:21:59,013 WARN  [io.qua.config] (main) Unrecognized configuration key "quarkus.langchain4j.vertexai.devopi.location" was provided; it will be ignored; verify that the dependency extension for this configuration is set or that you did not make a typo
2024-05-09 10:21:59,016 WARN  [io.qua.config] (main) Unrecognized configuration key "quarkus.langchain4j.vertexai.devopi.projectId" was provided; it will be ignored; verify that the dependency extension for this configuration is set or that 
...
Caused by: io.smallrye.config.ConfigValidationException: Configuration validation failed:
	SRCFG00014: The config property quarkus.langchain4j.vertexai.devopi.location is required but it could not be found in any config source
	at io.quarkiverse.langchain4j.vertexai.runtime.gemini.VertexAiGeminiRecorder.chatModel(VertexAiGeminiRecorder.java:26)
	at io.quarkus.deployment.steps.VertexAiGeminiProcessor$generateBeans1607297205.deploy_0(Unknown Source)
	at io.quarkus.deployment.steps.VertexAiGeminiProcessor$generateBeans1607297205.deploy(Unknown Source)
	... 8 more
```

my inital config that lead to the error:
```
  langchain4j:
    devopi:
      chat-model:
        provider: vertexai-gemini

    vertexai:
      devopi:
        projectId: "<my-project>"
        location: "<region>"
      gemini:
        chat-model:
          model-id: "gemini-1.0-pro"

```

config that worked after  I later realized that the config path had to be  `quarkus.langchain4j.vertexai.gemini.devopi.location`:

```
  langchain4j:
    devopi:
      chat-model:
        provider: vertexai-gemini

    vertexai:
      gemini:
        devopi:
          projectId: "<my-project>"
          location: "<region>"
        chat-model:
          chat-model: # why is chat-model repeated?
            model-id: "gemini-1.0-pro"

```

[...]

PR #558, opened 2024-05-09
Fix configuration error in Gemini

exaV:
The path has to be the same as in `LangChain4jVertexAiGeminiConfig`

- Fixes: https://github.com/quarkiverse/quarkus-langchain4j/issues/555


geoand (approved):
Thanks!

Issue #563, opened 2024-05-11
Email me a poem is not very friendly


iocanel:
When opening `http://localhost:8080` for email me a poem, users get a 404 that informs them that they actually have to go to `http://localhost:8080/email-me-a-poem`. Also, if they haven't read the readme ( **cough** **cough**), they won't understand why they sample doesn't work.

PR #564, opened 2024-05-11
Add minimal html  UI for email me a peom

iocanel:
Resolves: #563

So, the pull request adds a minimal index.html that provides a UI for Email me a poem.
It also catches errors and hints users.

Note, the UI is as minimal as it gets (no vaadin components here) for minimalism shake. If we prefer to have a consistent UI all across our samples, I am all for aligning.

**edit**: The PR also improves the prompt, so that the poem is actually displayed in the response.

### DIFF
--- a/samples/email-a-poem/src/main/java/io/quarkiverse/langchain4j/sample/MyAiService.java
+++ b/samples/email-a-poem/src/main/java/io/quarkiverse/langchain4j/sample/MyAiService.java
@@ -15,7 +15,7 @@ public interface MyAiService {
      * @return the poem
      */
     @SystemMessage("You are a professional poet")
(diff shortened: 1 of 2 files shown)


geoand (approved):
Makes perfect sense!

geoand:
There seems to be some bug in the UI as I can't see to type anything into it

iocanel:
> There seems to be some bug in the UI as I can't see to type anything into it
Let me check!

iocanel:
> There seems to be some bug in the UI as I can't see to type anything into it

You are not supposed to type anything. It's where the poem is displayed when you click the send me a poem button.

iocanel:
Which seems to be an indication that this is confusing.

geoand:
Ok yeah, in that case it is pretty confusing 😉

PR #565, opened 2024-05-11
Drop uneeded MyChatModelSupplier from email-me-a-peom

iocanel:
I think that the MyChatModelSupplier is not really used here, so it needs to go.


geoand (approved):
🎯

PR #571, opened 2024-05-11
Refactor jokes jbang bot to be picocli based

iocanel:
The pull request changes the jbang jokes bot, to be a cli application  instead of a web application.

The reasoning is that we already have a lot of samples going the web path and it seems to me like a missed opportunity to show case to also demonstrate how it can be done in the CLI.

Now, I understand that the reasoning might be to implement something with the fewer lines of code possible, but I argue that the code is hard to understand.

### DIFF
--- a/samples/jbang-joke-bot/README.md
+++ b/samples/jbang-joke-bot/README.md
@@ -1,4 +1,4 @@
-# JBang sample aka getting an AI-powered bot in 13 lines of Java
+# JBang sample aka getting an AI-powered bot in 20 lines of Java
 
 To run the sample, you need JBang installed. If you don't have it, choose
 one of the installation options from the [JBang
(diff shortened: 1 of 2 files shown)


iocanel:
@maxandersen

jmartisk:
I'm not completely sure I like this TBH, first it has gotten bigger (from 13 to 20 lines), second, unless it showcases some actual CLI stuff (like parameters) then CLI is unnecessary (and the point is to show a LLM app, not a CLI app). But I'll leave it up to you guys

iocanel:
Yeah, the number of lines has increased, but it's mostly metadata, annotations and imports. One the other hand it's less dense and thus way easier to understand. It's also much simpler to run, as it does in a single step.

It's  also a more realistic use case of jbang scripting.

cescoffier (approved):

cescoffier:
@iocanel Can you fix the conflict?

PR #582, opened 2024-05-14
Improve email-me-a-poem ui

iocanel:
As discussed in #564 the UI is a bit confusing, as the textarea used for showing the response implies that the user is supposed to add something there.

This PR changes the texteara, with a span (which is invisible by default) and it populates it with the response when available. It also adds a `Working...` message when the button is pressed to further improve the experience.


iocanel:
@geoand: let me know how this works for you.

geoand:
Much better, thanks!

geoand (approved):

PR #589, opened 2024-05-15
Make token limit configuration for Ollama optional

geoand:
This way Ollama can decide what the default should be (if any)


iocanel (approved):
I verify that it does solve the issue I was experiencing.
Code also looks good to me.

geoand:
Thanks!

Issue #442, opened 2024-04-03
Streaming chat in DevUI


geoand:
Would it be possible to use a `StreamingChatModel` in the DevUI if it's accessible?
This would allow to stream the response back

geoand:
@jmartisk do you think this is possible without too much work?

jmartisk:
What would that mean exactly, that the response from the model is a Multi and we would forward the events from the Multi to the UI as they arrive, over a websocket?

iocanel:
I can give it a shot.

iocanel:
I have something is working. However, it will only work when rag is disabled (and it's true by default).

Should we add a message / alert to let the user know when streaming chat is not enabled and why it's not enabled?
Would you think that a vaadin alert or a console.log is more appropriate for this?

geoand:
> Should we add a message / alert to let the user know when streaming chat is not enabled and why it's not enabled?

I think that's a good idea

PR #590, opened 2024-05-15
Support streaming in dev-ui chat

iocanel:
The pull request introduces a new option (streaming chat). The option is enabled when model supports streaming chat model and rag is disabled (streaming chat can't work with rag atm). When user enables streaming chat. Responses from the model are streamed to the devui chat screen.

- Resolves: #442

### DIFF
--- a/core/deployment/src/main/resources/dev-ui/qwc-chat.js
+++ b/core/deployment/src/main/resources/dev-ui/qwc-chat.js
@@ -10,6 +10,7 @@ import '@vaadin/progress-bar';
 import '@vaadin/text-field';
 import '@vaadin/icon';
 import '@vaadin/icons';
+import 'qui-alert';
 import { JsonRpc } from 'jsonrpc';
 import { systemMessages } from 'build-time-data';
 
@@ -69,7 +70,10 @@ export class QwcChat extends LitElement {
         _systemMessages: {state: true},
         _systemMessageDisabled: {state: true},
         _ragEnabled: {state: true},
-        _showToolRelatedMessages: {state: true}
+        _streamingChatSupported: {state: true},
+        _streaminChatEnabled: {state: true},
+        _showToolRelatedMessages: {state: true},
+        _observer: {state:false},
     }
 
     constructor() {
@@ -83,8 +87,29 @@ export class QwcChat extends LitElement {
         this._unfilteredChatItems = [];
         this._chatItems = [];
         this.jsonRpc.reset({systemMessage: this._systemMessage});
+        this._streamingChatSupported = this.jsonRpc.isStreamingChatSupported();
+        this._streamingChatEnabled = this._streamingChatSupported && !this._ragEnabled;
     }
 
+    _connect() {
+    }
+
+    _disconnect() {
+      if (this._observer) {
+        this._observer.unsubscribe();
+      }
+    }
+
+    connectedCallback() {
+        super.connectedCallback();
+        this._connect();
+    }
+    
+    disconnectedCallback() {
+        this._disconnect();
+        super.disconnectedCallback();
+    }
+    
     render() {
         this._filterChatItems();
         return html`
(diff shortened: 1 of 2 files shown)


jmartisk:
I'm at a conference so I'll have time to test it later, but I'll just throw in that, in case of a timeout, or whatever other error from the LLM, are we sure that we don't get the history in the Dev UI desynchronized with the one stored in the actual chat memory? That's why the original synchronous variant of the `chat` method always returns the whole history - so the Dev UI can replace the whole history by that, and prevent desynchronization. I think this might be a little tricky to implement robustly when you just stream a single message.

iocanel:
> I'm at a conference so I'll have time to test it later, but I'll just throw in that, in case of a timeout, or whatever other error from the LLM, are we sure that we don't get the history in the Dev UI desynchronized with the one stored in the actual chat memory? That's why the original synchronous variant of the `chat` method always returns the whole history - so the Dev UI can replace the whole history by that, and prevent desynchronization. I think this might be a little tricky to implement robustly when you just stream a single message.

The memory gets modified when `onComplete` is called, with the whole message. In case of an error the memory is restored from the backup in a way similar to regular chat. Do you think we need to do more?

geoand:
I plan on going to do a `0.14.0` today, so we can leave this for `0.14.1` once @jmartisk gives his final approval

core/deployment/src/main/resources/dev-ui/qwc-chat.js
```diff
@@ -96,6 +121,15 @@ export class QwcChat extends LitElement {
             <div><vaadin-checkbox checked label="Enable Retrieval Augmented Generation (if a RetrievalAugmentor bean exists)"
                                   @change="${(event) => {
                                       this._ragEnabled = event.target.checked;
+                                      this._streamingChatEnabled = this._streamingChatEnabled && !this._ragEnabled;
+                                      this.render();
+                                  }}"/></div>
+            <div><vaadin-checkbox label="Enable Streaming Chat (if supported by model & rag disabled)"
+                                  class="${this.jsonRpc.isStreamingChatSupported() ? 'show' : 'hide'}"
```
jmartisk, line 128: This jsonrpc method should be called only once - this way you call it again every time the form is rendered

iocanel:
I updated the PR and added error handling in all possible ways I could imagine:

- using `orError`
- handling the error as part of `onNext`
- wrapping the whole thing in try / catch

Even though the error is reaching the UI as I can tell from the console, I can't somehow capture it in my javascript code.
@phillip-kruger any idea what I am doing wrong here?

iocanel:
@jmartisk: I checked what you do in `qwc-chat.js` and it's not applicable to this case. In this case json rpc returns a multi errors are not just catch using `catch`. They should be caught using `onError` but the errors are not propagated. @phillip-kruger mentioned that it most probably an issue in json rpc impl.

phillip-kruger:
I have not looked at this yet, but in the mean time, can you see if anything prints in :
1) The browser console ?
2) The dev-ui log (the json message tab)

jmartisk:
I'm now realizing, this will also cause problems when Tools are used, not just RAG, no? Because the streaming response from the server will not contain tool-related messages. :(
To fix that, we would have to check the chat memory for changes (for the tool messages to appear in it) while the streaming is still going on, and somehow hack them into the stream (for sending the augmented message with RAG, we'd actually have to do something similar).

geoand:
Yeah, that sounds right

iocanel:
@jmartisk: the issue you had with errors being swallowed has been fixed on the quarkus side of the house by @phillip-kruger and should be available in 3.11.1. Nothing additional needed here.

Now, regarding your comment about tools, the suggested solution is not clear to me. Is it something that is required for this PR or something we could add later on?

jmartisk:
It would be nice to have RAG and tools working properly with streaming, but if you don't want to do it I can have a look. I have some heavier refactoring planned anyway, to be able to support images.

jmartisk (approved):

PR #596, opened 2024-05-16
Fix update sample version when releasing

iocanel:
Follow up fix for: #556

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,13 +58,9 @@ jobs:
 
       - name: Update dependency versions for the samples
         run: |
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/email-a-poem
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/fraud-detection
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/review-triage
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/chatbot
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/csv-chatbot
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/cli-translator
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/chatbot-easy-rag
+          for pom in `find samples/ -name pom.xml`; 
+            ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set-property -Dproperty=quarkus-langchain4j.version -DnewVersion=${{steps.metadata.outputs.current-version}}  -f $(dirname "$pom"); 
+          done
           git commit -a -m "Update dependencies in samples"
 
       - name: Push changes to ${{github.base_ref}} branch


geoand:
Did you run this to see that is make the change we want (to the `quarkus-langchain4j.version` property and not the `version` of the project itself)?

jmartisk:
Huhm, how could I be so dumb to not notice..
Yeah, to update the property, you need to do
```
 ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:update-property -Dproperty=quarkus-langchain4j.version -DnewVersion=${{steps.metadata.outputs.current-version}} -DallowDowngrade=true -f samples/email-a-poem
```

.. for all samples

.github/workflows/release.yml
```diff
@@ -58,13 +58,13 @@ jobs:
...
+          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -f samples/email-a-poem
+          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -f samples/fraud-detection
+          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -f samples/review-triage
+          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -f samples/chatbot
+          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -f samples/csv-chatbot
+          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -f samples/cli-translator
+          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -f samples/chatbot-easy-rag
```
gsmet, line 67: Probably a good idea to use some bash to automate it?

```
for module in samples/*; do echo $module; done
```

Make sure you define the shell as bash.

jmartisk: I thought about that too, but there's one sample (the jbang one) that isn't maven based so we'd have to exclude it, and that felt ugly enough for me to persuade me to go with hardcoding the list of modules. But of course we can do it.

gsmet: If you insist:

```
for pom in `find samples/ -name pom.xml`; do module=$(dirname "$pom"); echo $module; done
```

geoand: @iocanel do you want to update the PR as per Guillaume's suggestion?

.github/workflows/release.yml
```diff
@@ -58,13 +58,9 @@ jobs:
...
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/fraud-detection
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/review-triage
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/chatbot
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/csv-chatbot
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/cli-translator
-          ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}} -pl samples/chatbot-easy-rag
+          for pom in `find samples/ -name pom.xml`; 
+            ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:set -DnewVersion=${{steps.metadata.outputs.current-version}}  -f $(dirname "$pom");
```
geoand, line 62: Shouldn't it be ` ./mvnw org.codehaus.mojo:versions-maven-plugin:2.16.2:update-property -Dproperty=quarkus-langchain4j.version  ...` ?

.github/workflows/release.yml
geoand: I have added a comment below

.github/workflows/release.yml
iocanel: Updated to use set-property.

geoand (approved):

PR #597, opened 2024-05-17
Properly configure for embedding model

csotiriou:
Fixes wrong configuration being passed while initializing Azure Embeddings client.


geoand (approved):
Thanks!

Issue #237, opened 2024-01-23
Azure configuration for multiple deployments


Yanick-Salzmann:
In my opinion there is an issue with the configuration for Azure. At the moment the following config options are possible (some more are there but relevant to my question here):
```yaml
quarkus.langchain4j.azure-openai:
    - resourceName
    - deploymentId
    - baseUrl

    - embeddingModel:
       - ...
    
    - chatModel:
       - ...  
```

The deploymentId will identify the model in Azure. With the configuration currently I am not able to configure a different model for embeddings and for chat, however my situation is the following, that under the resource name `my-company-playground-use` I have two deployments, one is `chat-completion` which is a gpt4 model and one is `embedding-model` which is an ada-002 model. I want to use the `chat-completion` deployment for the `chatModel` above and the deployment `embedding-model` for the `embeddingModel` above. This is not currently possible. I would need to have the possibility to configure:

```properties
quarkus.langchain4j.azure-openai.resourceName=mycompany-playground-use
quarkus.langchain4j.azure-openai.embedding-model.deploymentId=embedding-model
quarkus.langchain4j.azure-openai.chat-model.deploymentId=chat-model
```

I am willing to work on this, however I was wondering what your opinions are and/or if you have already had some considerations regarding that.

geoand:
I'd like to hear what @edeandrea and maybe even @agoncal have to say since I have done very little work on Azure AI.

edeandrea:
Does that limitation exist in upstream langchain4j?

geoand:
cc @langchain4j

edeandrea:
@geoand it looks like upstream langchain4j has separate endpoints for each of the different models...

https://github.com/langchain4j/langchain4j/blob/main/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java

https://github.com/langchain4j/langchain4j/blob/main/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiEmbeddingModel.java

etc

So it looks as if the Quarkus configuration around this should respect that. Maybe a global set of properties (like we have now) which can be overridden by similar properties on a model-by-model basis?

geoand:
Yeah, that makes sense

geoand:
@Yanick-Salzmann as you said you were willing to work on it, how do you feel about implementing what @edeandrea mentioned?

Yanick-Salzmann:
[...]

PR #598, opened 2024-05-17
Allow multi-deployment configuration for Azure OpenAI

csotiriou:
Relevant to #237, this will allow overriding the following properties per model in the Azure OpenAI configuration

- `endpoint`
- `resource-name`
- `deployment-name`

Example configuration that is now possible:

```properties
quarkus.langchain4j.azure-openai.ad-token=token
quarkus.langchain4j.azure-openai.api-version=2023-07-01-preview
quarkus.langchain4j.azure-openai.chat-model.enabled=true
quarkus.langchain4j.azure-openai.deployment-name=my-deployment-name
quarkus.langchain4j.azure-openai.embedding-model.enabled=true
quarkus.langchain4j.azure-openai.resource-name=name
quarkus.langchain4j.azure-openai.embedding-model.deployment-name=another-deployment-name
```

In ht e above example, all models in the parent configuration will have the `my-deployment-name`, but the Embedding model will have the `another-deployment-name` specifically as a name.


geoand (commented):
Very nice!

My only ask is to not use lambdas as we generally try to avoid them in runtime code

csotiriou:
@geoand I removed the lamdas in favor of `Supplier`s I hope that's ok (I saw that there were other cases with Suppliers as well. 

I suppose the reason for the change is performance?

geoand:
Yeah, at some point in the past we saw there was a tiny impact related to lambdas so we decided to try and limit their use in our code

geoand (commented):
Thanks!

Now all we need is to have the commits squashed 😀

csotiriou:
Ah, yes. done.

geoand (approved):
🎉

geoand:
I force pushed an update that fixes the formatter  complaints

Issue #581, opened 2024-05-14
Explore adding Open WebUI to the DevUI when using Ollama


geoand:
I see [Open WebUI](https://github.com/open-webui/open-webui) being mentioned a lot in the context of Ollama.

We should look into adding it to the WebUI (probably via it's [Docker option](https://docs.openwebui.com/getting-started/#quick-start-with-docker-)).

iocanel:
Do you see it as a new extension or something that would be triggered when OpenAI and Ollama extensions are used ?
Also, would you wire it as a Dev Service or you'd explicitly start it from the Dev UI? In the later case is there an existing extension doing something similar ?

geoand:
> Do you see it as a new extension or something that would be triggered when OpenAI and Ollama extensions are used ?

The latter

> Also, would you wire it as a Dev Service or you'd explicitly start it from the Dev UI

Again the latter :)

phillip-kruger:
Nothing that I know about. But it should be easy enough.

PR #600, opened 2024-05-18
Allow creating and launching an OpenWebUI container via Dev UI 

iocanel:
Resolves: #581 

The pull request introduces the following:

- A new Open WebUI card under langchain4j extension (when using OpenAI or Ollama).
- The Open WebUI card allows starting, stoping or launching Open WebUI.
- OpenAI api key is automatically passed to the container (for OpenAI)
- Ollama base url is automatically passed to the container (for Ollama)
- Options for custom env variables, env variable mapping and port bindings are provided. 
- Added support for automatically signing up

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/OpenWebUIJsonRPCService.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/OpenWebUIJsonRPCService.java
@@ -1,0 +1,148 @@
+package io.quarkiverse.langchain4j.runtime.devui;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import jakarta.enterprise.context.control.ActivateRequestContext;
+
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.testcontainers.DockerClientFactory;
+
+import com.github.dockerjava.api.command.CreateContainerResponse;
+import com.github.dockerjava.api.command.InspectContainerResponse;
+import com.github.dockerjava.api.exception.NotFoundException;
+import com.github.dockerjava.api.model.Bind;
+import com.github.dockerjava.api.model.Container;
+import com.github.dockerjava.api.model.DeviceRequest;
+import com.github.dockerjava.api.model.HostConfig;
+import com.github.dockerjava.api.model.PortBinding;
+import com.github.dockerjava.api.model.Volume;
+
+@ActivateRequestContext
+public class OpenWebUIJsonRPCService {
+
+    public static final Runnable CLOSE_TASK = () -> {
+        OpenWebUIJsonRPCService service = new OpenWebUIJsonRPCService();
+        service.stopOpenWebUI();
+    };
+
+    private static final String CONTAINER_NAME_PREFIX = "quarkus-open-webui-";
+
+    public boolean isOpenWebUIRunning() {
+        var container = inspectOpenWebUIContainer();
+        return container != null && container.getState().getRunning();
+    }
+
+    public String getOpenWebUIUrl() {
+        var container = inspectOpenWebUIContainer();
+        if (container != null) {
+            return container.getNetworkSettings().getPorts().getBindings().values().stream().flatMap(Arrays::stream)
+                    .map(p -> "http://localhost" + ":" + p.getHostPortSpec())
+                    .findFirst()
+                    .orElse(null);
+        }
+        return null;
+    }
+
+    public InspectContainerResponse inspectOpenWebUIContainer() {
+        return DockerClientFactory.lazyClient().listContainersCmd().exec()
+                .stream()
+                .filter(OpenWebUIJsonRPCService::isOpenWebUIContainer)
+                .findFirst()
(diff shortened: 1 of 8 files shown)


iocanel:
There is an issue with the volume handling. It does not appear if you've created the container manually, but will hit users that never used open-webui before and just tried to created from devui. 

Converting to draft till I solve this one.

geoand:
@iocanel is this ready or still in draft?

iocanel:
> @iocanel is this ready or still in draft?
This is ready

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/OpenWebUIJsonRPCService.java
```diff
@@ -0,0 +1,145 @@
...
+            }
+
+            HostConfig hostConfig = new HostConfig()
+                    .withBinds(allBinds)
+                    .withPortBindings(allPortBindings)
+                    .withExtraHosts("host.docker.internal:host-gateway")
+                    .withDeviceRequests(allDeviceRequests);
+
+            CreateContainerResponse container = DockerClientFactory.lazyClient()
+                    .createContainerCmd(image)
```
jmartisk, line 115: I had errors that the image doesn't exist

```
 com.github.dockerjava.api.exception.NotFoundException: Status 404: {"cause":"image not known","message":"no such image: ghcr.io/open-webui/open-webui:main: image not known","response":404}
```

and had to manually run `podman pull ghcr.io/open-webui/open-webui:main` to get it working. Could it be made to pull the image automatically?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/OpenWebUIJsonRPCService.java
```diff
@@ -0,0 +1,145 @@
...
+    public boolean isOpenWebUIRunning() {
+        var container = inspectOpenWebUIContainer();
+        return container != null && container.getState().getRunning();
+    }
+
+    public String getOpenWebUIUrl() {
+        var container = inspectOpenWebUIContainer();
+        if (container != null) {
+            ContainerNetwork network = container.getNetworkSettings().getNetworks().get("bridge");
+            if (network != null) {
```
jmartisk, line 43: I don't understand podman too much, but here I get `network == null` so it can't find the port. The only network I get from `container.getNetworkSettings().getNetworks()` is named `podman`, not `bridge`

jmartisk:
IIUC this just connects to the remote model provider directly by itself, so we can't have it working with our RAG mechanisms etc, right? Or could we make it work by perhaps, exposing the Ollama/OpenAI API by Quarkus as some kind of wrapper that adds our RAG and applies our configuration and having the OpenWebUI call the remote model through that?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/OpenWebUIJsonRPCService.java
iocanel: I'll look into that.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/OpenWebUIJsonRPCService.java
iocanel: I'll try to make it work regardless of docker/podman and fallback to the knwon port.

iocanel:
> IIUC this just connects to the remote model provider directly by itself, so we can't have it working with our RAG mechanisms etc, right? Or could we make it work by perhaps, exposing the Ollama/OpenAI API by Quarkus as some kind of wrapper that adds our RAG and applies our configuration and having the OpenWebUI call the remote model through that?

Your understanding is correct, it just runs an Open WebUI instance that is aligned with the app in terms of model.
Open WebUI does provide in its interface some options for configuring RAG, but I don't know if there is room for integration. It's definetely something we can explore in the future.

geoand:
Should we merge this or are we waiting for some changes?

iocanel:
For anyone trying it with ollama running locally don't forget to set OLLAMA_HOST=0.0.0.0 otherwise ollama won't be visible from the container.

geoand:
I just built this locally but I don't see anything new the DevUI so I guess I am missing something :)

iocanel:
> @iocanel any idea about ^?

The only thing I can think of is that either you sample is not using ollama, openai or its not using a SNAPSHOT. Let me try it again.

iocanel:
After rebasing, I can no longer see anything myself

geoand:
So it's not just me :)

iocanel:
> So it's not just me :)
Yeah, the PR needed alignment with the new module structure. Fixed now.

iocanel:
> So it's not just me :)
Yeah, the PR needed alignment with the new module structure. Fixed now.

geoand:
I just tried it and it woks nicely! The UI could certainly be improved, but I think we should merge now and improve later

geoand:
cc @jmartisk

jmartisk (approved):

PR #605, opened 2024-05-21
Rename csv-chatbot to sql-chatbot

jmartisk:
- also updates `csv.adoc` documentation to use a retrieval augmentor instead of retriever (even though we don't really have any CSV sample using this anymore, because we changed that one to use SQL... [should we re-create one?])


geoand (approved):
Makes sense!

cescoffier (approved):

PR #606, opened 2024-05-21
Restructure the code of the repository

geoand:
The idea is to group modules with similar functionality into directories thus cutting down 
on the number of top level directories and making it easier to navigate the project


geoand:
> but I am all for it before it's too late.

Exactly

jmartisk (approved):

iocanel (approved):
+1000

Issue #611, opened 2024-05-21
Embedding Ollama problem - No default constructor found


duchamk:
When trying to do embedding on ollama like in the code below:
```java
package prv.duchamk.quarkus.test.main;

import dev.langchain4j.model.embedding.EmbeddingModel;
import io.quarkus.runtime.QuarkusApplication;
import jakarta.enterprise.context.control.ActivateRequestContext;
import jakarta.inject.Inject;

@ActivateRequestContext
public class MyQuarkusApplication implements QuarkusApplication {
    
    @Inject EmbeddingModel aiembed;

    @Override
    public int run(String... args) throws Exception {
        // embedding
        var v = aiembed.embed("Jakiś tekst.").content().vector();
        System.out.println("vector: "+v.length);
        return 0;
    }

}
```
I get the answer:
```log
2024-05-21 18:54:24,699 INFO  [io.qua.lan.oll.OllamaRestApi$OllamaLogger] (vert.x-eventloop-thread-2) Request:
- method: POST
- url: http://localhost:11434/api/embeddings
- headers: [Accept: application/json], [Content-Type: application/json], [User-Agent: Quarkus REST Client], [content-length: 64]
- body: {
  "model" : "nomic-embed-text",
  "prompt" : "Jakiś tekst."
}
2024-05-21 18:54:25,794 INFO  [io.qua.lan.oll.OllamaRestApi$OllamaLogger] (vert.x-eventloop-thread-2) Response:
- status code: 200
- headers: [Content-Type: application/json; charset=utf-8], [Date: Tue, 21 May 2024 16:54:25 GMT], [Transfer-Encoding: chunked]
[...]

PR #612, opened 2024-05-21
Added default constructor

duchamk:
Resolves: #611 
Added default constructor to EmbeddingResponse.java for correctly unmarshal response.


geoand:
Thanks for the contribution, but as I explain [here](https://github.com/quarkiverse/quarkus-langchain4j/issues/611#issuecomment-2123880592), this is not correct.

Issue #42, opened 2023-11-20
Allow declaring tools on methods


cescoffier:
Tools declaration provides access to tools for all the methods. However, you may want to control which method can access which tools.

PR #613, opened 2024-05-22
Introduce a way to configure tools per method of an AiService

geoand:
This is done using the new `@ToolBox` annotation

- Closes: #42


geoand:
@cescoffier can you give this a try with the example you had in mind for the `@ToolBox` use case?

cescoffier:
@geoand I will try it tomorrow.

geoand:
🙏

geoand:
I'll fix the conflict if we decide we want this

geoand:
@cescoffier can you have a look at this one?

cescoffier:
@geoand Yes, I will try to use it on the demo we did last week.

geoand:
🙏🏼

cescoffier:
Works! 

```java

    @UserMessage("""
            Extract the policy number from the given claim.
            Only return the policy number.
            
            {text}
            """)
    String extractPolicyNumber(String text);

    @UserMessage("""
            Check if there is an existing claim related to the given claim.
            Return the claim id, null if none.
                        
            The existing claim can be found by identifying the policy number from the given claim.
            
            If no claim is found, return -1.
            
            {text}
            
             """)
    @ToolBox(ClaimService.class)
    Long findRelatedClaims(String text);
```

geoand:
Thanks for checking!

I will check it

geoand:
Done

PR #616, opened 2024-05-22
fix issue with api-key for open-ai to have a default of empty and val…

MoheyEl-DinBadr:
…idate if the default base url is used then the api key must be provided


geoand:
I am pretty sure this does **not**  work properly in the case where multiple named configurations exist.

What is the use case you are trying to fix? It's not clear to me what your end goal is with this change.

MoheyEl-DinBadr:
I have tested it with the default and a named configuration named test and both worked just fine

PR #634, opened 2024-05-28
Update azure-openai to check OIDC TokenCredential

sberyozkin:
This PR is similar to #626. 

If #626 is approved then, as far as this PR is concerned, once the azure-openai and openai-common changes are reviewed, I'd trim it by avoiding adding yet another demo identical to the one added in #626 and instead, try to merge both demos into a single one, showing multiple quarkus-langchain4j models in action and complementing it with annotation based OIDC tenant selection (either Google or Microsoft Entra ID)

PR #646, opened 2024-05-31
Allow dynamic authorization headers for OpenAI + Azure OpenAI

csotiriou:
This allows configuring the Azure OpenAI authorization headers on the fly, by instantiating named beans and implementing an interface.

The concept is that the developer can create a bean like this:

```java
@ApplicationScoped
@ModelName("my-model-name") //you can omit this if you have only one model or if you want to use the default model
public class TestClass implements OpenAiRestApi.AuthProvider {
    @Inject MyTokenProviderService tokenProviderService;

    @Override
        return "Bearer " + tokenProviderService.getToken();
    }
}
```

One can create beans on a per-configuration scenario using the `@Named` annotation. 

### Use case
This will allow using Azure OpenAI integration in complex environments where IAM policies and short-lived Azure tokens are used. In such scenarios, the tokens need to be refreshed periodically using custom security policies, thereby needing such solutions.


model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/QuarkusOpenAiClient.java
```diff
@@ -526,6 +545,11 @@ public Builder azureAdToken(String azureAdToken) {
             return this;
         }
 
+        public Builder configName(String modelName) {
```
geoand, line 548: We should probably stick with either `configName` or `modelName`

model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/QuarkusOpenAiClient.java
```diff
@@ -113,6 +116,21 @@ public void filter(ClientRequestContext requestContext) {
...
+                    OpenAiRestApi.DynamicAuthorizer authorizer;
+                    Instance<OpenAiRestApi.DynamicAuthorizer> beanInstance = builder.modelName == null
+                            ? CDI.current().select(OpenAiRestApi.DynamicAuthorizer.class)
+                            : CDI.current().select(OpenAiRestApi.DynamicAuthorizer.class, NamedLiteral.of(builder.modelName));
+
+                    authorizer = beanInstance
+                            .handlesStream()
+                            .findFirst()
+                            .map(Instance.Handle::get)
+                            .orElse(null);
```
geoand, line 128: I would prefer this be done without streams

sberyozkin:
@csotiriou By the way, is #634 relevant ? It you don't use an of the quarkus extensions related to oidc or jwt then it won't be relevant, otherwise it should be... In #634 I'd like verify that OIDC authorization code flow token can be used, but if the OIDC bearer tokens are already coming in then #634 should take of them

sberyozkin:
FYI, I'm not suggesting in anyway that #634 might be a better solution, as I can imagine these tokens may be coming via different channels altogether, so having a dedicated dynamic authorization support is nice on its own.

geoand:
Let's figure it out

csotiriou:
@sberyozkin @geoand This PR is not relevant to #634 , since it doesn't have something to do with strictly OIDC. Authenticating with Azure endpoints (or Google endpoints, for that matter) does not strictly rely on OIDC for all use cases. It's just one mechanism coming out of the box. 

For example, you may work behind a proxy managing the authentication itself (custom authentication mechanism towards the proxy), or when using IAM on a cloud provider, you may also rely on mounted secrets (changing each X minutes!) inside the pod. A notable example of this (which is not my use case, but it shows how diverse the IAM world can be) is the way AWS may handle IAM inside a pod with their SDK: https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/WebIdentityTokenCredentialsProvider.html - this reads a mounted file from the disk, which contains the secret you need to use, and it's calculated during runtime.

My use case is even more complex, relying on connecting IAM worlds from two cloud providers to actually retrieve a short lived token - falling well outside of the OIDC libraries scope. Depending on the enterprise security guidelines of companies, this is not uncommon.

model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/QuarkusOpenAiClient.java
csotiriou: nice catch. I would prefer sticking with configName since modelName is taken by something else during the setup of the ImageModel.

sberyozkin:
@csotiriou Thanks for the nice explanation, makes total sense to have a generic dynamic mechanism in place to support various dynamic token inputs, beyond the typical standard ones. 

I was thinking a bit if such filters should also work for other providers, like `vertex-gemini`, but it can complicate things with issues like which filter applies to which provider, if the new interface would be in the core module, as it would require a named model support similarly to how it is done for AI services.
 
As far as #646 is concerned, I hope, if that is proven to work and agreed it makes sense, that it will be free for users, i.e, they won't have to write and register custom dynamic filter. I'll CC you there as well to discuss things related to that PR further 

Thanks

csotiriou:
I made a rename there, and added some documentation in the OpenAI section. 

I'm currently contemplating adding more power to the developer by changing the 
```
interface AuthProvider {
    String getAuthorization();
}
```

to
```
interface AuthProvider {
    String getAuthorization(String method, URI url, MultivaluedMap<String, String> headers);
}
```

Haven't decided yet, just wanted to ask. The thought is that in the scope of the named model one may want to differentiate the behaviours and the authentication per deployment or by things passed as headers. The deployment name is not passed in this context but it is inferrable via the request URI.

Not sure about this, yet, though, what do you thing?

geoand:
I am +1 for adding the more "powerfull" version, however I would write it like this:

```java
interface AuthProvider {
    String getAuthorization(Input input);

    interface Input {
       String getMethod();
 
       ...
    }
}
```

This way we can add stuff in the future without breaking the method signature

geoand (approved):
Awesome!

I just added one last comment

model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/OpenAiRestApi.java
```diff
@@ -179,6 +182,32 @@ public boolean test(SseEvent<String> event) {
         }
     }
 
+    interface AuthProvider {
+        String getAuthorization(Input input);
+
+        record Input(
```
geoand, line 188: This should be an interface as it is not something created by users, but instead by the framework

geoand (approved):

Issue #656, opened 2024-06-05
Dev UI: provide some kind of library of prepared system messages


jmartisk:
Perhaps this could be a dropdown menu that contains some useful system messages, and when you select one of them, it will get copied to the 'system message' field.

Some patterns are https://github.com/danielmiessler/fabric/tree/main/patterns but maybe these are too complex and we may need some simpler ones?

geoand:
@langchain4j also mentioned https://smith.langchain.com/hub. I wonder if we could add a search bar that would get results from it

jmartisk:
One problem with [https://smith.langchain.com/hu](https://smith.langchain.com/hub) is that the prompts contain placeholders (parameters) and the UI doesn't have any support for that, the user would have to manually replace the placeholders with something, which isn't exactly nice UX

 https://github.com/danielmiessler/fabric/tree/main/patterns doesn't have this problem

jmartisk:
I have a prototype implementation of the UI part and will send a PR soon. Then we can decide what it will actually show...

PR #658, opened 2024-06-05
Offer sample system messages and user prompts in the Dev UI

jmartisk:
Just a very rough draft for now, to continue the discussion. Added one sample where the LLM summarizes a git diff that I took from this repository. Maybe the samples should be moved to separate file(s) but I'd have to figure out how. Also, I'm not sure if we can limit the size of the text areas, they grow very big after selecting a long system+user message.

- Fixes: https://github.com/quarkiverse/quarkus-langchain4j/issues/656

### DIFF
--- a/core/deployment/src/main/resources/dev-ui/qwc-chat.js
+++ b/core/deployment/src/main/resources/dev-ui/qwc-chat.js
@@ -98,10 +98,22 @@ export class QwcChat extends LitElement {
                                       this._ragEnabled = event.target.checked;
                                       this.render();
                                   }}"/></div>
+            <vaadin-select
+                    label="Select a sample system message"
+                    @change="${(event) => {
+                        this._systemMessage = event.target.value.systemMessage;
+                        this.shadowRoot.querySelector('.userMessageInput').value = event.target.value.userMessage;
+                        this._enableSystemMessageInputField();
+                        this._cementSystemMessage();
+                        this.render();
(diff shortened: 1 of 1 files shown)


iocanel:
Why provide a hard coded list of message and not collect the system messages defined by the application ?

geoand:
What I had in mind when opening the issue was for us to leverage the awesome prompt library available in https://github.com/danielmiessler/fabric/tree/main/patterns

geoand:
@jmartisk what's the status of this one?

jmartisk:
The status is that I would like somebody to chime in and say if this is worth pursuing further this way, or some other way, or not at all :)

jmartisk:
Oh and yeah, around https://github.com/danielmiessler/fabric/tree/main/patterns
The problem is that the prompts contain placeholders. In the UI, it's not very pleasant to work with, as one has to read the prompt and manually replace all placeholders

jmartisk:
Unless we also come up with a way to allow pleasantly defining values for placeholders, it doesn't seem very useful.

There's also the question of whether pre-defined prompts from some kind of library really make sense in this context, while the main point of the UI page is to allow easy testing of RAG and tools of the application.

geoand:
I am not saying that we must have those prompts in this screen, but there should _somewhere_ that we surface those prompts to users as they are super helpful

Issue #637, opened 2024-05-30
Check if we could implement a Cache for LLMs


iocanel:
I've discussed that a little bit with @geoand and @cescoffier and it seems that there might be value in providing the ability to have a cache for LLMs.

It could work by creating and embedding for the prompt and specify threshold for the distance. If the distance is shorter than the specified threshold we could serve the cached response.

Extra points for integrating it with `@Cache`.

geoand:
The only issue I see is when we are caching a response that makes no sense (which can obviously happen due to the non-deterministic nature of LLMs)

andreadimaio:
What do you mean by creating and embedding a prompt? Your idea is to embed the user's queries and, if two queries are similar based on a threshold, return the cached response from the LLM?

cescoffier:
@andreadimaio Yes, exactly.

For each user request, we compute an embedding that we store (with an eviction strategy like a TTL).  It should only be used for a "stateless" query (so the context only contains the system message (if any) and the user message, no other message - as it could leak). 

When a subsequent query occurs and its embedding is close to a stored one (the threshold is configurable, of course), we would use the cached response.

The rationale is that computing an embedding, especially with in-process/local embedding, is cheaper and faster than calling the LLM.

andreadimaio:
Thanks @cescoffier , so if I understand, the cache will only contain the last message for each user? In my mind the cache could contain more messages, but now if I think about this, multiple messages could generate strange behaviors.

geoand:
I am also wondering that when the user wants to try different prompts in order to test how best to instruct the model to behave, how we are going to handle the cache

andreadimaio:
> I am also wondering that when the user wants to try different prompts in order to test how best to instruct the model to behave, how we are going to handle the cache

The `@CacheResult` annotation can help with this, or not? What I mean is, if the annotation is not present, caching will not be enabled for a particular method or class.

andreadimaio:
Let me go back for a moment to what could happen with a cache of multiple messages.
This behavior could be obtained by using LLM as a chatbot.

`Threshold = 1 # The queries must be equals.`

```
UserMessage:
Can you create a poem about dog of 5 lines? 

AIMessage:
In morning light, the faithful dog does prance,
With wagging tail and eyes that always dance.
A friend so true, in joy and sorrow near,
A heart so pure, no trace of doubt or fear.
In every bark, a symphony of chance.

UserMessage: Json version please.
AiMessage: {
[...]

PR #659, opened 2024-06-05
Enable sematic cache with @CacheResult 

andreadimaio:
In this PR, the idea is to implement the semantic cache as discussed in #637.
The idea is to allow the developer to use the _@CacheResult_ annotation to decide which methods need to use the semantic cache. 

The idea is to create different classes following what was done for `ChatMemory` and use `DefaultMemoryIdProvider` to create an isolated cache for each user if possible (I don't know if this is enough to avoid possible data leakage).

An open point is what to do if the `@Tools` annotation is present. 
Shouldn't `@CacheResult` be usable in this case?

**TODO LIST:**
- Improve tests 
- Understand how to implement the cache concept for stream responses.
- Write documentation

There is still work to be done, but I think there is enough code to understand if the direction taken is the right one or if changes need to be made.


geoand:
> An open point is what to do if the @Tools annotation is present.
Shouldn't @CacheResult be usable in this case?

I think it should

> Understand how to implement the cache concept for stream responses.

This is a tricky one... Maybe we should skip it initially

jmartisk:
If a prompt leads to tool executions, I think we shouldn't cache the response. Because if we then served the same response again without actually invoking the tools again, that would be really weird.

geoand:
It's certainly weird either way

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -186,16 +194,31 @@ public void accept(Response<AiMessage> message) {
...
-        Response<AiMessage> response = toolSpecifications == null
-                ? context.chatModel.generate(messages)
-                : context.chatModel.generate(messages, toolSpecifications);
-        log.debug("AI response obtained");
+        Response<AiMessage> response;
+
+        if (cache != null) {
+            log.debug("Attempting to obtain AI response from cache");
+
+            Embedding query = context.embeddingModel.embed(userMessage.text()).content();
```
jmartisk, line 202: IIUC here we should also check that there is no previous conversation (no user message in the memory), because we want to only support caching for the first message, right? Or are you checking this elsewhere already?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/cache/AiCacheStore.java
```diff
@@ -0,0 +1,43 @@
...
+import dev.langchain4j.data.embedding.Embedding;
+import dev.langchain4j.data.message.AiMessage;
+
+/**
+ * Represents a store for the {@link AiCache} state.
+ * Allows for flexibility in terms of where and how cache is stored.
+ */
+public interface AiCacheStore {
+
+    public record CacheRecord(Embedding embedded, AiMessage response, Instant creation) {
```
jmartisk, line 15: I think we should also store the embedding of the system message (if it existed), not just the user message.

Because if the system message is different, then even when the user message is very similar, we shouldn't use the cache for answering.

andreadimaio: Yes, I'm not worried about that in this "first version". I will add SystemMessage in the next commits.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
andreadimaio: The cache is configurable with the property `quarkus.langchain4j.cache.max-size` (default 1). Should this be hidden from developers?

andreadimaio:
I'm also thinking about a cache property that allows the developer to decide which type of embedding to use, something like `quarkus.langchain4j.cache.embedding-model=dev.langchain4j.model.embedding.AllMiniLmL6V2EmbeddingModel` or `quarkus.langchain4j.cache.embedding-model=bam`

jmartisk:
> I'm also thinking about a cache property that allows the developer to decide which type of embedding to use, something like `quarkus.langchain4j.cache.embedding-model=dev.langchain4j.model.embedding.AllMiniLmL6V2EmbeddingModel` or `quarkus.langchain4j.cache.embedding-model=bam`

That should point not at the provider name, but at the name of the configured model, or whatever is the proper term here (I mean like the `c1`, `c2`, `c3`,.... in https://github.com/quarkiverse/quarkus-langchain4j/blob/0.15.1/integration-tests/multiple-providers/src/main/resources/application.properties)

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
jmartisk: Maybe we're not understanding each other - cache max size is the maximum number of prompts stored in it, right?

But here I am talking about the fact that we should only attempt to use the cache when it's for the first prompt in the conversation, so that we  avoid the situation that you described in https://github.com/quarkiverse/quarkus-langchain4j/issues/637#issuecomment-2144429198

iocanel:
> An open point is what to do if the `@Tools` annotation is present. Shouldn't `@CacheResult` be usable in this case?

I think that the whole cache idea is that in general similar question should produce similar result, so caching in some cases makes sense. When tools are involved this is less likely to happen, as we have absolutely no idea what response the tool will return even for the exact same input.

So, I'd say that it should not cache result, unless the user somehow defines that the tool is idempotent (always produces same output for a given input) or cachable itself.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
andreadimaio: Yes, `max-size` is the maximum number of prompts stored, where prompts should be the combination of `SystemMessage + UserMessage`. Each cache as id, if possible the id of the cache is unique (for example if we are in a `RequestScope` context), if not the id will always be the value `default`.

Id with context:
`<ContextId>#<Class>.<Method>`

Id without context:
`default`

However, I don't think we understand each other (but that's my problem :dizzy_face:). 
To avoid returning a wrong cached response like in the example in [#637 (comment)](https://github.com/quarkiverse/quarkus-langchain4j/issues/637#issuecomment-2144429198), is it not enough to have a `max-size` of 1?

jmartisk: > Having said that, I don't think we understand each other (but that's my problem 😵). To avoid returning a wrong cached response like in the example in [#637 (comment)](https://github.com/quarkiverse/quarkus-langchain4j/issues/637#issuecomment-2144429198), is it not enough to have a `max-size` of 1?

A single `Cache` instance contains prompts (system+user message pairs) from MULTIPLE conversations; and from each such conversation, it stores only the initial prompt, do I get that right? Therefore, the `max-size` refers to the number of conversations from which we store the initial prompt? But then that has nothing to do with how many messages have passed in the current conversation no?

andreadimaio: > A single Cache instance contains prompts (system+user message pairs) 

Correct.

> from MULTIPLE conversations;

Depends on the `id` used to get the `Cache`. It's working like what we have today for the `ChatMemory`.
If your application can have a scope, the `id`  of the `Cache` will be composed by the `unique id of the request + class-name + method-name`, this means that each user will have an instance of a `Cache` for each method in the class annotated with `@RegisterAiService`.

If your application doesn't have a scope, the value id `default` will be used, this means that the `Cache` is only one for all `@RegisterAiService` (all the users has the same `Cache`).

> it stores only the initial prompt

The default implementation is a `MessageWindow` with a `max-size` of `1`, which means that only the most recent prompt is stored.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/cache/AiCacheStore.java
andreadimaio: Fixed with this [commit](https://github.com/quarkiverse/quarkus-langchain4j/pull/659/commits/631e9043c13984edeb8f9b7b44e90ed6daf7a9a2#diff-2dd3bec40934ad6d175f6f14dad1af0e11c234cf5fec69739a89460d472ab55b).

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
andreadimaio: My first approach doesn't work and now I understand your point, so I'll work on improving it.
The idea is to have a cache for each method (if annotated) of an `@RegisterAiService` interface that is shared by all users. The number of prompts stored in the cache is equal to the `max-size` configuration value **(default 1)**.

**Example:**

```java
package org.acme.example;

@RegisterAiService
@CacheResult
@SystemMessage("You are a poet")
public interface LLMService {

	UserMessage("Write a poem of 5 lines about {poem}")
	public String fiveLinesPoem(String poem);
	
	UserMessage("Write a poem of 10 lines about {poem}")
	public String tenLinesPoem(String poem);
}
```

In this case, two caches are created:

- `Cache#1` with id: `#org.acme.exampleLLMService.fiveLinesPoem`
- `Cache#2` with id: `#org.acme.exampleLLMService.tenLinesPoem`

Now suppose we have two users, `ADM` and `JM`, who want to create poems.

`ADM` calls 
 -> fiveLinesPoem("dog"); 
 -> No value in the cache;
 -> Calls the LLM;
 -> Stores the response in the `Cache#1` **[MAX-SIZE REACHED]**;
 

`JM` calls 
 -> tenLinesPoem("dog"); 
 -> No value in the cache;
 -> Calls the LLM;
 -> Stores the response in the `Cache#2` **[MAX-SIZE REACHED]**;


`ADM` calls
 -> tenLinesPoem("dog"); 
 -> Response found in the `Cache#2` from the call executed by `JM`;
 -> Return the cached value;
 
`ADM` calls
 -> tenLinesPoem("cat"); 
 -> No value in the cache;
 -> Calls the LLM;
 -> Value not stored in `Cache#1` for **MAX-SIZE REACHED**! 
 
 
The TTL configuration can be used to clean up the cached values. 
What do you think about it? Do you see any potential problems? If it is okay with you, I can start with these changes.

andreadimaio: One point I see here is that the `ADM` user can have access to `JM` responses, not so if we want to make the cache user-bound (figure out how to do this).

andreadimaio: Maybe something like `@MemoryId` could help.

andreadimaio: This idea of implementation has a limitation, in the example above, the cache will be available only during the request context, this means that if you will do more request to the LLM in the same context you can get a cached response, but when the request will end, the cache will be cleared and everything will start from zero.

The same behavior is present when a developer uses an `@ApplicationScope`, but in this case there is something worse which is that they will be stored new cache for each call until  the applicatikon memory is saturate (and of course this is not acceptable).

A solution could be to have another annotation `@CacheId` that is mandatory if you are in the `@ApplicationScope` scenario, in this case the number of cache will be equal to the users.

This doesn't solve the out-of-memory problem, because if the number of users is too high, the problem persists, so the developer must have a fine control of the cache to avoid this problem (this means to have the same behavior of the ChatMemory).

I don't know if I'm making the implementation too complicated and you have something in mind that solves the problem in a much simpler way, in this case your help is welcome

jmartisk: Hmm I thought we were going to make the cache shared between users, not scoped to each user individually. 

Yes, that has potential security implications, but I don't find it very useful if we only consider cached results for each user separately - how often will a single user repeat the same questions again? Not very often, I think.

Scoping the cache to an HTTP request is too limiting too - one HTTP request usually involves just one prompt. We need to keep the cached responses for longer so that they can be reused in later conversations. If you scope it to a HTTP request, you dispose of it immediately after one request.

My understanding was we would have a cache for each AI service method, but its life would be application-scoped and store the first prompt of every conversation that starts by calling that method.

andreadimaio: > My understanding was we would have a cache for each AI service method, but its life would be application-scoped and store the first prompt of every conversation that starts by calling that method.

Ok, so in this case the id of the cache will always be: `#<ClassName>.<MethodName>`
In the example above: `#org.acme.exampleLLMService.fiveLinesPoem` and `#org.acme.exampleLLMService.tenLinesPoem` for all the users.

> How to avoid memory problems, that's an important question - set a reasonable TTL and periodically clean it, or I'm not sure we could use weak references to have the GC clean up stuff over time.

Yes, in this case it might be better to set a default TTL and of course document the fact that the developer can find an out-of-memory problem if the TTL value is too high.

jmartisk: Ok and the default `max-size` will surely be more than one, because we will store prompts from potentially different users. So actually, the `max-size` will probably be the main safeguard against eating too much memory

andreadimaio: Ok, let me go back to the example with the last changes we have in mind:

```java
package org.acme.example;

@RegisterAiService
@CacheResult
@SystemMessage("You are a poet")
public interface LLMService {

	@UserMessage("Write a poem of 5 lines about {poem}")
	public String fiveLinesPoem(String poem);
	
	@UserMessage("Write a poem of 10 lines about {poem}")
	public String tenLinesPoem(String poem);
}
```

**Configuration:**
```properties 
max-size: 2
ttl: 5m
```

`ADM` and `JM`, wants to create poems.

**Two caches are created:**
```
Cache#1:  #org.acme.exampleLLMService.fiveLinesPoem
Cache#2:  #org.acme.exampleLLMService.tenLinesPoem
```

**Steps:**
1) ADM calls
-> fiveLinesPoem("dog");
-> No value in the cache;
-> Calls the LLM;
-> Stores the response in the `Cache#1`

2) JM calls
-> fiveLinesPoem("cat");
-> No value in the cache;
-> Calls the LLM;
-> Stores the response in the `Cache#1` **[MAX-SIZE REACHED]**

3) ADM calls
-> fiveLinesPoem("turtle");
-> No value in the cache;
-> Calls the LLM;
-> The value is NOT cached [max size reached]

4) ADM calls
-> fiveLinesPoem("cat");
-> Return the value in the cache.

The two cached responses will be deleted after 5 minutes.

jmartisk: Yes this makes sense to me. But let's ask @cescoffier to confirm

andreadimaio:
Do you think it would make sense to enable caching for the Streaming API if the method or class has the `@Blocking` annotation?

geoand:
I think we can leave streaming out for a first version and think about it some more

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
andreadimaio: Any suggestion about that? @cescoffier

lordofthejars:
What is the state of this?

andreadimaio:
> What is the state of this?

It's been a while, but the last few messages were about the scope of the cache, specifically whether or not it should be shared between users.

PR #666, opened 2024-06-11
Reuse embeddings

### LINKED ISSUE #663: Way to write out embeddings to a file?
### Discussed in https://github.com/quarkiverse/quarkus-langchain4j/discussions/643

<div type='discussions-op-text'>

<sup>Originally posted by **edeandrea** May 30, 2024</sup>
I'm using the easy-rag extension which loads a pdf file at startup (in dev mode) to an embedding model which is stored in redis.

Is there a way that I could write out the embeddings to a file rather than re-computing the embeddings? I'm using ollama with mistral, so my dev mode startup is about 10-15 minutes, and then 10-15 mins each time I make a change while in dev mode.

If it were to write out the embeddings to a file the first time, then each subsequent time it would only need to load the embeddings into the store, that would be much faster, especially since (in my case) the file isn't changing.</div>

It actually already is there in upstream, at least when using the in memory embedding store. See https://github.com/langchain4j/langchain4j/blob/eba2b1cd500f7b8874d4da5e406f09935febb9b3/langchain4j/src/main/java/dev/langchain4j/store/embedding/inmemory/InMemoryEmbeddingStore.java#L132-L164

I think all we'd need here in the short term would be some stuff in the easy rag extension that did something with these methods.

What I'm thinking is that, for now, we only support it for in memory embeddings. We introduce these properties

```properties
# Whether or not to turn on this feature
# Default is false, preserving existing functionality
quarkus.langchain4j.easy-rag.reuse-embeddings.enabled=false|true

# Path to where to store/read the embeddings file
# Must be set to something if quarkus.langchain4j.easy-rag.reuse-embeddings.enabled==true
quarkus.langchain4j.easy-rag.reuse-embeddings.file=<some file path>
```

If `quarkus.langchain4j.easy-rag.reuse-embeddings.enabled==false` then there won't be any change to existing functionality.

If `quarkus.langchain4j.easy-rag.reuse-embeddings.enabled==true`, then on startup `quarkus.langchain4j.easy-rag.reuse-embeddings.file` will be examined. If it exists, then the embedding store will be loaded from it (rather than querying the `EmbeddingModel`). If it doesn't exist, then the embeddings will be loaded as they are today, and then they will be written out to that file, so that during a restart the `EmbeddingModel` doesn't need to be re-queried.

edeandrea:
Reuse embeddings

Closes #663


rag/easy-rag/runtime/src/main/java/io/quarkiverse/langchain4j/easyrag/runtime/EasyRagRecorder.java
```diff
@@ -35,37 +40,84 @@ public void ingest(EasyRagConfig config, BeanContainer beanContainer) {
...
         LOGGER.info("Ingested " + documents.size() + " files as " + splitDocuments.size() + " documents");
     }
 
-    public Supplier<InMemoryEmbeddingStore> inMemoryEmbeddingStoreSupplier() {
+    private Path getReuseEmbeddingsFilePath(EasyRagConfig config) {
+        return config.reuseEmbeddings().file()
+                .map(String::trim)
+                .map(Paths::get)
+                .orElseThrow(() -> new ConfigValidationException(createConfigProblems("reuse-embeddings.file")));
+    }
```
geoand, line 103: You know I very much dislike this, but in this case my main complaint is that the error message should state that the file should be specified because a certain feature has been turned on. 
I am also wondering if just using some temp file makes sense, it certainly would reduce the friction in using the feature

docs/pom.xml
```diff
@@ -76,6 +76,11 @@
             <artifactId>quarkus-langchain4j-watsonx</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
```
edeandrea, line 79: @geoand I wasn't sure if I needed to do this in order for the easy-rag configuration adocs to get generated?

docs/pom.xml
```diff
@@ -100,6 +105,19 @@
             <artifactId>quarkus-langchain4j-anthropic-deployment</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.langchain4j</groupId>
```
edeandrea, line 109: @geoand I wasn't sure if I needed to do this in order for the easy-rag configuration adocs to get generated?

docs/pom.xml
geoand: Yeah, IIRC you do need it

docs/pom.xml
geoand: +1

rag/easy-rag/runtime/src/main/java/io/quarkiverse/langchain4j/easyrag/runtime/EasyRagRecorder.java
edeandrea: Without using the map stuff, this method would look like

```java
private Path getReuseEmbeddingsFilePath(EasyRagConfig config) {
        if (config.reuseEmbeddings().file().isPresent()) {
            return Paths.get(config.reuseEmbeddings().file().get().trim());
        }

        throw new ConfigValidationException(createConfigProblems("reuse-embeddings.file"));
}
```

I can make it that way if you'd prefer. I think the `map`/`orElseThrow` is much more readable, but I guess thats personal opinion.

geoand: > and the user wouldn't really know where the embeddings file was in case they wanted to examine and/or save it

Do you think they would really need to do this?

edeandrea: > > and the user wouldn't really know where the embeddings file was in case they wanted to examine and/or save it
> 
> Do you think they would really need to do this?

If I were building a demo I'd want to commit the embeddings file into source control. Or maybe in my tests I configure it to use a file, and in that case I'd want to know where the file was so that I could create it.

> > I can make it that way if you'd prefer. I think the map/orElseThrow is much more readable, but I guess thats personal opinion.
>

> My main gripe if we keep the file property optional is that the user does not know why this new property is needed - the message does not help much

I could default it to something (like somewhere in `Files.createTempDirectory`) and that way the user can override if they'd like, or if not, they don't have to care?

geoand: > If I were building a demo I'd want to commit the embeddings file into source control. Or maybe in my tests I configure it to use a file, and in that case I'd want to know where the file was so that I could create it and/or put in somewhere in src/test/resources and commit it

Fair enough. Then maybe it can default to the current working directory?

edeandrea: > Fair enough. Then maybe it can default to the current working directory?

What would the current working directory actually be? I think its value would be different if the app was run in dev mode vs `java -jar` vs in native?

At least with temp dir we could have something thats deterministic, and we could print it out to the console.

geoand: > What would the current working directory actually be? I think its value would be different if the app was run in dev mode vs java -jar vs in native?

It depends on where you run it from, but with `mvn quarkus:dev` and `java -jar target/quarkus-app/quarkus-run.jar` and `./target/whatever-runner` it would be the root of the project which seems like a reasonable choice.

In any case, it's not terribly important and we don't need to spend too much time finding the best default. I trust you'll use whatever you find best by working with the feature.

edeandrea: just pushed new changes

geoand (approved):
LGTM

rag/easy-rag/runtime/src/main/java/io/quarkiverse/langchain4j/easyrag/runtime/EasyRagRecorder.java
```diff
@@ -35,37 +38,77 @@ public void ingest(EasyRagConfig config, BeanContainer beanContainer) {
...
+
         EmbeddingStore<TextSegment> embeddingStore = beanContainer.beanInstance(EmbeddingStore.class);
         EmbeddingModel embeddingModel = beanContainer.beanInstance(EmbeddingModel.class);
 
+        if (config.reuseEmbeddings().enabled() && (embeddingStore instanceof InMemoryEmbeddingStore<TextSegment>)) {
+            Path embeddingsFile = Path.of(config.reuseEmbeddings().file()).toAbsolutePath();
+
+            // If the embeddings file already exists it would have been ingested
+            // when the InMemoryEmbeddingStore bean was created
+            // See the inMemoryEmbeddingStoreSupplier method
```
jmartisk, line 50: One minor problem is that if you don't call anything on the `embeddingStore` object here, it won't get initialized, and thus the ingestion (from the file) will run lazily (when the store is actually accessed for the first time).  Could we add at least a dummy `embeddingStore.toString()` call or something like this to make sure that the file gets ingested eagerly at startup, for consistency?

jmartisk:
Added just one minor comment, otherwise, this is great!

rag/easy-rag/runtime/src/main/java/io/quarkiverse/langchain4j/easyrag/runtime/EasyRagRecorder.java
edeandrea: Couldn't you mark the `EmbeddingStore` bean as `@Startup` so that it gets initialized eagerly?

jmartisk: Maybe changing its type from `ApplicationScoped` to `Singleton` (in the `EasyRagProcessor` class) would help? Not sure.
But no we can't add annotations directly on it, it's not from this codebase

jmartisk: Maybe @geoand knows a better trick, but this is how I've been doing it :)

geoand: That's how I do it as well

edeandrea: Ok I've updated it and added some comments in the code.

jmartisk (approved):

Issue #667, opened 2024-06-12
Exception when using a named datasource for pgvector


laurentperez:
Hello

I'm getting a 

`2024-06-12 10:14:13.380 CEST+0200 (executor-thread-2 #137) (pid:1225493) io.quarkiverse.langchain4j.runtime.devui.ChatJsonRPCService:154 WARN java.util.concurrent.CompletionException: jakarta.enterprise.inject.CreationException: Error creating synthetic bean [NcsXENBovnGlSJZ1aaFjoohrtKM]: java.lang.IllegalArgumentException: A synthetic injection point was not declared for required type [interface io.agroal.api.AgroalDataSource and qualifiers: [@jakarta.enterprise.inject.Default()] {} `

My database is configured as

```
quarkus.langchain4j.pgvector.dimension=1536
quarkus.datasource.hd.db-kind=postgresql
quarkus.datasource.hd.username=postgres
quarkus.datasource.hd.password=postgres
quarkus.datasource.hd.jdbc.url=jdbc:postgresql://localhost:5433/hd
quarkus.langchain4j.pgvector.datasource=hd
```

Are named datasources supported ? please note that if I make an explicit mistake and make a typo in "hd" (as in "hdfoo") it does not boot at all. "hd" boots, but this exception happens when I attempt the first call to the vector DS

thanks

geoand:
cc @jmartisk

PR #668, opened 2024-06-12
Avoid trying to retrieve the default ds if a named one should be used

jmartisk:
Fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/667


embedding-stores/pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/runtime/PgVectorEmbeddingStoreRecorder.java
```diff
@@ -19,10 +18,13 @@ public class PgVectorEmbeddingStoreRecorder {
     public Function<SyntheticCreationalContext<PgVectorEmbeddingStore>, PgVectorEmbeddingStore> embeddingStoreFunction(
             PgVectorEmbeddingStoreConfig config, String datasourceName) {
         return context -> {
-            AgroalDataSource dataSource = Optional.ofNullable(datasourceName)
-                    .map(DataSourceLiteral::new)
-                    .map(dl -> context.getInjectedReference(AgroalDataSource.class, dl))
-                    .orElse(context.getInjectedReference(AgroalDataSource.class, new Default.Literal()));
```
jmartisk, line 25: This was a sneaky one... The problem here was that `context.getInjectedReference(AgroalDataSource.class, new Default.Literal())` is not wrapped a lambda expression, therefore it gets evaluated always...

geoand: Ouch...

geoand: One more reason to hate lambdas  @edeandrea

geoand (approved):
Thanks!

geoand:
Any chance we can get a test?

jmartisk:
Hmm, the tests in pgvector module are using `@QuarkusTest`. We will probably need to rewrite them to use `quarkus-junit5-internal` because they can't be mixed together.

jmartisk:
I'll try that

geoand:
🙏

embedding-stores/pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/runtime/PgVectorEmbeddingStoreRecorder.java
edeandrea: Thanks for the zing @geoand :) I'll still keep my stance that, generally speaking, functional code is still easier to read and figure out what's going on than imperative code. 

That being said, in this particular case, one has to understand the subtle differences between `Optional.orElse` and `Optional.orElseGet`.

geoand:
Indeed they don't AFAIR

jmartisk:
We can work around it by temporarily extending `EmbeddingStoreIT` instead of `EmbeddingStoreWithFilteringIT`, then we won't have any parametrized tests, but we would lose test coverage of metadata filtering...

geoand:
Up to you

geoand (approved):

PR #678, opened 2024-06-14
Update the email me a poem sample

geoand:
The following changes have been made:

* Use quarkus-mailpit
   - This is done to make the setup simpler
* Introduce the LGTM observability Dev Service
   - This is done so users don't have to start anything manually
   - For this to work, users need to use -Dobservability
* Use gpt-4o
   - This is done because GTP-3.x can sometimes result in multiple tool invocations
   - Improve the prompt to ensure that no markers are added
* Update README accordingly

### DIFF
--- a/embedding-stores/chroma/deployment/src/test/java/io/quarkiverse/langchain4j/chroma/deployment/ChromaEmbeddingStoreCDITest.java
+++ b/embedding-stores/chroma/deployment/src/test/java/io/quarkiverse/langchain4j/chroma/deployment/ChromaEmbeddingStoreCDITest.java
@@ -5,6 +5,7 @@ import jakarta.inject.Inject;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
 import dev.langchain4j.model.embedding.AllMiniLmL6V2QuantizedEmbeddingModel;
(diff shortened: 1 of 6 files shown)


jmartisk:
IMO the mail server component is quite important for demoing (I've shown this at several conferences now)...
"sending an email" by just printing something in the log? :( not happy about that TBH

geoand:
What I really don't like is that unless I actually do a prerequisite step, the sample doesn't work - I really want demoers to have not to do anything

jmartisk:
Could we then switch to https://github.com/quarkiverse/quarkus-mailpit? I haven't tried that, but it claims it can run an SMTP server as a dev service

geoand:
Oh, that's nice! 

Let me try it and see how it goes

maxandersen (commented):
Good changes Overall but should we make the telemetry optional as it's quite heavy to fetch and it's optional for showing the main funktionality of tools calling. ?

geoand:
We can do that

geoand:
@jmartisk the Mailpit extension is amazing, I'm adding it

maxandersen:
> Could we then switch to https://github.com/quarkiverse/quarkus-mailpit? I haven't tried that, but it claims it can run an SMTP server as a dev service, and the emails can be viewed in the Dev UI, that actually sounds cool

+1 for mailpit. I'm actually wondering if we shouldn't simply make the devservice that comes by default with mailer. But that's a separate issue.

geoand:
Yeah, mailpit could be a very nice default

geoand:
PR updated

geoand:
@jmartisk I temporarily disabled `ChromaEmbeddingStoreCDITest`

jmartisk:
Yeah I'm looking at it but I'm puzzled, because it just started failing independently of any changes in our codebase (and the Chroma image version didn't get updated either)

PR #683, opened 2024-06-16
Allow ollama devservice container to create an image with any configured model

Malandril:
Allows ollama devservice container to create an image with any configured model.

This renames model tags replacing the `:` by `_` so that the created image has a correctly named tag.


Malandril:
Running in tests in pipelines is easier with the devservices docker version as i don't have much to change compared to "classic" test containers pipelines.

But if is to be removed then, i guess i need to make an alternative setup.

geoand:
> But if is to be removed then, i guess i need to make an alternative setup.

It will at some point because the developer experience is much worse than with the new way

PR #684, opened 2024-06-16
Allow ollama devservices to run with docker gpus

Malandril:
This adds a configuration to the ollama devservices to allow the started container to request gpus.


maxandersen:
Nice. Which yeah platforms and container runtimes supports this?

geoand:
Thanks for the contribution.

This module is actually on the way out because we now have non-docker based [Ollama](https://docs.quarkiverse.io/quarkus-langchain4j/dev/ollama.html) support.

PR #689, opened 2024-06-18
Dev UI: Allow RAG and tools together with streaming

jmartisk:
Rag>
rag

Tools>
tools


jmartisk:
There's a slight problem here that streaming is always executed on the event loop, so it blows up if some of the tools are blocking. Need to figure out how to properly solve this.

jmartisk:
> There's a slight problem here that streaming is always executed on the event loop, so it blows up if some of the tools are blocking. Need to figure out how to properly solve this.

ok this should be fixed, I'm not sure if there's a better solution but I just added an assumption that tools and retrieval augmentors are blocking and should execute on a worker thread

geoand (approved):

jmartisk:
Merging it, next I'd like to support images and do some more refactoring, the code of the page is getting really ugly

PR #694, opened 2024-06-20
Putting Generic Authentication flow to the Core

csotiriou:
- Moving the authentication interface from Azure OpenAI introduced in #646 to the Quarkus Langchain Core so that it can be reused across other model providers
- Making Azure OpenAI use this.

Implementation of request mentioned in #626

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/auth/ModelAuthProvider.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/auth/ModelAuthProvider.java
@@ -1,0 +1,42 @@
+package io.quarkiverse.langchain4j.runtime.auth;
+
+import java.net.URI;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import jakarta.enterprise.inject.Instance;
+import jakarta.enterprise.inject.spi.CDI;
+
+import io.quarkiverse.langchain4j.ModelName;
+
+public interface ModelAuthProvider {
+    String getAuthorization(Input input);
+
+    interface Input {
+        String method();
+
+        URI uri();
+
+        Map<String, List<Object>> headers();
+    }
+
+    static Optional<ModelAuthProvider> resolve(String modelName) {
+        Instance<ModelAuthProvider> beanInstance = modelName == null
+                ? CDI.current().select(ModelAuthProvider.class)
+                : CDI.current().select(ModelAuthProvider.class, ModelName.Literal.of(modelName));
+
+        //get the first one without causing a bean1 resolution exception
+        ModelAuthProvider authorizer = null;
+        for (var handle : beanInstance.handles()) {
+            authorizer = handle.get();
+            break;
+        }
+        return Optional.ofNullable(authorizer);
+    }
+
+    static Optional<ModelAuthProvider> resolve() {
+        return resolve(null);
+    }
+
+}
--- a/docs/modules/ROOT/pages/openai.adoc
+++ b/docs/modules/ROOT/pages/openai.adoc
@@ -183,7 +183,7 @@ public class RequestFilter implements ResteasyReactiveClientRequestFilter {
 ----
 
 ==== Using `AuthProvider`
-One can implement the `AuthProvider` interface and provide the implementation of the `getAuthorization` method.
+One can implement the `ModelAuthProvider` interface and provide the implementation of the `getAuthorization` method.
 
 This is useful when you need to provide different authorization headers for different OpenAI models. The `@Named` annotation can be used to specify the model name in this scenario.
 
@@ -191,12 +191,13 @@ This is useful when you need to provide different authorization headers for diff
 ----
 import io.quarkiverse.langchain4j.ModelName;
 import io.quarkiverse.langchain4j.openai.OpenAiRestApi;
+import io.quarkiverse.langchain4j.runtime.auth.ModelAuthProvider;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 
 @ApplicationScoped
 @ModelName("my-model-name") //you can omit this if you have only one model or if you want to use the default model
-public class TestClass implements OpenAiRestApi.AuthProvider {
+public class TestClass implements ModelAuthProvider {
     @Inject MyTokenProviderService tokenProviderService;
 
     @Override
--- a/model-providers/openai/openai-common/deployment/src/main/java/io/quarkiverse/langchain4j/openai/deployment/OpenAiCommonProcessor.java
+++ b/model-providers/openai/openai-common/deployment/src/main/java/io/quarkiverse/langchain4j/openai/deployment/OpenAiCommonProcessor.java
@@ -15,8 +15,6 @@ import java.util.zip.ZipEntry;
 
 import com.knuddels.jtokkit.Encodings;
 
-import io.quarkiverse.langchain4j.openai.OpenAiRestApi;
-import io.quarkus.arc.deployment.UnremovableBeanBuildItem;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.IndexDependencyBuildItem;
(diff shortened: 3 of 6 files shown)


csotiriou:
@geoand @sberyozkin, I am putting this as a draft here in reference to #626 , is this what you were talking about? I am going to update the documentation once we agree that this is what's needed.

I opted to keep the usage of the `ModelAuthProvider` inside the Azure OpenAI since I imagine that each provider will have their own authentication implementation using this interface. 

I welcome every proposal / change to this logic.

geoand:
The formatter is complaining:

```
Failed to execute goal net.revelc.code:impsort-maven-plugin:1.9.0:check (check-imports) on project quarkus-langchain4j-openai-common: Imports are not sorted in /home/runner/work/quarkus-langchain4j/quarkus-langchain4j/model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/OpenAiRestApi.java 
```

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/auth/ModelAuthProvider.java
```diff
@@ -0,0 +1,44 @@
...
+
+    interface Input {
+        String method();
+
+        URI uri();
+
+        Map<String, List<Object>> headers();
+    }
+
+    class Resolver {
```
geoand, line 24: Actually, we don't even need this class, we can just add the `resolve()` methods as static methods to `ModelAuthProvider`

csotiriou:
@geoand @sberyozkin thanks,

I just corrected the documentation, since atm OpenAI is the only module that is using it, I suppose we don't need to move it elsewhere. If we want to move it elsewhere, I wonder where this might be (a new section, perhaps?)

sberyozkin:
Thanks @csotiriou 

Having model provider sections like it is done for OpenAI is good IMHO, perhaps, once I have it supported for one more model provider, Vertex Gemini, it would make sense to have a higher level short section on `ModelAuthProvider` linking to model providers supporting it... Does it sound reasonable ? If you prefer, please add it now, or I can do it later...

csotiriou:
I agree with this approach, let's keep things in the documentation simple for now.

geoand:
Hey folks,

sorry for the delay, I was travelling last week and a ton of stuff had piled up!

I will review shortly

docs/modules/ROOT/pages/openai.adoc
```diff
@@ -183,20 +183,20 @@ public class RequestFilter implements ResteasyReactiveClientRequestFilter {
...
+One can implement the `ModelAuthProvider` interface and provide the implementation of the `getAuthorization` method.
 
 This is useful when you need to provide different authorization headers for different OpenAI models. The `@Named` annotation can be used to specify the model name in this scenario.
 
 [source,java]
 ----
 import io.quarkiverse.langchain4j.ModelName;
 import io.quarkiverse.langchain4j.openai.OpenAiRestApi;
-import jakarta.enterprise.context.ApplicationScoped;
+import io.quarkiverse.langchain4j.runtime.auth.ModelAuthProvider;import jakarta.enterprise.context.ApplicationScoped;
```
geoand, line 194:
```suggestion
import io.quarkiverse.langchain4j.runtime.auth.ModelAuthProvider;
import jakarta.enterprise.context.ApplicationScoped;
```

model-providers/openai/openai-common/deployment/src/main/java/io/quarkiverse/langchain4j/openai/deployment/OpenAiCommonProcessor.java
```diff
@@ -29,11 +27,6 @@ void indexDependencies(BuildProducer<IndexDependencyBuildItem> producer) {
         producer.produce(new IndexDependencyBuildItem("dev.ai4j", "openai4j"));
     }
 
-    @BuildStep
-    UnremovableBeanBuildItem unremovableBeans() {
-        return UnremovableBeanBuildItem.beanTypes(OpenAiRestApi.AuthProvider.class);
-    }
-
```
geoand, line 36: Has this moved somewhere else?

geoand (commented):
Thanks a lot! I added a small comment

model-providers/openai/openai-common/deployment/src/main/java/io/quarkiverse/langchain4j/openai/deployment/OpenAiCommonProcessor.java
csotiriou: Nice catch, I had forgotten to commit the changed file!

I hope I put it in the right place.

geoand: Looks great!

geoand:
Thanks a lot for this!

@sberyozkin do you want to test it locally to make sure if fits your needs?

sberyozkin:
Hi @geoand @csotiriou , sure, give me a few days please,  I'll rework vertex gemini PR, cheers

model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/OpenAiRestApi.java
```diff
@@ -182,35 +183,25 @@ public boolean test(SseEvent<String> event) {
...
-            MultivaluedMap<String, Object> headers();
-        }
-    }
-
     class OpenAIRestAPIFilter implements ResteasyReactiveClientRequestFilter {
-        AuthProvider authorizer;
+        ModelAuthProvider authorizer;
 
-        public OpenAIRestAPIFilter(AuthProvider authorizer) {
+        public OpenAIRestAPIFilter(ModelAuthProvider authorizer) {
```
sberyozkin, line 189: IMHO it should be aligned with how it is done for Vertex AI Gemini `TokenFilter`, to have `ExecutorService` injected as well and then call a custom provider from this `ExecutorService` as providers may want to use some blocking calls etc. Using `ManagedExecutor` is best as it retains the original `RequestScope` which lets users inject Quarkus API beans which require it, as proposed in #708.

geoand: Good point!

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/auth/ModelAuthProvider.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+import jakarta.enterprise.inject.Instance;
+import jakarta.enterprise.inject.spi.CDI;
+
+import io.quarkiverse.langchain4j.ModelName;
+
+public interface ModelAuthProvider {
+    String getAuthorization(Input input);
```
sberyozkin, line 14: Hi @csotiriou @geoand, only a couple of questions/suggestions for you to consider.

I was thinking, can we expect situations going forward where local models will require a secure token access, and where a token is passed, for example, as a property ?

I guess one can say that if `Input` is null then do not return a `Bearer ` prefix in the response... This is related to the next question...

The other question, so `ModelAuthProvider` is expected to return a complete HTTP `Authorization` header value, including the scheme, `Bearer <token>`, instead of only `<token>`. In #708, moving to `ModelAuthProvider` has an impact of having to update the test filter to include `Bearer` - I don't think it is a breaking change though, since the current `AuthProvider` is not public.

I wonder, should we expect providers returns the token only, and then filters, depending on a context (access to the remote or local model) will either add `Bearer ` or not.

IMHO it might be a bit better, and a little bit simpler for custom auth providers too...

geoand: I don't think we need to worry about that now, it's very early days to be able to know exactly how systems using auth will evolve.
For now, I think the proposal in this PR is fine.

sberyozkin: @geoand Sure, Ok then, this PR works well for #708 so it is good to go from my perspective, thanks

geoand (approved):

PR #695, opened 2024-06-20
add stop sequence config for openAI chat config model

sebastienblanc:
As specified [here](https://platform.openai.com/docs/api-reference/chat/create#chat-create-stop) in a chat completion request we should be able to pass a `stop` parameter containing a string or list of strings.
It's implemented in upstream `langchain4j` , this PR just add it to the config and the recorder.

### DIFF
--- a/model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/OpenAiRecorder.java
+++ b/model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/OpenAiRecorder.java
@@ -104,7 +104,8 @@ public class OpenAiRecorder {
                     .topP(chatModelConfig.topP())
                     .presencePenalty(chatModelConfig.presencePenalty())
                     .frequencyPenalty(chatModelConfig.frequencyPenalty())
-                    .responseFormat(chatModelConfig.responseFormat().orElse(null));
+                    .responseFormat(chatModelConfig.responseFormat().orElse(null))
+                    .stop(chatModelConfig.stop().orElse(null));
 
             openAiConfig.organizationId().ifPresent(builder::organizationId);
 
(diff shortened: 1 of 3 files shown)


model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/OpenAiRecorder.java
```diff
@@ -59,7 +59,8 @@ public Supplier<ChatLanguageModel> chatModel(LangChain4jOpenAiConfig runtimeConf
                     .topP(chatModelConfig.topP())
                     .presencePenalty(chatModelConfig.presencePenalty())
                     .frequencyPenalty(chatModelConfig.frequencyPenalty())
-                    .responseFormat(chatModelConfig.responseFormat().orElse(null));
+                    .responseFormat(chatModelConfig.responseFormat().orElse(null))
+                    .stop(chatModelConfig.stop().orElse(null));
```
geoand, line 63: This should also be done for `streamingChatModel()` too, right?

sebastienblanc: Indeed ! Let me add it (And it's even in streaming mode where this stop sequence will make the most sense )

geoand:
It would be great if we could have some kind of test for this

sebastienblanc:
> It would be great if we could have some kind of test for this

I agree but honestly how can we test this without hitting a real openAI endpoint ?

sebastienblanc:
> Can't we use Wiremock and configure it to respond with what OpenAI responds with and then just validate that we sent what we expected to send?

I have been looking a bit around but tbh I have not a clear idea on how to write this test. You say you just want to test the request is well formed ?

geoand:
My idea is to validate that the request is what you expect it to be and have the response be whatever OpenAI gives (which you would hardcode by just copying the response from a real OpenAI request)

geoand (approved):

Issue #660, opened 2024-06-08
Azure endpoint pattern domain name


duchamk:
Azure provides Open AI not only in the *.openai.azure.com domain but also in *.azure-api.net.
Is it possible to change the pattern?
From
```java
static final String AZURE_ENDPOINT_URL_PATTERN = "https://%s.openai.azure.com/openai/deployments/%s";
```
to e.g.
```java
static final String AZURE_ENDPOINT_URL_PATTERN = "https://%s/openai/deployments/%s";
```

geoand:
Thanks for reporting.

Is there any benefit from using one or the other?

duchamk:
The domain is used to separate data processed by public AI. This is used by commercial companies and people who do not want their data processed by public artificial intelligence (possible leak). The issue of security of corporate or private data.

geoand:
Thanks.

cc @csotiriou

csotiriou:
I find no reason why this would not change, I don't have access to such domains, but it seems that @duchamk has access. 

I would propose `https://%s.openai.azure.com/openai/deployments/%s` to change to `https://%s.%s/openai/deployments/%s`, where the second argument would be the domain name, which would be configurable, and would be default be `openai.azure.com`, what do you think, @geoand ?

geoand:
Sounds reasonable to me.

@duchamk would you like to contribute this change?

csotiriou:
@geoand  @duchamk I just found that quarkus-langchain4j has the `endpoint` configuration, which allows the entire model endpoint to be set, and be customized to every different model. 

Perhaps @duchamk can use this property? And if that's the case, do we still need to make the change in the configuration (honest question, I would personally be open to all cases)

duchamk:
Yes, there is indeed such a configuration  variable ```quarkus.langchain4j.azure-openai.endpoint```, but it has a serious drawback. Easy RAG does not work and switching between models has to be done from scratch.

I like @csotiriou's idea with the domain, if possible add the ``quarkus.langchain4j.azure-openai.domain-name``` config variable, then you will be able to use the full potential of the extension.

What do you think about it?

PR #698, opened 2024-06-22
Add domain config for Azure AI

duchamk:
- Resolves: #660


geoand:
Thanks a lot!

@csotiriou do you want to review?

csotiriou (approved):

csotiriou:
@geoand @duchamk I approved this by mistake, sorry about that. I did find an issue. 

Since this property is overridable, it should follow the same pattern as with `endPointFor()`, `resourceNameFor()`, and the `@WithDefault` would should only apply to the parent property, not the one that overrides the parent property. 

Can we change that?

geoand:
No problem!

Feel free to do a follow-upi

geoand:
@csotiriou ^

Issue #679, opened 2024-06-14
Allow the outputFormatInstructions to be included or not in the prompt


andreadimaio:
Today, when an AIService method returns an object, the schema is included in the prompt. It might be useful to create a property to give the developer the ability to choose whether this must be included or not, because in some cases (when the developer uses LLM-specific tags) the place where the schema is added (at the end of the prompt) can cause unexpected behavior.

Another improvement might be to reserve a tag for the schema and give the developer the ability to place it anywhere in the prompt.

geoand:
Interesting. 

Do you have specific examples where the current approach might fail?

andreadimaio:
Let's say I'm using `watsonx` with the `llama3` model.

This is my prompt:

```
<|begin_of_text|><|start_header_id|>system<|end_header_id|>

You are poet<|eot_id|><|start_header_id|>user<|end_header_id|>

Generate a poem of 5 lines about {topic}<|eot_id|><|start_header_id|>assistant<|end_header_id|>
```

And I want to put the poem in a specific java object

```java
@RegisterAiService
public interface LLMService {

	@SystemMessage(...)
	@UserMessage(...)
	public Poem poem(String topic)
}
```

In this particular case, the schema of the `Poem` object is appended at the end of the prompt, but the correct place should be in the system or user "tags".

Now, in this particular example, this behavior may not cause a problem, but having the ability to enable/disable the schema, or better yet, give the developer the ability to choose where to put it, would be great.

geoand:
I see, thanks. 

I think it makes sense.

andreadimaio:
What I have in mind is to have a property `quarkus.langchain4j.response-schema` with values of `true` **(default)** or `false`, which will be inherited by all LLM providers. This configuration allows the developer to decide whether or not to create the schema for each object returned in the AIService methods.

With the new property, there is also a new reserved placeholder called `{response_schema}` that can be used by the developer to choose where to place the schema in the prompt (could be in SystemMessage, UserMessage, or both).

If `quarkus.langchain4j.response-schema` is set to `true` and there is no `{response_schema}`, the schema will be placed at the end of the prompt (same behavior as today), otherwise if there is at least one `{response_schema}`, it will only be inserted where specified.

If `quarkus.langchain4j.response-schema` is set to `false`, no schemas will be generated and included in prompts.
If `quarkus.langchain4j.response-schema` is set to `false` and a `{response_schema}` exists, an exception will be thrown.

Make sense?

[...]

PR #699, opened 2024-06-23
Allow the use of the {response_schema} in the prompt

andreadimaio:
As mentioned in #679, the idea of this PR is to increase the flexibility of prompt creation with the use of the `{response_schema}` placeholder.

This new placeholder allows the developer to choose where to place the schema in the prompt.

```java
@RegisterAiService
public interface LLMExtractor {


    @UserMessage("""
    <|begin_of_text|><|start_header_id|>system<|end_header_id|>

     # IDENTITY and PURPOSE

     You are a entity extractor. You take input and output a JSON.
     Your goal is to:

    * Extract the entities from the text: ["firstname", "lastname", "email", "city"];
    * Respond only with the entities extracted from the text;
    * Capitalize the first letter of the firstname, lastname, and city entities;
    * {response_schema};

    <|eot_id|>
    <|start_header_id|>user<|end_header_id|>

     # INPUT

     INPUT: {text}

    <|eot_id|><|start_header_id|>assistant<|end_header_id|>
     """)
    public User user(String text);
}

```

If there is no `{response_schema}` placeholder in the `@SystemMessage` or `@UserMessage`, the current default behavior is executed (the schema is added at the end of the prompt).

If necessary, the developer can disable the schema generation using the `quarkus.langchain4j.response-schema` property to `false`.

The new placeholder works with all possible AIService interactions except `@StructuredPrompt`.

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/config/LangChain4jBuildConfig.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/config/LangChain4jBuildConfig.java
@@ -36,6 +36,13 @@ public interface LangChain4jBuildConfig {
      */
     DevServicesConfig devservices();
 
+    /**
+     * Configuration property to enable or disable the use of the {response schema} placeholder in
+     * the @SystemMessage/@UserMessage.
+     */
+    @WithDefault("true")
+    boolean responseSchema();
+
     interface BaseConfig {
         /**
          * Chat model
@@ -61,8 +68,9 @@ public interface LangChain4jBuildConfig {
     @ConfigGroup
     interface DevServicesConfig {
         /**
-         * If DevServices has been explicitly enabled or disabled. DevServices is generally enabled
-         * by default, unless there is an existing configuration present.
+         * If DevServices has been explicitly enabled or disabled. DevServices is generally enabled by default, unless there is
+         * an
+         * existing configuration present.
          * <p>
          * When DevServices is enabled Quarkus will attempt to automatically serve a model if there are any matching ones.
          */
--- a/docs/modules/ROOT/pages/ai-services.adoc
+++ b/docs/modules/ROOT/pages/ai-services.adoc
@@ -113,6 +113,14 @@ TriagedReview triage(String review);
 
 In this instance, Quarkus automatically creates an instance of `TriagedReview` from the LLM's JSON response.
 
+To enhance the flexibility of prompt creation, the `\{response_schema\}` placeholder can be used within the prompt. This placeholder is dynamically replaced with the defined schema of the method's return object.
+
+[IMPORTANT]
+====
+By default, if the placeholder `\{response_schema\}` is not present in `@SystemMessage` or `@UserMessage`, it will be added to the end of the prompt.
+To avoid the generation of the schema, the property `quarkus.langchain4j.response-schema` can be set to `false`.
+====
+
 === Receiving User Message as a Parameter
 
 For situations requiring the user message to be passed as a parameter, you can use the `@UserMessage` annotation on a parameter. Exercise caution with this feature, especially when the AI has access to _tools_:
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -54,12 +54,15 @@ import dev.langchain4j.exception.IllegalConfigurationException;
 import dev.langchain4j.service.Moderate;
 import io.quarkiverse.langchain4j.ModelName;
 import io.quarkiverse.langchain4j.ToolBox;
+import io.quarkiverse.langchain4j.deployment.config.LangChain4jBuildConfig;
 import io.quarkiverse.langchain4j.deployment.items.SelectedChatModelProviderBuildItem;
 import io.quarkiverse.langchain4j.runtime.AiServicesRecorder;
 import io.quarkiverse.langchain4j.runtime.NamedConfigUtil;
 import io.quarkiverse.langchain4j.runtime.RequestScopeStateDefaultMemoryIdProvider;
+import io.quarkiverse.langchain4j.runtime.ResponseSchemaUtil;
 import io.quarkiverse.langchain4j.runtime.aiservice.AiServiceClassCreateInfo;
 import io.quarkiverse.langchain4j.runtime.aiservice.AiServiceMethodCreateInfo;
+import io.quarkiverse.langchain4j.runtime.aiservice.AiServiceMethodCreateInfo.ResponseSchemaInfo;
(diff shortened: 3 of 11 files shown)


docs/modules/ROOT/pages/ai-services.adoc
```diff
@@ -113,6 +113,14 @@ TriagedReview triage(String review);
 
 In this instance, Quarkus automatically creates an instance of `TriagedReview` from the LLM's JSON response.
 
+To enhance the flexibility of prompt creation, the `{response_schema}` placeholder can be used within the prompt. This placeholder is dynamically replaced with the defined schema of the method's return object.
```
jmartisk, line 116:
```suggestion
To enhance the flexibility of prompt creation, the `+{response_schema}+` placeholder can be used within the prompt. This placeholder is dynamically replaced with the defined schema of the method's return object.
```

This is to avoid Antora trying to resolve this as an Antora attribute named `response_schema` - I can see this in the build log:
```
[WARNING] [io.quarkiverse.antora.deployment.NativeImageBuildRunner] /antora/docs/modules/ROOT/pages/ai-services.adoc:0: skipping reference to missing attribute: response_schema
```

docs/modules/ROOT/pages/ai-services.adoc
```diff
@@ -113,6 +113,14 @@ TriagedReview triage(String review);
...
+
+[IMPORTANT]
+====
+By default, if the placeholder `{response_schema}` is not present in `@SystemMessage` or `@UserMessage`, it will be added to the end of the prompt.
```
jmartisk, line 120:
```suggestion
By default, if the placeholder `+{response_schema}+` is not present in `@SystemMessage` or `@UserMessage`, it will be added to the end of the prompt.
```

dtto as in https://github.com/quarkiverse/quarkus-langchain4j/pull/699/files#r1650503191

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/config/LangChain4jBuildConfig.java
```diff
@@ -36,6 +36,12 @@ public interface LangChain4jBuildConfig {
      */
     DevServicesConfig devservices();
 
+    /**
+     * Whether the response schema can be used in the prompt
```
jmartisk, line 40: This description isn't very clear to me - should it say 
```
the {response_schema} placeholder
```
instead of just
```
response schema
```
?

jmartisk: maybe repeat here what you said in https://github.com/quarkiverse/quarkus-langchain4j/issues/679#issuecomment-2168722583 :)

jmartisk:
Btw are we sure this isn't something that would better fit into the upstream langchain4j project? For example, what if that project decides to introduce the same feature at some point in the future, how would we handle the duplication...

docs/modules/ROOT/pages/ai-services.adoc
andreadimaio: It seems that one solution to avoid the warning is to use `\{response_schema\}`.

docs/modules/ROOT/pages/ai-services.adoc
andreadimaio: It seems that one solution to avoid the warning is to use `\{response_schema\}`.

andreadimaio:
> Btw are we sure this isn't something that would better fit into the upstream langchain4j project? For example, what if that project decides to introduce the same feature at some point in the future, how would we handle the duplication...

Yes, maybe this is something that can be useful to have in the langchain4j repo. I sped up the implementation in quarkus because otherwise I was blocked from moving forward with creating prompts in projects.

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/config/LangChain4jBuildConfig.java
andreadimaio: Ok, I can think of way to improve this comment.

andreadimaio:
Of course I can use a workaround like returning a `String` and using the `ObjectMapper` to have the output bean. 
But this PR should avoid all that. If you think it is better to implement this feature directly in langchain4j we can close this PR :)

jmartisk:
Yeah, I would say that we should generally attempt to implement things in langchain4j wherever appropriate... to avoid the potential problems in the future where it decides to do a similar feature, and we have to reconcile the duplication, etc. That is, unless the maintainer is against having this - what do you think about this feature @langchain4j ?

jmartisk:
Even if it's implemented upstream, it should still be possible to have a Quarkus-side configuration property wired to it (I'm not against that)

andreadimaio:
> I am not fully sure about this exact solution (need more time to check), but it definitely makes sense to make it configurable in LC4j. BTW, there are a couple of related PRs: [langchain4j/langchain4j#1176](https://github.com/langchain4j/langchain4j/pull/1176) [langchain4j/langchain4j#1126](https://github.com/langchain4j/langchain4j/pull/1126)

The two PRs don't seem to address what I have in mind. In my case, I need to decide where to place the schema of the output object in the prompt. Today, the default behavior is to add the schema to the end of the `UserMessage`, and this is not always correct.

The linked PRs talk about how to customize the parser of the schema, which is something very interesting in any case.

andreadimaio:
@jmartisk, does it make sense at this point to make the changes you requested?

jmartisk:
I guess some parts of this PR will still be relevant even if we do it in upstream (we will probably want to have the configuration property), but I'd perhaps suggest doing the upstream part first, and then, after/along with upgrading the langchain4j dependency later, appropriately update this PR with what will still be relevant?!

geoand:
I personally think this makes a lot of sense. If we can have the same in upstream LangChain4j, if not, we can live with a little drift

andreadimaio:
Not wanting to leave things half done, I made the changes @jmartisk requested. I don't know how you want to handle this PR, if you want to merge it, close it or wait for something in the langchain4j repo. From my point of view, there is nothing else to do with this PR at the moment :)

andreadimaio:
> out of curiosity, which LLM provider do you use that you need to manually format the prompt with `<|begin_of_text|>` and other special tokens?

I'm using `llama-3-8b-instruct` in `watsonx`. In this case there is no `chat` API to do the templating automatically. This means that putting the object schema at the end of the prompt is an error (in this particular case). And in any case, even if there was a `chat` API, I think it's always better to have control over where the schema is placed.

geoand:
I like it!

@jmartisk WDYT?

jmartisk:
IMO if it's going to langchain4j itself, I'd refrain from merging this.

But maybe we will need our separate implementation anyway because our implementation of AI services is too diverged (@geoand  has more knowledge of this), so I'll let you decide

geoand:
>  our implementation of AI services is too diverged

This will likely continue into the future, as there is no real other way of incorporating build time processing

geoand (approved):

jmartisk:
Ok then let's merge it

PR #704, opened 2024-06-25
Switch category of extensions to AI

cescoffier:
Waiting for "code.quarkus.io" to have an AI category (Ping @ia3andy).

### DIFF
--- a/core/runtime/src/main/resources/META-INF/quarkus-extension.yaml
+++ b/core/runtime/src/main/resources/META-INF/quarkus-extension.yaml
@@ -7,7 +7,7 @@ metadata:
     - langchain4j
   guide: "https://docs.quarkiverse.io/quarkus-langchain4j/dev/index.html"
   categories:
-    - "miscellaneous"
+    - "ai"
   status: "experimental"
   codestart:
     name: langchain4j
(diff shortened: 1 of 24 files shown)


geoand (approved):

ia3andy:
@cescoffier I think this is the place to add it: https://github.com/quarkusio/quarkus/blob/main/devtools/bom-descriptor-json/src/main/resources/catalog-overrides.json

core/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
     - langchain4j
   guide: "https://docs.quarkiverse.io/quarkus-langchain4j/dev/index.html"
   categories:
-    - "miscellaneous"
+    - "AI"
```
maxandersen, line 10: should be lower case "ai", its an ID not a name/text.

maxandersen (changes requested):
use "ai" rather than "AI" then all matches other category ids

maxandersen:
related https://github.com/quarkusio/quarkus/pull/41425

cescoffier:
Switch to "ai" instead of "AI"

geoand:
Is there something blocking this?

cescoffier:
Nothing blocking.

PR #711, opened 2024-06-28
Update pgvector-store.adoc - remove reference to Redis

reconway:
Line 7 - replace reference to "Redis" with reference to "PostgreSQL pgVector" as this page deals with Postgres and pgVector.


geoand (approved):
Thanks!

PR #713, opened 2024-07-02
Add a warning about missing -parameters flag for javac

geoand:
Without this warning, users can end up with only an
error complaining about a duplicate null key

This was prompted by https://stackoverflow.com/q/78695273/2504224, but it's 
far from the first time I've seen the same issue.


geoand:
@jmartisk if you don't have time to look at the Neo4j failures, let me know please so I can disable them

jmartisk:
@geoand Hmmm supercool, another problematic embedding store that just stops working? I'll look into it asap, but feel free to disable for now

jmartisk (approved):

jmartisk:
Ok I can't even reproduce it locally, this will be tricky

jmartisk:
I think it's something with neo4j, because with `neo4j:5.20.0` image it works, with the recently released `5.21.0` it fails

PR #718, opened 2024-07-04
Prevent false positive of javac -parameters

geoand:
This could happen before when there were no template params


core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -947,7 +947,7 @@ private List<TemplateParameterInfo> gatherTemplateParamInfo(List<MethodParameter
             templateParams.add(new TemplateParameterInfo(0, "it"));
         }
 
-        if (templateParams.stream().map(TemplateParameterInfo::name).allMatch(Objects::isNull)) {
+        if (!templateParams.isEmpty() && templateParams.stream().map(TemplateParameterInfo::name).allMatch(Objects::isNull)) {
```
geoand, line 950: The Javadoc of `allMatch` says it returns `true`  when the stream is empty

Issue #700, opened 2024-06-24
Tracker for upgrading to LangChain4j 0.32


jmartisk:
Checklist:

- [x] Switch the parallel execution of in-process embedding models to use the quarkus worker pool (see https://github.com/langchain4j/langchain4j-embeddings/pull/31) - IMPLEMENTED IN https://github.com/jmartisk/quarkus-langchain4j/commit/74fd385163d84e042be392da25afc74b1ab15cf3
- [x] Breaking changes in Mistral models due to https://github.com/langchain4j/langchain4j/pull/1226 
 - [x] Use new observability interfaces (see https://github.com/langchain4j/langchain4j/pull/1229)

jmartisk:
@geoand let's try tracking the necessary changes like this, WDYT?
Once we become aware of more, we can add them here

geoand:
Sounds like a good idea!

jmartisk:
I've also pushed a branch https://github.com/quarkiverse/quarkus-langchain4j/tree/langchain4j-0.32 that integrates the mandatory changes for the upgrade. I've been doing this for SmallRye GraphQL for quite a while, maintaining an upgrade branch over time, rather than implementing everything at once the new version is released... Let's see how it works... We can always back off

geoand:
Sure, let's see how it goes

geoand:
@jmartisk I assume you haven't done anything for MistralAI, correct? 

Asking so I don't duplicate work

jmartisk:
Correct, I haven't looked at that

geoand:
👍🏼

geoand:
I have force pushed an update

PR #721, opened 2024-07-04
Update to LangChain4j 0.32.0

geoand:
Also includes:

* Use the Quarkus worker pool for in-process embedding models
* Add span for each call to a ChatLanguageModel
* Add metrics for each call to a ChatLanguageModel

- Closes: #700


geoand:
@jmartisk would you take a look at the Redis failures please?

jmartisk:
Yeah, looking.

jmartisk:
Ok so the problem is this new line:

https://github.com/langchain4j/langchain4j/blob/0.32.0/langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreIT.java#L97

which adds a metadata field or type `UUID`, and I was only expecting numbers and strings when 'guessing' whether the corresponding field in Redis should be type `NUMERIC` or `TEXT`, and my guessing mechanism guesses that this is a number. But when you place a UUID (as a string) into such field and create an embedding for it, Redis won't  return it from the index search operation, because it's invalid (unfortunately, it won't throw any error...).

So I think we can, in this case adjust the guessing mechanism to treat the UUID as a String instead of a Number. 

But, FYI @langchain4j , maybe we should put in place some rules about what values you can put in the metadata. You probably shouldn't be allowed to put any arbitrary objects in there, because it wouldn't be clear how to store them in each underlying embedding store type, and how to perform `equals`, `greaterThan`, etc. operations on it. UUID is probably ok because it can be stored as a string. But I'm thinking maybe we should restrict Metadata values to be String or Number only.

jmartisk:
@langchain4j Ah, sorry I just noticed you already have some checks in place, and explicitly allow UUID along with numbers and strings. That's all good then :)

jmartisk:
Ok great, now we have some failures in Milvus instead

geoand:
lovely...

geoand:
I squashed the the last two (test fix) commits into the first one so we don't have broken intermediate states

jmartisk (approved):

Issue #726, opened 2024-07-08
Using PGVector storage with Plain Hibernate Postgres breaks the Embeddings implementation


csotiriou:
it seems that when having an application which needs to make use of both Hibernate and the PGVector Store, there are conflicts which prohibit the implementation of the PGVector store to work.

Example dependencies:

```
 <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-hibernate-orm-panache</artifactId>
        </dependency>

        <!-- JDBC driver dependencies -->
        <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-jdbc-postgresql</artifactId>
        </dependency>
        <dependency>
            <groupId>io.quarkiverse.langchain4j</groupId>
            <artifactId>quarkus-langchain4j-pgvector</artifactId>
            <version>${quarkus.langchain.version}</version>
        </dependency>
```

When trying to make any operation which requires the "vector" object to be enabled in the database, quarkus throws an error:
```
Caused by: java.lang.RuntimeException: Failed to execute 'CREATE TABLE IF NOT EXISTS embeddings (embedding_id UUID PRIMARY KEY, embedding vector(1536), text TEXT NULL, metadata JSON NULL )'
        at dev.langchain4j.store.embedding.pgvector.PgVectorEmbeddingStore.initTable(PgVectorEmbeddingStore.java:173)
        at dev.langchain4j.store.embedding.pgvector.PgVectorEmbeddingStore.<init>(PgVectorEmbeddingStore.java:81)
        at io.quarkiverse.langchain4j.pgvector.PgVectorEmbeddingStore.<init>(PgVectorEmbeddingStore.java:45)
        at io.quarkiverse.langchain4j.pgvector.runtime.PgVectorEmbeddingStoreRecorder.lambda$embeddingStoreFunction$1(PgVectorEmbeddingStoreRecorder.java:31)
        at io.quarkiverse.langchain4j.pgvector.PgVectorEmbeddingStore_NcsXENBovnGlSJZ1aaFjoohrtKM_Synthetic_Bean.createSynthetic(Unknown Source)
        ... 42 more
Caused by: org.postgresql.util.PSQLException: ERROR: type "vector" does not exist
```

It seems that when including plain dependencies of Hibernate with the PGVector Extension, the `PgVectorAgroalPoolInterceptor->onConnectionCreate()` method fails to run, therefore some things are not able to run.

PR #727, opened 2024-07-08
Fix race condition with use of PgVectorAgroalPoolInterceptor

geoand:
Prior to this, the application could open a connection that did not use the interceptor and then try and
reuse that connection against a PgVector database
without having the interceptor (as Agroal
would not take it into account because the connection was already open)

Fixes: #726


geoand:
Once this is in, I am going to do a release of `0.16.1` as this could easily be a blocker

jmartisk:
> @jmartisk I think this interceptor should not be applied to the prod run, WDYT?

By that you mean connections for any other datasource than the one configured for embedding storage? If so, then yeah.

embedding-stores/pgvector/deployment/src/main/java/io/quarkiverse/langchain4j/pgvector/deployment/PgVectorEmbeddingStoreProcessor.java
```diff
@@ -68,6 +72,15 @@ public void createBean(
...
                 .done());
 
+        beanProducer.produce(SyntheticBeanBuildItem
+                .configure(PG_VECTOR_AGROAL_POOL_INTERCEPTOR)
+                .types(ClassType.create(AGROAL_POOL_INTERCEPTOR))
+                .setRuntimeInit()
+                .unremovable()
+                .supplier(recorder.pgVectorAgroalPoolInterceptor())
+                .qualifiers(datasourceQualifier)
+                .done());
```
jmartisk, line 82: The default scope is application-scoped?  Just making sure we don't unnecessarily create a lot of instances of the interceptor for no reason. I'm trying to find out what the default scope is, but I can't find it.

geoand: That's a very good question. I think it is, but let's add it just to be on the safe side.

geoand:
> By that you mean connections for any other datasource than the one configured for embedding storage

I mean that from what I understand, this interceptor creates a type, which I think should not never be done by the framework in production (but it currently is).

embedding-stores/pgvector/deployment/src/main/java/io/quarkiverse/langchain4j/pgvector/deployment/PgVectorEmbeddingStoreProcessor.java
geoand: Actually, I think `@Dependent` is the default...

jmartisk:
> > By that you mean connections for any other datasource than the one configured for embedding storage
> 
> I mean that from what I understand, this interceptor creates a type, which I think should not never be done by the framework in production (but it currently is).

But how else should it work? The application would have to register a similar interceptor by itself... Which would be annoying

geoand:
> But how else should it work? The application would have to register a similar interceptor by itself... Which would be annoying

My point is that a production database would already be setup properly - in the same way you don't expect the application to create your relational schema

jmartisk:
> > But how else should it work? The application would have to register a similar interceptor by itself... Which would be annoying
> 
> My point is that a production database would already be setup properly - in the same way you don't expect the application to create your relational schema

For the database, sure, but this is not about the database, this is about registering an extension of the JDBC driver (client)

geoand:
Ah okay, I misunderstood it's functionality then - my bad!

geoand:
So this PR is ready then

jmartisk:
Ah, so IIUC it seems that actually the `CREATE EXTENSION` command is indeed server-side...
But the interceptor also calls
```
PGvector.addVectorType(connection);
```
which is the client-side setup.
So perhaps you're right and we could skip the `CREATE EXTENSION` part of the interceptor in production mode.

jmartisk:
The question is if it's really a big burden to have it executed, just to simplify things

geoand:
Thanks for checking further. 

Let's leave it as is for now, and I'll open a new issue to keep the enhancement in mind for later.
I am going to assume we will need it because I believe lots of applications won't have the karma to do `CREATE EXTENSION`

jmartisk (approved):

Issue #731, opened 2024-07-08
Upgrading 0.15.1 to 0.16.1 causes issues with micrometer


edeandrea:
I have a pull request in the Quarkus superheroes that upgrades Quarkus Langchain4j from 0.15.1 to 0.16.1 and I'm getting all kinds of startup errors related to Micrometer.

See https://github.com/quarkusio/quarkus-super-heroes/actions/runs/9840344352/job/27164373444?pr=1069

```
[INFO] Running io.quarkus.sample.superheroes.narration.rest.AzureOpenAiNarrationResourceIT
Executing "/usr/lib/jvm/temurin-17-jdk-amd64/bin/java -Dquarkus.http.port=0 -Dquarkus.http.ssl-port=8444 -Dtest.url=http://0.0.0.0:0/ -Dquarkus.log.file.path=/home/runner/work/quarkus-super-heroes/quarkus-super-heroes/rest-narration/target/quarkus.log -Dquarkus.log.file.enable=true -Dquarkus.log.category."io.quarkus".level=INFO -Dquarkus.profile=azure-openai -Dquarkus.langchain4j.azure-openai.enable-integration=true -Dquarkus.wiremock.devservices.port=42405 -Dquarkus.langchain4j.azure-openai.endpoint=http://localhost:${quarkus.wiremock.devservices.port}/v1/ -Dquarkus.langchain4j.azure-openai.max-retries=2 -Dquarkus.langchain4j.azure-openai.log-requests=true -Dquarkus.test.arg-line= -Dquarkus.langchain4j.azure-openai.dalle3.enable-integration=true -Dquarkus.langchain4j.azure-openai.dalle3.timeout=3s -Dquarkus.langchain4j.azure-openai.dalle3.endpoint=http://localhost:${quarkus.wiremock.devservices.port}/v1/ -Dquarkus.langchain4j.azure-openai.timeout=3s -Dquarkus.configuration.build-time-mismatch-at-runtime=fail -Dquarkus.langchain4j.azure-openai.dalle3.max-retries=2 -Dquarkus.langchain4j.azure-openai.dalle3.log-responses=true -Dquarkus.http.host=0.0.0.0 -Dquarkus.langchain4j.azure-openai.dalle3.log-requests=true -Dquarkus.langchain4j.azure-openai.log-responses=true -jar /home/runner/work/quarkus-super-heroes/quarkus-super-heroes/rest-narration/target/quarkus-app/quarkus-run.jar"

[...]

Issue #57, opened 2023-11-22
I tried whatever and it did not work


geoand:
This is how I reprodiced the problem:
'
BLKA VLA VLA

Issue #50, opened 2023-11-22
Unable to use sentence-transformers/all-mpnet-base-v2 with HuggingFace


cescoffier:
This embedding model is the default when using Python.

See: https://api.python.langchain.com/en/latest/embeddings/langchain.embeddings.huggingface.HuggingFaceEmbeddings.html

The exception is the following:

```
jakarta.ws.rs.ProcessingException: The timeout period of 10000ms has been exceeded while executing POST /pipeline/feature-extraction/sentence-transformers/all-mpnet-base-v2 for server api-inference.huggingface.co:443
        at org.jboss.resteasy.reactive.client.impl.InvocationBuilderImpl.unwrap(InvocationBuilderImpl.java:223)
        at org.jboss.resteasy.reactive.client.impl.InvocationBuilderImpl.method(InvocationBuilderImpl.java:344)
        at io.quarkiverse.langchain4j.huggingface.HuggingFaceRestApi$$QuarkusRestClientInterface.embed(Unknown Source)
        at io.quarkiverse.langchain4j.huggingface.QuarkusHuggingFaceClientFactory$QuarkusHuggingFaceClient.embed(QuarkusHuggingFaceClientFactory.java:84)
        at io.quarkiverse.langchain4j.huggingface.QuarkusHuggingFaceEmbeddingModel.embedTexts(QuarkusHuggingFaceEmbeddingModel.java:70)
        at io.quarkiverse.langchain4j.huggingface.QuarkusHuggingFaceEmbeddingModel.embedAll(QuarkusHuggingFaceEmbeddingModel.java:63)
        at dev.langchain4j.model.embedding.EmbeddingModel_39f15f757608db0afd626cecc7d217137da45ff3_Synthetic_ClientProxy.embedAll(Unknown Source)
        at dev.langchain4j.store.embedding.EmbeddingStoreIngestor.ingest(EmbeddingStoreIngestor.java:62)
        at io.quarkiverse.langchain4j.sample.openshift.ingestion.IngestionJob.run(IngestionJob.java:63)
        at picocli.CommandLine.executeUserObject(CommandLine.java:2026)
        at picocli.CommandLine.access$1500(CommandLine.java:148)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2461)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2453)
[...]

PR #732, opened 2024-07-08
Fix issue with Azure OpenAI and metrics usage

geoand:
Essentially the ChatLanguageModel was being
built too early thus messing up the construction
order of the dependencies

- Fixes: #731

### DIFF
--- a/integration-tests/multiple-providers/src/test/java/org/acme/example/multiple/MultipleTokenCountEstimatorProvidersTest.java
+++ b/integration-tests/multiple-providers/src/test/java/org/acme/example/multiple/MultipleTokenCountEstimatorProvidersTest.java
@@ -1,7 +1,6 @@
 package org.acme.example.multiple;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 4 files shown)


geoand:
This fails because of the use cce63037 - I'll first need to fix that

andreadimaio:
I moved the `builder.build()` outside of the `Supplier<>` to have the same instance for the ChatLanguageModel and the TokenCountEstimator. Is this also done for the `watsonx` and `bam` modules.

andreadimaio:
And also remove the following lines from the [MultipleTokenCountEstimatorProvidersTest](https://github.com/quarkiverse/quarkus-langchain4j/blob/main/integration-tests/multiple-providers/src/test/java/org/acme/example/multiple/MultipleTokenCountEstimatorProvidersTest.java) class: `#50`, `#57`, `#64`

geoand:
Yup, I'm going to revert those changes

geoand:
I tried a few things, but in the end the obvious fix is the one that makes most sense

PR #734, opened 2024-07-08
Fix Moderation doc

andreadimaio:
After PR #372, the `@RegisterAiService(moderationModelSupplier = RegisterAiService.BeanModerationModelSupplier.class)`
is no longer necessary.


geoand (approved):
Thanks!

PR #756, opened 2024-07-18
Add `prompt-joiner` property for watsonx provider

andreadimaio:
Before these changes, all `ChatMessage`(s) were joined using the delimiter `\n` or `\n\n`.
The idea is to give the choice to the developer by using the `quarkus.langchain4j.watsonx.chat-model.prompt-joiner` property.

This way the developer can choose the preferred way of concatenating messages to ensure the prompt is structured correctly.


geoand (approved):
Thanks!

PR #770, opened 2024-07-23
Replace Singleton with ApplicationScoped

emmanuelbernard:
This aligns with the rest of the examples in the documentation page. There might be a good reason as I'm not an expert but it felt odd to see singleton here. If there is no reason, better align tot he more accepted `ApplicationScoped` and be consistent with the rest of the examples in the doc page.


geoand (approved):
Good catch, let's keep things consistent

PR #780, opened 2024-07-25
Improve JSON sanitize

### LINKED ISSUE #776: Improve JSON sanitize
The [sanitize method](https://github.com/quarkiverse/quarkus-langchain4j/blob/b52c8414bc61640347b47d93283d8f084f2148bc/core/runtime/src/main/java/io/quarkiverse/langchain4j/QuarkusJsonCodecFactory.java#L34-L35) doesn't work if the LLM returns as ouput:

Example 1:
```
```\n{ ... }\n```
```
Example 2:
```
```json\n{ ... }\n
```
Example 3:
```
```\n{ ... }\n
```

The second and third examples can occur when a developer uses the `chat-model.stop-sequences=}\n` property.
One solution might be to use the regex: **```(json)***

andreadimaio:
This PR resolves #776

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/QuarkusJsonCodecFactory.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/QuarkusJsonCodecFactory.java
@@ -5,6 +5,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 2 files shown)


geoand (approved):
Thanks!

andreadimaio:
Oops,, I have imported `org.apache.commons.lang3.stream.Streams` instead of `java.util.stream.Stream;`

geoand:
Nice catch :)

andreadimaio:
I'm thinking about this new regex, and it works, but maybe it's not the best we can do. What is the end result we want to achieve? We want to extract only the JSON, maybe the best regex version is `{.*}|\[.*\]`, this will allow to remove all the useless characters outside the object/array json. @geoand

geoand:
> We want to extract only the JSON

Correct!

> maybe the best regex version is {.*}|\[.*\], this will allow to remove all the useless characters outside the object/array json

Have you tried it?

andreadimaio:
The current tests are all green, I'm trying to add new tests to see if anything is missing. But first impressions are that the new regex looks much better.

geoand:
Agreed :)

geoand:
Yeah, I don't think we need to worry about that for now

andreadimaio:
Ok, just the time to rerun all the tests, and I will open a PR.

geoand:
🎉

Issue #786, opened 2024-07-26
EasyRag when multiple providers are present erroring when running tests


edeandrea:
I'm working on a workshop that uses EasyRag. It's a pretty simple project. The project has multiple provider extensions present (openai & ollama).

https://github.com/rh-rad-ai-roadshow/parasol-insurance/tree/private-docs-rag-solution/app

My `application.properties` has (among other things)

```properties
quarkus.langchain4j.parasol-chat.chat-model.provider=ollama # There are multiple chat models
quarkus.langchain4j.embedding-model.provider=ollama # There is only a single embedding model
```

When I run in dev mode things seem to be fine, but when I run `./mvnw verify` it blows up:

```
[ERROR] org.parasol.resources.ClaimResourceTests.getAllNoneFound -- Time elapsed: 0.002 s <<< ERROR!
java.lang.RuntimeException: 
java.lang.RuntimeException: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
        [error]: Build step io.quarkiverse.langchain4j.deployment.BeansProcessor#handleProviders threw an exception: io.quarkus.runtime.configuration.ConfigurationException: A EmbeddingModel bean was requested, but since there are multiple available providers, the 'quarkus.langchain4j.embedding-model.provider' needs to be set to one of the available options (openai,ollama).
        at io.quarkiverse.langchain4j.deployment.BeansProcessor.selectEmbeddingModelProvider(BeansProcessor.java:333)
        at io.quarkiverse.langchain4j.deployment.BeansProcessor.handleProviders(BeansProcessor.java:175)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at io.quarkus.deployment.ExtensionLoader$3.execute(ExtensionLoader.java:849)
        at io.quarkus.builder.BuildContext.run(BuildContext.java:256)
        at org.jboss.threads.ContextHandler$1.runWith(ContextHandler.java:18)
        at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2516)
        at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2495)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1521)
[...]

PR #792, opened 2024-07-30
Fix embedding model handling with Easy RAG

jmartisk:
Fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/786

### DIFF
--- a/embedding-stores/chroma/deployment/src/main/java/io/quarkiverse/langchain4j/chroma/deployment/ChromaProcessor.java
+++ b/embedding-stores/chroma/deployment/src/main/java/io/quarkiverse/langchain4j/chroma/deployment/ChromaProcessor.java
@@ -43,7 +43,7 @@ class ChromaProcessor {
                         ParameterizedType.create(EmbeddingStore.class, ClassType.create(TextSegment.class)))
                 .defaultBean()
                 .setRuntimeInit()
-                .defaultBean()
+                .unremovable()
                 .scope(ApplicationScoped.class)
                 .supplier(recorder.chromaStoreSupplier(config))
                 .done());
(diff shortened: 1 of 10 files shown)


geoand (approved):

embedding-stores/chroma/deployment/src/main/java/io/quarkiverse/langchain4j/chroma/deployment/ChromaProcessor.java
```diff
@@ -43,7 +43,7 @@ public void createBean(
                         ParameterizedType.create(EmbeddingStore.class, ClassType.create(TextSegment.class)))
                 .defaultBean()
                 .setRuntimeInit()
-                .defaultBean()
+                .unremovable()
```
geoand, line 46: Is this really correct?

jmartisk: Now that the Dev UI page doesn't have a injection point for it anymore, I think we need to make it unremovable if we want to make it work for dev mode purposes. One test that uses Chroma was failing without this change

geoand: At the very least, I would assume that we need the same thing everywhere

jmartisk: Yeah right, I'll add it to others.. I think it's useful to have the embedding store working for Dev UI purposes even if the app doesn't have any injection point for it.

geoand: Agreed

edeandrea (approved):
LGTM! Thanks for the quick turnaround!

Issue #765, opened 2024-07-22
Set programmatically when to load documents in EasyRag


lordofthejars:
Currently, when using EasyRAG, documents are read at startup time. This is great, and it should be this by default, as this makes EasyRAG Easy. Still, it would also be nice to create a new property that disables this behavior, and as developers, we can select the exact time when the documents are loaded programmatically.

For me, this is an addition that can be useful:

* In the case of extensive documents, you choose when to import them.
* In case you are using external storage and want to dynamically load documents (because they are added to the storage from another service)
* You want to test the app before and after RAG

geoand:
cc @jmartisk

jmartisk:
So you want to be able to turn off the automatic ingestion (which already is possible), and add some sort of programmatic trigger for it? What do you envision the "trigger" to look like? Some Quarkus-specific public static method that invokes the logic that we have in `EasyRagRecorder`?

I'm thinking that maybe such a case would be better with "regular" RAG (programmatically create and call a recursive `DocumentSplitter` and an `EmbeddingStoreIngestor` etc) instead of introducing some new (kind of dirty) API, but maybe I'm not understanding your idea properly.

lordofthejars:
Hi, yes, the only thing I am thinking about is setting when to trigger the EasyRag ingestion. Something similar to what we can do with Flyway, to automatically run it or decide when programmatically. 

I think it is not necessary to have a static method. It can be a class that can be injected.

jmartisk:
Ok I'll try to come up with something and let you review it

PR #794, opened 2024-07-30
Introduce the MANUAL ingestion strategy for Easy RAG

jmartisk:
Fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/765

@lordofthejars

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-langchain4j-easy-rag.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-langchain4j-easy-rag.adoc
@@ -117,7 +117,7 @@ a| [[quarkus-langchain4j-easy-rag_quarkus-langchain4j-easy-rag-ingestion-strateg
 
 [.description]
 --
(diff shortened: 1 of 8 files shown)


lordofthejars:
Yes, that's what I was thinking, I love it. Just one minimal thing, and I know that naming things is complex, in fact, I don't know a better name, but `EasyRagManualIngestionTrigger` seems a Spring Boot class rather than a Quarkus class 🤣 because of the length of the class name. Maybe`EasyRagIngestionTrigger`, or `EasyRagManualIngestion`.

 If you `remove` the trigger part in the name, then maybe the method should be called `ingest` instead of `trigger`.

As I said, I am fine with the current names, just offering a point of view on the names, that's all.

jmartisk:
Ok I've renamed it to `EasyRagManualIngestion` :laughing:

lordofthejars:
> Ok I've renamed it to `EasyRagManualIngestion` 😆

:) Great job, as I said it was just a personal opinion about the name, but thanks for doing it.

PR #803, opened 2024-08-01
Improve JSON sanitize

andreadimaio:
I have thought more about how the regex should be written to sanitize the JSON string that comes from the LLM and my final consideration is that the regex to use is `(?s)\\{.*\\}|\\[.*\\]`. This will allow to extract the contents of the object/array JSON without any other characters.


geoand (approved):
Thanks a lot!

PR #812, opened 2024-08-10
 Replaced `ProfileManager` with `ConfigUtils` to check active profiles.

fmatar:


embedding-stores/pgvector/runtime/src/main/java/io/quarkiverse/langchain4j/pgvector/PgVectorAgroalPoolInterceptor.java
```diff
@@ -16,8 +16,9 @@ public class PgVectorAgroalPoolInterceptor implements AgroalPoolInterceptor {
 
     @Override
     public void onConnectionCreate(Connection connection) {
+        ConfigUtils.isProfileActive("dev");
```
jmartisk, line 19: this seems unnecessary

fmatar: Realized it was a statement I didn't delete (Ask Copilot). Already removed it

jmartisk (approved):

jmartisk:
Thanks!

PR #814, opened 2024-08-12
Fix #813 to add proxy support for openai

### LINKED ISSUE #813: Add proxy support for openAI
There is a `Proxy` which can be used in chat model builder but we have not relevant config items.

zhfeng:


jmartisk (approved):
Worked nicely for me, thanks!

PR #834, opened 2024-08-28
Properly handle resource missing case for user and system messages

### LINKED ISSUE #831: @SystemMessage(fromResource = "bad-path.txt") fails with `@SystemMessage's template parameter cannot be empty`
If an AI service is condfigured with a bad path it fails with: `@SystemMessage's template parameter cannot be empty`

```
java.lang.RuntimeException: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
	[error]: Build step io.quarkiverse.langchain4j.deployment.AiServicesProcessor#handleAiServices threw an exception: dev.langchain4j.exception.IllegalConfigurationException: @SystemMessage's template parameter cannot be empty. Offending method is 'io.quarkiverse.langchain4j.openai.test.ServiceWithResource#findTalks'
	at dev.langchain4j.exception.IllegalConfigurationException.illegalConfiguration(IllegalConfigurationException.java:12)
```

It should fail with something like "resource not found". 

Relevant code is: https://github.com/quarkiverse/quarkus-langchain4j/blob/323966c4d3f155bda7d57e9d824f512517a01e5c/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java#L1127

Probably there should be an else case on reading the `is` like: 

```java
                if (is != null) {
                    return new String(is.readAllBytes());
                }else {
                    throw new FileNotFoundException("Resource not found: " + fromResource);
                }
```

dastrobu:
- Closes: #831.

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -9,6 +9,7 @@ import static io.quarkiverse.langchain4j.deployment.LangChain4jDotNames.NO_RETRI
 import static io.quarkiverse.langchain4j.deployment.LangChain4jDotNames.SEED_MEMORY;
 import static io.quarkiverse.langchain4j.deployment.LangChain4jDotNames.V;
 
+import java.io.FileNotFoundException;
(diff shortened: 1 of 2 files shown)


geoand (approved):
Thanks a lot!

PR #844, opened 2024-08-30
Add `streamingChatLanguageModelSupplier` property to `@RegisterAiService`

### LINKED ISSUE #842: Support custom `StreamingChatLanguageModel` suppliers
`RegisterAiService` currently allows adding a `chatLanguageModelSupplier`. Unfortunately, I could not find a straightforward way of implementing a custom `Supplier<StreamingChatLanguageModel>`. 

I'd suggest adding an additional property `streamingChatLanguageModelSupplier` to `RegisterAiService`, so we can use custom models.

I could not find a viable alternative for providing my own implementation, except for writing my own extension (like in `quarkus-langchain4j-azure-openai`), which seems to be overkill for that task. 

I also think `quarkus-langchain4j` would profit from that feature, as it would allow to add more test cases with `StreamingChatLanguageModel`s, when a custom model can be implemented like in: 

```java
public class StreamingTest {

    static class MyModelSupplier implements Supplier<StreamingChatLanguageModel> {
        @Override
        public StreamingChatLanguageModel get() {
            return (messages, handler) -> {
                handler.onComplete(new Response<>(new AiMessage("42")));
            };
        }
    }
    
    @RegisterAiService(streamingChatLanguageModelSupplier = MyModelSupplier.class)
    interface MyService {
        @UserMessage("What is the Answer to the Ultimate Question of Life, the Universe, and Everything?")
        Multi<String> chat();
    }

// ...
}
```

dastrobu:
This is need to support custom `StreamingChatLanguageModel` in Ai Services

- Closes: #842

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -27,6 +27,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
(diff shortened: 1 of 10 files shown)


geoand:
Thanks!

I'll review the PR next week!

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -359,6 +362,20 @@ public void findDeclarativeServices(CombinedIndexBuildItem indexBuildItem,
         }
     }
 
+    private DotName dotName(AnnotationValue instanceAnnotation, DotName supplierDotName, Consumer<DotName> validator) {
```
geoand, line 365: Can we please have a better name for this method?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/AiServicesRecorder.java
```diff
@@ -252,4 +254,11 @@ public T apply(SyntheticCreationalContext<T> creationalContext) {
             }
         };
     }
+
+    private static <T> Supplier<T> getSupplier(String className) throws InstantiationException, IllegalAccessException,
```
geoand, line 258:
```suggestion
    private static <T> Supplier<T> createSupplier(String className) throws InstantiationException, IllegalAccessException,
```

geoand (commented):
Looks good!

I just added a couple small comments

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
dastrobu: changed it to `getSupplierDotName`, maybe you have an even better idea?

geoand: Sounds good!

geoand (approved):

PR #846, opened 2024-08-30
Execute tools for streaming models on `Infrastructure.getDefaultWorker`

### LINKED ISSUE #845: Tool executions on streaming models lead to `The current thread cannot be blocked: vert.x-eventloop-thread-0`
When combining a streaming model with a tool, that does e.g. a rest call, it leads to the following error message: 

```
The current thread cannot be blocked: vert.x-eventloop-thread-0
```

Here is an example: 

```java

    @Singleton
    static class MyTool {

        @Tool(value = "the answer to life, the universe and everything")
        public String answer() {
            return Uni.createFrom().item("42")
                    .onItem().delayIt().by(Duration.ofMillis(500))
                    .await().indefinitely();
        }
    }

    @RegisterAiService(tools = MyTool.class)
    interface MyService {
        Multi<String> chat(String msg);
    }

```

The reason is, that 

https://github.com/quarkiverse/quarkus-langchain4j/blob/0cd6f69fc7092b67534e7d60ef2ec87df4d8f88d/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java#L572

tries to execute the tool on the event loop. 

I think this should be easily fixable by running it on `Infrastructure.getDefaultWorkerPool()`. 

I can contribute a fix, but implementing a test will not be possible without merging first #844. 

This is not to be confused withe the request for supporting `Uni` as in #837.

dastrobu:
This fixes `The current thread cannot be blocked: vert.x-eventloop-thread-0` on tool executions.

Closes #845

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/StreamingChatLanguageModelToolTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/StreamingChatLanguageModelToolTest.java
@@ -1,0 +1,117 @@
+package io.quarkiverse.langchain4j.test;
+
+import static io.restassured.RestAssured.get;
+import static org.hamcrest.Matchers.containsString;
+
+import java.time.Duration;
+import java.util.List;
+import java.util.function.Supplier;
+
+import jakarta.inject.Singleton;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import dev.langchain4j.agent.tool.Tool;
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.data.message.ToolExecutionResultMessage;
+import dev.langchain4j.data.message.UserMessage;
+import dev.langchain4j.model.StreamingResponseHandler;
+import dev.langchain4j.model.chat.StreamingChatLanguageModel;
+import dev.langchain4j.model.output.Response;
+import io.quarkiverse.langchain4j.RegisterAiService;
+import io.quarkus.logging.Log;
+import io.quarkus.test.QuarkusUnitTest;
+import io.smallrye.mutiny.Multi;
+import io.smallrye.mutiny.Uni;
+
+public class StreamingChatLanguageModelToolTest {
+    private static final String TOOL_NAME = "tool";
+
+    @Path("/test")
+    static class MyResource {
+
+        private final MyService service;
+
+        MyResource(MyService service) {
+            this.service = service;
+        }
+
+        @GET
+        public Multi<String> streaming() {
+            return service.chat("what is the Answer to the Ultimate Question of Life, the Universe, and Everything?");
+        }
+    }
+
+    public static class MyModelSupplier implements Supplier<StreamingChatLanguageModel> {
+        @Override
+        public StreamingChatLanguageModel get() {
+            ToolExecutionRequest.Builder builder = ToolExecutionRequest.builder();
+            builder.name(TOOL_NAME);
+            ToolExecutionRequest toolExecutionRequest = builder.build();
+            return new StreamingChatLanguageModel() {
+                @Override
+                public void generate(List<ChatMessage> messages, List<ToolSpecification> toolSpecifications,
+                        StreamingResponseHandler<AiMessage> handler) {
+                    ChatMessage last = messages.get(messages.size() - 1);
+                    if (last instanceof UserMessage) {
+                        Log.info("requesting tool execution");
+                        handler.onComplete(Response.from(AiMessage.from(toolExecutionRequest)));
+                    } else if (last instanceof ToolExecutionResultMessage) {
+                        Log.info("processing tool execution result");
+                        String text = ((ToolExecutionResultMessage) last).text();
+                        handler.onNext(text);
+                        handler.onComplete(Response.from(AiMessage.from("")));
+                    } else {
+                        throw new IllegalStateException();
+                    }
+                }
+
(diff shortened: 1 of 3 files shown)


geoand:
Thanks for looking into this.

I need to take a closer look as I am pretty sure this will cause trouble in some circumstances.

dastrobu:
Sure, there is a unit test for reproducing the issue from #845 in [StreamingChatLanguageModelToolTest](https://github.com/dastrobu/quarkus-langchain4j/blob/execute-tool-on-infra/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/StreamingChatLanguageModelToolTest.java)

This should help adressing the issue. Maybe you could provide test cases for those "circumstances" you have in mind, as the build is passing for this PR. Then I can look into the issues. 

I can cleanup this draft PR as soon as #844 is merged. Then it will also be easier to add test cases and fix the implementation.

geoand:
I just merged https://github.com/quarkiverse/quarkus-langchain4j/pull/844 so indeed this now needs a rebase. 

I plan on reviewing this today, so I'll let you know soon enough

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/StreamingChatLanguageModelToolTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+        private final MyService service;
+
+        MyResource(MyService service) {
+            this.service = service;
+        }
+
+        @GET
+        public Multi<String> streaming() {
+            return service.chat("what is the Answer to the Ultimate Question of Life, the Universe, and Everything?");
```
geoand, line 48: I would be interested in seeing which thread the items are emitted on when tools are used.

dastrobu: Currently, all calls are executed on the event-loop thread. 

So, waiting in a call to generate a response will also fail: 

```
    public static class MyModelSupplier implements Supplier<StreamingChatLanguageModel> {
        @Override
        public StreamingChatLanguageModel get() {
            return (messages, handler) -> {
                handler.onNext("4");

                // block thread intentionally to verify there is no: The current thread cannot be blocked: vert.x-eventloop-thread-1
                Uni.createFrom().voidItem().onItem().delayIt().by(Duration.ofMillis(500)).await().indefinitely();

                handler.onNext("2");
                handler.onComplete(new Response<>(new AiMessage("")));
            };
        }
    }
```

However, this is probably not something we need to support, as the API is callback-based.

A nicer implementation would be 

```
            return (messages, handler) -> {
                Multi.createFrom().items("4", "2").onCompletion().invoke(() -> {
                            handler.onComplete(new Response<>(new AiMessage("")));
                        })
                        .subscribe().with(handler::onNext);
```

This is different for tool calls, though, where a response needs to be returned blocking, as all calls after `handler.onComplete(Response.from(AiMessage.from(toolExecutionRequest)));` run synchronously on one thread. 

The question is, what is the benefit of running on the event loop?

Here are some logs, executing the test case from the main branch:

```
2024-09-02 15:19:20,973 INFO  [io.qua.lan.tes.StreamingChatLanguageModelToolTest$MyModelSupplier$1] (vert.x-eventloop-thread-1) requesting tool execution
2024-09-02 15:19:20,976 INFO  [io.qua.lan.tes.StreamingChatLanguageModelToolTest$MyTool] (vert.x-eventloop-thread-1) executing tool call
2024-09-02 15:19:20,983 ERROR [io.qua.lan.run.too.QuarkusToolExecutor] (vert.x-eventloop-thread-1) Error while executing tool 'class io.quarkiverse.langchain4j.test.StreamingChatLanguageModelToolTest$MyTool': java.lang.IllegalStateException: The current thread cannot be blocked: vert.x-eventloop-thread-1
```

And here, after the change to `Infrastructure.getDefaultWorkerPool()`:

```
2024-09-02 15:21:36,177 INFO  [io.qua.lan.tes.StreamingChatLanguageModelToolTest$MyModelSupplier$1] (executor-thread-1) requesting tool execution
2024-09-02 15:21:36,180 INFO  [io.qua.lan.tes.StreamingChatLanguageModelToolTest$MyTool] (executor-thread-1) executing tool call
2024-09-02 15:21:37,124 INFO  [io.qua.lan.tes.StreamingChatLanguageModelToolTest$MyModelSupplier$1] (executor-thread-1) processing tool execution result
```

Note that there is already some offloading to `Infrastructure.getDefaultWorkerPool()` in [this line of code](https://github.com/quarkiverse/quarkus-langchain4j/blob/2b014289dd3f8c2d14bc6494bae76a0b66a0cbea/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java#L151).

After looking again through the code, I see the options of  
 * running `AiServiceTokenStream` on `Infrastructure.getDefaultWorkerPool` as proposed here. 
 * requesting an API extension from langchain to support an asynchronous API for tool executions.

dastrobu: I missed your comment https://github.com/quarkiverse/quarkus-langchain4j/pull/846#issuecomment-2324667832, let me check....

geoand: > Note that there is already some offloading to Infrastructure.getDefaultWorkerPool() in [this line of code](https://github.com/quarkiverse/quarkus-langchain4j/blob/2b014289dd3f8c2d14bc6494bae76a0b66a0cbea/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java#L151).

Indeed, but that happens before the `Multi` is created. 

> requesting an API extension from langchain to support an asynchronous API for tool executions.

Yeah, this is a far better option as `running AiServiceTokenStream on Infrastructure.getDefaultWorkerPool as proposed here` is not the proper way to go about this

dastrobu: Resolving this conversation in favor of the main conversation on the PR.

geoand:
I'm glad it works!

I agree it's not a great solution and I want to put a little more thought into what can be done for this situation

PR #859, opened 2024-09-04
[WATSONX] Enable tool functionality for the model llama3.1

andreadimaio:
Added the ability to use tools with models:
- `meta-llama/llama-3-405b-instruct`
- `meta-llama/llama-3-1-70b-instruct`


geoand:
Thanks!

PR #868, opened 2024-09-05
Evaluate 'required' attribute of @P annotation in tool for optional parameters.

jpohlmeyer:
Currently tool parameters are always added as required parameters.
This PR evaluates the required attribute of the P annotation and adds not required parameters as optional parameters to the ToolSpecification.


geoand (approved):
Thanks!

PR #883, opened 2024-09-10
Avoid using exceptions for control flow in guardrail's API

mariofusco:
This is a tentative and incomplete proposal to avoid exception in control flow for the new Guardrails API. For now I've done this only for the input part, but of course the same should be done also for the output if the general design is approved.


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/GuardrailsSupport.java
```diff
@@ -140,15 +143,26 @@ private static InputGuardRailsResult invokeInputGuardRails(AiServiceMethodCreate
...
 
+        InputGuardrailResult result = InputGuardrailResult.success();
+
         for (Class<? extends InputGuardrail> bean : classes) {
-            try {
-                CDI.current().select(bean).get().validate(params);
-            } catch (Exception e) {
-                return new InputGuardRailsResult(false, bean, e);
-            }
+            result = compose(result, CDI.current().select(bean).get().validate(params).validatedBy(bean));
```
mariofusco, line 149: Not using exceptions as a result for a validation, allow to easily compose more validations. This implies a change in the behaviour: now all the validations are performed regardless of the result instead of stopping at the first failure. I see this as an improvement since, in general, when performing a set of validations you want to run all of them and have a complete list of everything that went wrong instead of stopping at the first one. However I don't know if this reasoning applies well also to this specific context, but if not it will be easy to revert it and stop at the first failure. Even better we could leave this decision to users and allow them to return a `InputGuardrailResult` that can be _terminal_ (no other validation will be executed) or not.

geoand:
cc @cescoffier

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/InputGuardrailResult.java
```diff
@@ -0,0 +1,112 @@
...
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class InputGuardrailResult {
+
+    private static final InputGuardrailResult SUCCESS = new InputGuardrailResult();
+
+    private final boolean success;
+    private List<Failure> failures;
```
mariofusco, line 11: Since the validations are composable now, we accumulate all the failures here. This could be removed if we decide that we don't want composition in any case.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/InputGuardrailResult.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public static InputGuardrailResult success() {
+        return SUCCESS;
+    }
+
+    public static InputGuardrailResult fail(String message) {
+        return new InputGuardrailResult(List.of(new Failure(message)));
+    }
+
+    public static InputGuardrailResult fail(String message, Throwable cause) {
```
mariofusco, line 30: Each validation failure has a message and optionally also an exception that can be used to track the original cause.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/InputGuardrailResult.java
cescoffier: I think we need both. 

It was designed to not do that but allows complex guards to used suppressed exceptions to give multiple validation issues (like a guard using bean validation, or a guard detecting offensive words (each of them in a suppressed). We need a way for a single guard to pass multiple potential failures.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/GuardrailsSupport.java
cescoffier: Maybe failure needs a 'fatal' flag that gives us the possibility to do both.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/GuardrailsSupport.java
```diff
@@ -15,18 +9,28 @@
...
-        if (!result.success()) {
-            throw new GuardrailException(
-                    "Input validation failed. The guardrail " + result.bean().getName() + " thrown an exception",
-                    result.failure());
+        InputGuardrailResult result = null;
+        try {
+            result = invokeInputGuardRails(methodCreateInfo,
+                    new InputGuardrail.InputGuardrailParams(userMessage, chatMemory, augmentationResult));
+        } catch (Exception e) {
+            throw new GuardrailException(e.getMessage(), e);
```
cescoffier, line 30: Why do you have both validation exception and guardrail exception. I think the latter is enough

mariofusco: This is necessary to cover the case when a `validate` implementation still throw a runtime exception like in [this case](https://github.com/quarkiverse/quarkus-langchain4j/pull/883/files#diff-20a556aa33caf2e1c7a85c82a63dfacc54d8cc0b88f31a615e37de63eed4c2a3R168).

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/guardrails/OutputGuardrailValidationTest.java
```diff
@@ -91,14 +92,6 @@ void testFatalException() {
         assertThat(fatal.spy()).isEqualTo(1);
     }
 
-    @Test
-    @ActivateRequestContext
-    void testRepromptingWithoutRetry() {
```
mariofusco, line 96: The new API makes impossible to ask for a reprompt without a retry, so I simply removed this test.

mariofusco:
I also implemented the refactor that I mentioned in my former comment. 
@cescoffier @geoand please give a look.

gsmet:
This is a bit unrelated to this patch but a bit related to this Guardrail feature as I have been experimenting with Quarkus LangChain4j so I thought I could share it here.

I'm doing various AI tasks related to GitHub issues. In most prompts, I use a class delimiter approach which is prone to prompt injection.

Now where it is related to this task is that I don't want to fail, I just want to make sure the input doesn't contain the delimiter (and I can replace it with something else as it won't change the meaning).

So I don't want an error, I actually just want to filter the input so that it doesn't contain the delimiter.

mariofusco:
> So I don't want an error, I actually just want to filter the input so that it doesn't contain the delimiter.

This is a very interesting feedback, thanks a lot. Something definitively to consider and very likely to be implemented as a follow up to this commit (or independently to it anyway). I'd be glad to work also on this.

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/guardrails/OutputGuardrailValidationTest.java
cescoffier: +1 - I was very unhappy about that.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/Guardrail.java
```diff
@@ -0,0 +1,5 @@
+package io.quarkiverse.langchain4j.guardrails;
+
+public interface Guardrail<P extends GuardrailParams, R extends GuardrailResult<R>> {
```
cescoffier, line 3: Don't you fear clashes with this if we have a bean implementing both input and output guardrails?

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailParams.java
```diff
@@ -0,0 +1,4 @@
+package io.quarkiverse.langchain4j.guardrails;
+
+public interface GuardrailParams {
```
cescoffier, line 3: Don't we have some common aspect? MemoryStore and AugmentationResults?

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailResult.java
```diff
@@ -0,0 +1,58 @@
...
+
+    Throwable getFirstFailureException();
+
+    GR validatedBy(Class<? extends Guardrail> guardrailClass);
+
+    class Failure {
+        private final String message;
+        private final Throwable cause;
+
+        private Class<? extends Guardrail> guardrailClass;
```
cescoffier, line 26: should be guarded.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/InputGuardrailResult.java
```diff
@@ -0,0 +1,92 @@
...
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class InputGuardrailResult implements GuardrailResult<InputGuardrailResult> {
+
+    private static final InputGuardrailResult SUCCESS = new InputGuardrailResult();
+
+    private final Result result;
+    private List<? extends Failure> failures;
```
cescoffier, line 11: should be guarded.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/OutputGuardrail.java
```diff
@@ -122,7 +50,8 @@ default void validate(OutputGuardrailParams params)
      * @param memory the memory, can be {@code null} or empty
      * @param augmentationResult the augmentation result, can be {@code null}
      */
-    record OutputGuardrailParams(AiMessage responseFromLLM, ChatMemory memory, AugmentationResult augmentationResult) {
+    record OutputGuardrailParams(AiMessage responseFromLLM, ChatMemory memory,
+            AugmentationResult augmentationResult) implements GuardrailParams {
```
cescoffier, line 54: I didn't know that records could implement interfaces. That may explain why the interface is just a marker. Can we try to define memory() and augmentationResult() in the interface? Would that work? 

(again discovering this feature)

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/OutputGuardrailResult.java
```diff
@@ -0,0 +1,154 @@
...
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class OutputGuardrailResult implements GuardrailResult<OutputGuardrailResult> {
+
+    private static final OutputGuardrailResult SUCCESS = new OutputGuardrailResult();
+
+    private final Result result;
+    private List<Failure> failures;
```
cescoffier, line 11: should be guarded - actually, can it be final?

cescoffier (changes requested):
Looks good! 

Can you update the documentation?

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailResult.java
mariofusco: > should be guarded

Sorry, but I don't understand what you mean, can you clarify?

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/OutputGuardrail.java
mariofusco: Done.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/Guardrail.java
mariofusco: This indeed makes impossible for a single class to implement both guardrails, and this [latest commit](https://github.com/quarkiverse/quarkus-langchain4j/pull/883/commits/96c09e208df75abda50ded7c24482074b3c909c2) worsen the situation even more, but also makes the API much nicer to be used because now the methods to create the guardrail results are directly available in the respective interface. 

I believe that not allowing a single class to to implement both guardrail interfaces is really a negligible problem (in all honesty I don't think I would like to do this even if possible), but let me know if you think that this is a showstopper instead.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/InputGuardrailResult.java
mariofusco: I made it final.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/OutputGuardrailResult.java
mariofusco: I made it final.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailResult.java
cescoffier: it can be written and read concurrently. So, we either make it final (with a copy on write implementation), or make sure there is concurrent access (not to the list itself, but to the field)

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/Guardrail.java
cescoffier: I agree and actually rejected that use case in my initial design (see my response to Sergei's comment).

If someone needs to do that, they can just create 3 beans: the bean with the common logic, and the beans with the guards.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/OutputGuardrail.java
cescoffier: I learn a thing about records today!

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailResult.java
mariofusco: I structurally solved the problem turning all classes belonging to the API into records.

mariofusco:
@cescoffier I added Javadocs, rewrote documentation and squashed everything into a single commit.

geoand:
@cescoffier are you good with the latest changes?

cescoffier (approved):

cescoffier:
Yes, all good.

geoand:
🙏🏽

mariofusco:
> Yes, all good.

Thanks a lot!

Issue #861, opened 2024-09-04
No bean found for required type [interface dev.langchain4j.model.embedding.EmbeddingModel]


andreadimaio:
Since version `0.17.0` the `EmbeddingModel` is not injected correctly.
If I downgrade to `0.16.4` the problem disappears. 

```xml
  <dependency>
            <groupId>io.quarkiverse.langchain4j</groupId>
            <artifactId>quarkus-langchain4j-watsonx</artifactId>
            <version>0.17.0</version>
        </dependency>
        <dependency>
            <groupId>io.quarkiverse.langchain4j</groupId>
            <artifactId>quarkus-langchain4j-milvus</artifactId>
            <version>0.17.0</version>
        </dependency>
   ...
```

```
2024-09-04 15:28:53,034 ERROR [io.qua.run.boo.StartupActionImpl] (Quarkus Main Thread) Error running Quarkus: java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:118)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at io.quarkus.runner.bootstrap.StartupActionImpl$1.run(StartupActionImpl.java:116)
        at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.ExceptionInInitializerError
        at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized0(Native Method)
        at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized(Unsafe.java:1160)
        at java.base/jdk.internal.reflect.MethodHandleAccessorFactory.ensureClassInitialized(MethodHandleAccessorFactory.java:300)
        at java.base/jdk.internal.reflect.MethodHandleAccessorFactory.newConstructorAccessor(MethodHandleAccessorFactory.java:103)
        at java.base/jdk.internal.reflect.ReflectionFactory.newConstructorAccessor(ReflectionFactory.java:200)
        at java.base/java.lang.reflect.Constructor.acquireConstructorAccessor(Constructor.java:549)
        at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486)
        at io.quarkus.runtime.Quarkus.run(Quarkus.java:70)
[...]

PR #886, opened 2024-09-11
Don't fail booting in Dev mode when embedding store is configured but…

jmartisk:
… no embedding model exists

Fixes: https://github.com/quarkiverse/quarkus-langchain4j/issues/861

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/EmbeddingStoreJsonRPCService.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/EmbeddingStoreJsonRPCService.java
@@ -1,36 +1,78 @@
 package io.quarkiverse.langchain4j.runtime.devui;
 
+import java.util.List;
 import java.util.regex.Pattern;
 
 import jakarta.enterprise.inject.Default;
-import jakarta.enterprise.inject.spi.CDI;
-import jakarta.enterprise.util.TypeLiteral;
+
+import org.jboss.logging.Logger;
 
 import dev.langchain4j.data.document.Metadata;
 import dev.langchain4j.data.segment.TextSegment;
 import dev.langchain4j.model.embedding.EmbeddingModel;
 import dev.langchain4j.store.embedding.EmbeddingMatch;
 import dev.langchain4j.store.embedding.EmbeddingStore;
+import io.quarkus.arc.All;
+import io.quarkus.arc.Arc;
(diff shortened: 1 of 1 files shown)


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/EmbeddingStoreJsonRPCService.java
```diff
@@ -20,9 +23,17 @@ public class EmbeddingStoreJsonRPCService {
...
 
+    private static Logger log = Logger.getLogger(EmbeddingStoreJsonRPCService.class);
+
     public EmbeddingStoreJsonRPCService() {
         // if there are more models/stores, try to choose the default
-        embeddingModel = CDI.current().select(EmbeddingModel.class, new Default.Literal()).get();
+        try {
+            embeddingModel = CDI.current().select(EmbeddingModel.class, new Default.Literal()).get();
+        } catch (UnsatisfiedResolutionException ex) {
+            log.info("");
```
jmartisk, line 33: I have zero idea why, but if I remove this, the log doesn't appear (or gets garbled up together with the log saying "powered by Quarkus...") @geoand any idea?

jmartisk: We could also simply remove these logging messages and only show errors on the dev UI page.

jmartisk:
I've changed the lookup strategy a bit, it should now look into all embedding models (and also stores) and pick the first one if there are multiple

geoand (approved):

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/EmbeddingStoreJsonRPCService.java
jmartisk: I removed it in the end because I wasn't able to reproduce it anymore today. Probably some weirdly compiled Quarkus...

jmartisk:
Weirdly, some dev ui tests failed on jdk 21, not on 17.. going to look tomorrow

geoand:
👍🏽

jmartisk:
Made it more predictable by attempting to choose the model that has a `@Default` qualifier if there are multiple models available. That should also fix the test.

geoand (approved):

Issue #888, opened 2024-09-13
Enable configuration of AI service method parameters allowed annotations


aldettinger:
Please find this proposal to improve the developer experience when using Camel + LangChain4j with Quarkus.

Actually, Camel has some [parameter binding annotations](https://camel.apache.org/manual/parameter-binding-annotations.html) that greatly improve developers life. We could expect such annotations to have dot name like below:
 * @org.apache.camel.jsonpath.JsonPath
 * @org.apache.camel.Header

It would be great if we could write a given AI service method as below:
```
@UserMessage("Prompt with {text} and {dateFormat}")
CustomPojo extractFromText(@JsonPath("$.content") String text, @Header("expectedDateFormat") String dateFormat);
```

However, today, up to version 0.18.0.CR1, it's not working and the following stack trace is output:
```
io.quarkus.qute.TemplateException: Rendering error: Key "text" not found in the template data map with keys [current_date, response_schema, current_time, current_date_time] in expression {text}
```

The proposed pr #889 open such possibilities. A concrete example of usage could be found [here](https://github.com/aldettinger/camel-quarkus-examples/tree/data-extract-experiment-with-annotations-binding/data-extract-langchain4j).

aldettinger:
As discussed [here](https://github.com/quarkiverse/quarkus-langchain4j/pull/889#discussion_r1758749292), we'll experiment implementing a more generic solution.

aldettinger:
Nothing left on my side, we could probably close this ticket. target version is probably 0.18.1 as discussed. Many thanks for help :+1:

PR #889, opened 2024-09-13
Enable configuration of AI service method parameters allowed annotations #888

aldettinger:


core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -1059,9 +1059,10 @@ private List<TemplateParameterInfo> gatherTemplateParamInfo(List<MethodParameter
         for (MethodParameterInfo param : params) {
             List<AnnotationInstance> effectiveParamAnnotations = effectiveParamAnnotations(param);
             if (effectiveParamAnnotations.isEmpty() // if a parameter has no annotations it is considered a template variable
-                    || effectiveParamAnnotations.stream().map(AnnotationInstance::name).anyMatch(MEMORY_ID::equals) // we allow @MemoryId parameters to be
-                                                                                                                    // part of the template
-            ) {
+                    || effectiveParamAnnotations.stream().map(AnnotationInstance::name).anyMatch(dn -> {
+                        // we allow @MemoryId and Camel annotated parameters to be part of the template
+                        return dn.toString().startsWith("org.apache.camel") || MEMORY_ID.equals(dn);
```
cescoffier, line 1064: I'm not sure we want such a hardcoded approach. 
Is there a way to find parameters from jandex?

cescoffier: Maybe we can imagine a specific Camel extension that would produce build items we would use here to map parameters?

cescoffier: (The reason behind this is that we would need to do the same for the HTTP header and KAfka keys, and it will be very wild quickly.)

cescoffier (changes requested):

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
geoand: Yeah, we definitely need something flexible like Clement proposes. 

@aldettinger is there a Camel extension that is used with Quarkus LangChain4j?

aldettinger: @cescoffier Ok, it looks there is good incentive for a generic solution right now. On the top my head, let's try to experiment with a `AiServiceMethodParameterAllowedAnnotationBuildItem`. Better name is welcome by the way ;)

@geoand There is already a `camel-quarkus-langchain4j-chat` extension, focusing on chat.  And, I think this pr use case would justify adding a `camel-quarkus-langchain4j` focusing on high level `AiService`. Let's experiment.

geoand: > And, I think this pr use case would justify adding a camel-quarkus-langchain4j focusing on high level AiService

Right, that's where I was getting at :)

aldettinger: Please, find the last update.

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/MethodParameterAsTemplateVariableAllowance.java
```diff
@@ -0,0 +1,7 @@
+package io.quarkiverse.langchain4j.deployment;
+
+public enum MethodParameterAsTemplateVariableAllowance {
```
geoand, line 3: I would like some Javadoc about this class

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/items/MethodParameterAllowedAnnotationsBuildItem.java
```diff
@@ -0,0 +1,20 @@
+package io.quarkiverse.langchain4j.deployment.items;
+
+import io.quarkus.builder.item.MultiBuildItem;
+
+/**
+ * All Ai service method parameters with annotation starting with allowedPrefix are allowed to be part of the prompt
```
geoand, line 6:
```suggestion
 * All Ai service method parameters with annotations starting with {@code allowedPrefix} are allowed to be part of the prompt
```

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/AiServiceMethodParametersAnnotationsTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.model.chat.ChatLanguageModel;
+import dev.langchain4j.model.output.Response;
+import dev.langchain4j.service.MemoryId;
+import dev.langchain4j.service.UserMessage;
+import io.quarkiverse.langchain4j.RegisterAiService;
+import io.quarkus.test.QuarkusUnitTest;
+
+public class AiServiceMethodParametersAnnotationsTest {
```
geoand, line 26: I don't see how this test tests the new functionality. Can you please explain it? 

Thanks

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/items/MethodParameterAllowedAnnotationsBuildItem.java
```diff
@@ -0,0 +1,20 @@
...
+ * template.
+ */
+public final class MethodParameterAllowedAnnotationsBuildItem extends MultiBuildItem {
+
+    private final String allowedPrefix;
```
geoand, line 11: Wouldn't it be better for the build item to use a `Predicate` for deciding?

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -1088,20 +1099,33 @@ private List<TemplateParameterInfo> gatherTemplateParamInfo(List<MethodParameter
...
+            } else {
+                // Any annotations matching predicates below are ignored
+                if (name.startsWith("kotlin") || name.startsWith("jakarta.validation.constraints")) {
+                    return IGNORE;
+                }
+                if (name.endsWith("NotNull")) {
+                    return IGNORE;
+                }
+                if (name.startsWith("io.opentelemetry")) {
+                    return IGNORE;
```
geoand, line 1120: Wouldn't it make sense to have a new Build Item that configures these `IGNORE` values in a similar way as `MethodParameterAllowedAnnotationsBuildItem` configures `FORCE_ALLOW`?

aldettinger: Not sure, it might probably be useful later. Probably beyond the scope of this pr ?

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/items/MethodParameterAllowedAnnotationsBuildItem.java
aldettinger: Looks more flexible indeed. Let's dig that path.

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
geoand: Yeah, you are probably right

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/items/MethodParameterAllowedAnnotationsBuildItem.java
geoand: 👍🏽

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/AiServiceMethodParametersAnnotationsTest.java
geoand: > Is there by any chance a proved way to generate a build item for test purpose only ? That would enhance explainability and maintenance IMHO.

Yes, there is. See [this](https://github.com/quarkusio/quarkus/blob/3.15.0.CR1/extensions/mongodb-client/deployment/src/test/java/io/quarkus/mongodb/MongoClientBuildItemConsumerTest.java) example.

geoand:
Looks good!

Can you squash the commits and also add the test?

Thanks!

aldettinger:
> Looks good!
> 
> Can you squash the commits and also add the test?
> 
> Thanks!

Sure, it will be time to rebase/push when the test is ready :+1:

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/AiServiceMethodParametersAnnotationsTest.java
aldettinger: Ok, the test should be more explicit now.

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -10,6 +10,7 @@
 import static io.quarkiverse.langchain4j.deployment.LangChain4jDotNames.OUTPUT_GUARDRAILS;
 import static io.quarkiverse.langchain4j.deployment.LangChain4jDotNames.SEED_MEMORY;
 import static io.quarkiverse.langchain4j.deployment.LangChain4jDotNames.V;
+import static io.quarkiverse.langchain4j.deployment.MethodParameterAsTemplateVariableAllowance.*;
```
geoand, line 13: Can we get rid of the star import please?

geoand (commented):
Thanks!

I only added a small comment

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
aldettinger: Ok, it should be fine with last force push.

aldettinger:
Finger crossed for ci :crossed_fingers:

Also, do you think this kind of changes could be candidate for a potential 0.18.1 release, if any ? Is there a need to backport on maintenance branch or so ?

geoand:
We can add it to `0.18.1` as it's a small change and this repository is expected to move fast anyway

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/MethodParameterAsTemplateVariableAllowance.java
cescoffier: I found this idea quite interesting! Nice!

cescoffier (approved):

PR #903, opened 2024-09-18
Make the Quarkus version parsing a bit safer

jmartisk:
In case that somebody tries to use it with a Quarkus snapshot, etc


geoand (approved):
Good idea!

cescoffier (approved):

PR #936, opened 2024-09-27
Introduce AiService class and method names into GenAI metrics

geoand:
This is done using the Duplicated Context as currently LangChain4j doesn't provide a more fine-grained way of providing information to the ChatModelListener.
Nevertheless, the use of Duplicated Context is done in way that ensures that the tags will only be added if the Duplicated Context exists. If they don't, metrics continue to work.


cescoffier (approved):

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/MetricsChatModelListener.java
```diff
@@ -79,6 +81,16 @@ public void onResponse(ChatModelResponseContext responseContext) {
         Tags tags = Tags.of(
                 Tag.of("gen_ai.request.model", request.model()),
                 Tag.of("gen_ai.response.model", response.model()));
+        if (ContextLocals.duplicatedContextActive()) {
+            String aiServiceClassName = ContextLocals.get(AiServiceConstants.AI_SERVICE_CLASS_NAME);
+            if (aiServiceClassName != null) {
+                tags = tags.and("ai_service.class_name", aiServiceClassName);
```
cescoffier, line 87: interface or class?

geoand: Up to you, I don't mind

geoand:
@jmartisk the `org.acme.example.openai.RAGWithMetadataFilterTest` is flaky in `quarkus-langchain4j-integration-test-rag-pgvector-flyway`.

Any chance you can take a look?

Thanks

jmartisk:
> @jmartisk the `org.acme.example.openai.RAGWithMetadataFilterTest` is flaky in `quarkus-langchain4j-integration-test-rag-pgvector-flyway`.
> 
> Any chance you can take a look?
> 
> Thanks

Ok looking

geoand:
🙏🏽

jmartisk:
I ran the tests (rag-pgvector and rag-pgvector-flyway modules) locally 200 times in a loop and no failure...
The error ("cache lookup failed for type 16386") looks like something from deep inside postgres that we may not be able to control, apart from trying to use a different postgres version.
Or it may be related to the reuse of containers between test modules (I tried with Ryuk disabled as well as enabled). 
I've submitted https://github.com/quarkiverse/quarkus-langchain4j/pull/944 but whether that will help anything, I don't know

PR #939, opened 2024-09-27
Output guardrails are now called with the new response

### LINKED ISSUE #938: Output Guardrails retry/reprompt not called with the latest result
When OutputGuardrails are retrying (plain retry or reprompt), the sub-sequent calls to the guardrails are still called with the initial response (which did not pass validation).

cescoffier:
Fix #938


geoand (approved):
Any way we can add a test?

cescoffier:
I'm surprised it is not tested. Looking....

geoand (approved):

cescoffier:
@geoand The test was there, but it was testing the last message from the chat history - so everything was fine. I added assertion against the passed message.

geoand:
🆗

Issue #958, opened 2024-10-01
Security issue using QuarkusTavilyWebSearchEngine


lordofthejars:
Security issue using QuarkusTavilyWebSearchEngine as, by default, it logs the API key:

```
2024-10-01 23:53:12,259 INFO  [io.qua.lan.tav.QuarkusTavilyWebSearchEngine$TavilyClientLogger] (vert.x-eventloop-thread-2) Request:
- method: POST
- url: https://api.tavily.com/search
- headers: [Accept: application/json], [Content-Type: application/json], [User-Agent: Quarkus REST Client], [content-length: 240]
- body: {"query":"\"flights from Barcelona to Brussels\"","api_key":"API key used
```

I think it should provide the first 4 chars and the `*`

@jmartisk

lordofthejars:
I can provide a PR

jmartisk:
> I can provide a PR

That would be great, thanks :)

lordofthejars:
I'm working now on this; no worries, I'll send you a PR.

lordofthejars:
Done :)

PR #961, opened 2024-10-02
Shadows api key from tavily logs

lordofthejars:
Fixes the #958 issue. 

There are two design decisions taken here that we might discuss for sure.

The first one is that the class is local to be used by the Tavily module, but it could be located in a "common" module so that any other system can use this feature for passwords. Maybe for now, we can leave it as is,  but it is something we can have in mind.

The second was to figure out whether to use the VertX JSON feature to parse the body or start using some string manipulations. To keep things simple, I decided to use JSON instead of String manipulation utilities. I know this has an impact on performance, but it is INFO logs, so I am not sure if they would be enabled in production. Also, to avoid having potential IndexOutOfBound errors.

### DIFF
--- a/web-search-engines/tavily/runtime/src/main/java/io/quarkiverse/langchain4j/tavily/QuarkusTavilyWebSearchEngine.java
+++ b/web-search-engines/tavily/runtime/src/main/java/io/quarkiverse/langchain4j/tavily/QuarkusTavilyWebSearchEngine.java
@@ -169,7 +169,8 @@ public class QuarkusTavilyWebSearchEngine implements WebSearchEngine {
             if (body == null) {
                 return "";
             }
-            return body.toString();
+
+            return ShadowSensitiveData.process(body, "api_key");
         }
 
         private String inOneLine(io.vertx.core.MultiMap headers) {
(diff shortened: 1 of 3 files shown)


lordofthejars:
I don't know why there is a format error in the test, in theory it is autoformatted, but then the goal fails.

geoand:
I force pushed to fix the formatting issue

jmartisk:
> I force pushed to fix the formatting issue

Gah, you were 10 seconds faster.

geoand:
😆

jmartisk:
> The first one is that the class is local to be used by the Tavily module, but it could be located in a "common" module so that any other system can use this feature for passwords. Maybe for now, we can leave it as is, but it is something we can have in mind.

I don't think we can apply this easily to other modules, because Tavily has the oddity that the API key is a JSON field in the body of the request, all other services that we integrate expect it as an HTTP header.

geoand:
CI now passes

jmartisk (approved):

lordofthejars:
Great, thank you very much. Would it be possible to have a release before Devoxx?. I need to show this in the demo, and I'd prefer to not have to worry about showing console by error :)

geoand:
Yeah, I'll release today

lordofthejars:
No rush, if it is for Monday morning it is fine :)

PR #964, opened 2024-10-02
Fix ONNX Runtime Execution in Native Executable

cescoffier:
The ONNX Runtime, which relies on JNI, requires specific configuration for native image compilation due to its unique code structure. This commit adds the necessary JNI configurations to ensure native execution.

Additionally, it consolidates the native instructions related to the ONNX Runtime by moving them from being spread across `core` and `base-parsers` to solely within the `core` module. This change reflects the broader applicability of local / in-process embedding models (requiring the ONNX Runtime) beyond RAG use cases.


cescoffier:
Next step is to update the CI build to run the native tests.

geoand (approved):

jmartisk (approved):

PR #966, opened 2024-10-02
Add support for BGE 1.5 (regular and quantized) in-process embedding

### LINKED ISSUE #897: BGE Small 1.5 missing
The embedding model https://huggingface.co/BAAI/bge-small-en-v1.5 is not supported (while supported in langchain4j)

cescoffier:
- Add the BGE 1.5 local embedding 
- Update EasyRAG to exclude the dependency (it was a transitive dependency) but fix the dependency on the HuggingFace tokenizer
- Add integration tests to verify the regular and quantized model in JVM mode and native
- Add the new models to the in-process-embedding.adoc

Fix #897.


cescoffier:
@jmartisk Please note the change in EasyRAG.

cescoffier:
That's a lot of failures...

cescoffier:
Oh my... forgot the last thing I had to do.

rag/easy-rag/deployment/pom.xml
```diff
@@ -35,15 +35,9 @@
         </dependency>
         <dependency>
             <groupId>dev.langchain4j</groupId>
-            <artifactId>langchain4j-embeddings-all-minilm-l6-v2-q</artifactId>
+            <artifactId>langchain4j-embeddings-bge-small-en-v15-q</artifactId>
```
jmartisk, line 38: Why is this needed in `deployment`? Easy RAG shouldn't depend on any particular embedding model.

The documentation also says

> After adding this extension to your application, all that is needed to ingest documents into an embedding store is to add a dependency that provides an embedding model

cescoffier: It's a test dependency. 
You need an embedding model in your tests.

Also, upstream easy RAG depends on it (which turned to be a bit of a nightmare).

rag/easy-rag/runtime/pom.xml
```diff
@@ -22,8 +22,16 @@
...
                 </exclusion>
+                <exclusion>
+                    <groupId>dev.langchain4j</groupId>
+                    <artifactId>langchain4j-embeddings-bge-small-en-v15-q</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-embeddings</artifactId>
```
jmartisk, line 33: Is this necessary? If the app declares a dependency on a particular in-process embedding model, this will be brought in transitively. And if they use a remote embedding model, then this isn't necessary at all, no?

rag/easy-rag/deployment/pom.xml
jmartisk: Ah, right, I didn't notice the scope.

Ok, so, another idea: I think we could switch from depending on  `dev.langchain4j:angchain4j-easy-rag` to just `dev.langchain4j:langchain4j-document-parser-apache-tika`?  I don't think we actually use anything from the Easy RAG module so I suppose it should still work. The upstream Easy RAG is different from ours. Then you also won't have the dependency on BGE.

cescoffier: It won't work; you are still using the HuggingFaceTokenizer in your code.

cescoffier:
Could we merge this? 

We can discuss EasyRAG dependencies in another issue (it's not that simple)

jmartisk (approved):

PR #974, opened 2024-10-03
[watsonx.ai] Enable tools in streaming API

andreadimaio:
These changes enable `watsonx.ai` module to execute tools in the streaming api.
I also set the default value of the `prompt-formatter` property to `true`.

### DIFF
--- a/model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxStreamingChatModel.java
+++ b/model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxStreamingChatModel.java
@@ -6,6 +6,7 @@ import java.util.Objects;
 import java.util.concurrent.Callable;
 import java.util.function.Consumer;
 
+import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.ChatMessage;
 import dev.langchain4j.model.StreamingResponseHandler;
@@ -29,28 +30,7 @@ public class WatsonxStreamingChatModel extends WatsonxModel implements Streaming
 
     @Override
     public void generate(List<ChatMessage> messages, StreamingResponseHandler<AiMessage> handler) {
-
-        LengthPenalty lengthPenalty = null;
-        if (Objects.nonNull(decayFactor) || Objects.nonNull(startIndex)) {
-            lengthPenalty = new LengthPenalty(decayFactor, startIndex);
-        }
-
-        Parameters parameters = Parameters.builder()
-                .decodingMethod(decodingMethod)
-                .lengthPenalty(lengthPenalty)
-                .minNewTokens(minNewTokens)
-                .maxNewTokens(maxNewTokens)
-                .randomSeed(randomSeed)
-                .stopSequences(stopSequences)
-                .temperature(temperature)
-                .topP(topP)
-                .topK(topK)
-                .repetitionPenalty(repetitionPenalty)
-                .truncateInputTokens(truncateInputTokens)
-                .includeStopSequence(includeStopSequence)
-                .build();
-
-        TextGenerationRequest request = new TextGenerationRequest(modelId, projectId, toInput(messages), parameters);
+        TextGenerationRequest request = generateRequest(messages, null);
         Context context = Context.of("response", new ArrayList<TextGenerationResponse>());
 
         client.chatStreaming(request, version)
@@ -65,8 +45,13 @@ public class WatsonxStreamingChatModel extends WatsonxModel implements Streaming
                                     if (response == null || response.results() == null || response.results().isEmpty())
                                         return;
 
+                                    String chunk = response.results().get(0).generatedText();
+
+                                    if (chunk.isEmpty())
+                                        return;
+
                                     ((List<TextGenerationResponse>) context.get("response")).add(response);
-                                    handler.onNext(response.results().get(0).generatedText());
+                                    handler.onNext(chunk);
 
                                 } catch (Exception e) {
                                     handler.onError(e);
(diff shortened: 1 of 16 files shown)


geoand:
This is definitely interesting!

It seems to me that this approach can be used in other modules too, WDYT?

andreadimaio:
> It seems to me that this approach can be used in other modules too, WDYT?

Ouh.. do the other modules not have this functionality? In this case the ToolExecution state can be used as an idea to implement the execution of the tools in the Streaming API, the only thing to understand better it what we have when we use the Chat API.

andreadimaio:
The Chat API should handle this behavior automatically, shouldn't it? I will do a test with Ollama to understand the behavior.

geoand:
I am pretty sure that the streaming API doesn't currently support tool invocation, but in any case, I really need to have a look

model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxStreamingChatModel.java
```diff
@@ -113,6 +98,111 @@ public void run() {
...
     }
 
+    @Override
+    public void generate(List<ChatMessage> messages, List<ToolSpecification> toolSpecifications,
+            StreamingResponseHandler<AiMessage> handler) {
+        TextGenerationRequest request = generateRequest(messages, toolSpecifications);
+        Context context = Context.of("response", new ArrayList<TextGenerationResponse>(), "toolExecution", false);
+
+        client.chatStreaming(request, version)
+                .subscribe()
```
cescoffier, line 108: If the client calls back on an event loop, tool invocation will not be allowed to block.

model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxStreamingChatModel.java
```diff
@@ -113,6 +98,111 @@ public void run() {
...
+                .with(context,
+                        new Consumer<TextGenerationResponse>() {
+                            @Override
+                            @SuppressWarnings("unchecked")
+                            public void accept(TextGenerationResponse response) {
+                                try {
+
+                                    if (response == null || response.results() == null || response.results().isEmpty())
+                                        return;
```
cescoffier, line 117: Is it a return or actually a completion signal?

andreadimaio: This is something that should never happen. This check could be removed.

model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxStreamingChatModel.java
andreadimaio: I'm sorry, I'm not following, can you explain?

model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxStreamingChatModel.java
andreadimaio: If we want to be precise, this should be a completion signal, but I don't think an empty list ever arrives from the

andreadimaio: I'm looking at the tests I've written, and this check is necessary in this scenario:

```
id: 1
event: message
data: {}

id: 2
event: message
data: {"modelId":"ibm/granite-13b-chat-v2","model_version":"2.1.0","created_at":"2024-05-04T14:29:19.162Z","results":[{"generated_text":"He","generated_token_count":0,"input_token_count":2,"stop_reason":"not_finished"}]}

...

id: N
event: close
data: {}
```

In this case, the first message doesn't return a `NullPointException`.

model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxStreamingChatModel.java
cescoffier: This callback will be invoked on an event loop, on thread you cannot block. It means that the tools you are going to invoke will be executed on that same thread, and thus won't be able to block. Typically, you will not be able to touch a database using a blocking API, or invoke another LLM using a  non reactive method.

model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxStreamingChatModel.java
cescoffier: Interesting, so it's not a completion signal.

model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxStreamingChatModel.java
andreadimaio: Thanks for the clarification. I now realise that the problem here is that all of the tool's functions have to be "asynchronous" because the thread can't be blocked.

andreadimaio: I think this is not enough, because the `core` module does not handle the Multi return type at the moment, or so it seems. Actually, there is still some work to be done here.

cescoffier: I think there is a way forward if we can know whether the tools will be blocking or not.
1) If not blocking (returning an Uni), we may have to do some work in _core_.
2) If blocking, we can imagine switching to a worker thread before  the tools invocation and switching back after. 

2) can be done without 1) 

@geoand WDYT?

andreadimaio: Maybe the implementation is more complex than I thought. Looking at the code, `langchain4j` wants a `json` as the result of the tool call, so the `QuarkusToolExecutor` at line [66](https://github.com/quarkiverse/quarkus-langchain4j/blob/4eada1dcc8188ddfaa181468fc4fbf3e4621d842/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/QuarkusToolExecutor.java#L66) does the conversion of the result, but in this case we have a `Uni`, and of course that's something `jackson` doesn't like.

andreadimaio: If I change the `handleResult` in this way:

```java
private static String handleResult(ToolInvoker invokerInstance, Object invocationResult) {
        if (invokerInstance.methodMetadata().isReturnsVoid()) {
            return "Success";
        }

        if (invocationResult instanceof Uni uniInvocation) {
            return Json.toJson(uniInvocation.await().indefinitely());
        } else {
            return Json.toJson(invocationResult);
        }
    }
```

And I have a tool like that:

```java
@Tool("Execute the embedding of a string")
    public Uni<List<Float>> embedding(String textToEmbed) {
        return Uni.createFrom()
            .item(textToEmbed)
            .map(embeddingModel::embed)
            .map(Response::content)
            .map(Embedding::vectorAsList);
            
    }
```

Everything works, but the host I don't know if it makes sense. :thinking:

andreadimaio: I was expecting a `BlockingNotAllowedException` which was not thrown since the `Embedding` API is blocking. Mutiny is something I need to study more :smile:

andreadimaio: So from my understanding for now the only way to have this feature enabled is to force `Mutiny` to work on a worker thread, for all other scenarios some changes in `langchain4j` are needed.

```java
client.chatStreaming(request, version)
                .emitOn(Infrastructure.getDefaultWorkerPool())
                .subscribe()
```

cescoffier: Yes, I guess for the time being, that's the only choice. It's quite unfortunate as it will reduce the concurrency and response time (well, the LLM is generally taking so much time that a thread switch is invisible).

andreadimaio:
> It seems to me that this approach can be used in other modules too, WDYT?

I've done some tests with `ollama` and it seems that `/chat/api` does not return any information about the tool execution.
To use my approach you should use the `/api/generate` endpoint, but in this case you have to manage all the LLM tags. One idea might be to move the `PromptFormatter` into the core module, but I don't know if this work is worth it given the limitations discussed above.

model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxStreamingChatModel.java
andreadimaio: Maybe the virtual threads (the magic option) can ease this pain 😄

andreadimaio: > (well, the LLM is generally taking so much time that a thread switch is invisible)

This is actually true, the time taken to respond today is not low.

cescoffier: > Maybe the virtual threads (the magic option) can ease this pain 😄

You need to make sure you read the small print :-)
But it's a good point. I'm wondering if we could change the dispatching strategy based on the annotation from the tool method.

For example:

```
@Tool 
String compute(...) {
}
```

This is an imperative method, we invoke it on the worker thread pool.

If the method is annotated with `@RunOnVirtualThread` like the following:

```
@Tool 
@RunOnVirtualThread
String compute(...) {
}
```

Instead of the regular worker thread pool, you would use the virtual thread scheduler (we expose it as a bean). 

It would be great in the future to support non-blocking tools returning `Uni<T>`, but not necessary something for today as it requires changing langchain4j.

andreadimaio: > You need to make sure you read the small print :-)

I know, I know, the magic doesn't always do what it's supposed to 😄 . 

> If the method is annotated with `@RunOnVirtualThread` like the following

Today it is not possible to annotate the tools method with `@RunOnVirtualThread`, but this is something that can be very useful. In this case the developer can create a worker thread or a virtual thread based on the business logic inside the method.

andreadimaio:
From today, the `watsonx.ai` provider gives the possibility to use the `/chat` API and it seems (from a quick view) that also for the streaming version some logic can be executed to enable the use of the tools. At this point this PR doesn't make sense anymore, one point that could be extracted and discussed in another PR is the possibility of using the `@RunOnVirtualThread` annotation.

I'm going to close this PR and start working on a new PR to move the current implementation to the new `/chat` API.

@geoand @cescoffier

geoand:
Thanks for the update

Issue #962, opened 2024-10-02
[FEATURE] Select tools dynamically on incoming message


MiggiV2:
I've already worked on an analogous feature within the [langchain4j/langchain4j](https://github.com/langchain4j/langchain4j) project, where I implemented [dynamic tool registration](https://github.com/langchain4j/langchain4j/pull/1644) for LangChain. Now, I want to extend this functionality to the **quarkus-langchain4j** extension.

### **Describe the solution you'd like**

The goal is to enable **dynamic tool registration** at runtime to allow for more flexible and scalable tool integration. Instead of declaring tools upfront in a static way, we propose allowing Quarkus LangChain4j users to **register tools dynamically** and **distribute them effectively**.

#### Code Example

```java
@RegisterAiService(toolProvider = MyCustomToolProvider.class)
public interface Bot {

    @SystemMessage("You are an AI named Bob answering questions about financial products.")
    String chat(@UserMessage String question);
}
```

In this example:

- The `Bot` interface is registered with the `AIService`, while the actual tools are injected at runtime using `MyCustomToolProvider.class`.
- This allows for dynamic selection and attachment of tools based on runtime conditions, granting flexibility over tool management and usage.
- Compiling with a locked set of tools is no longer needed, making the bot easier to customize per environment or customer.

---

### **Proposal and Commitment**

We at **EXP** would be happy to contribute a pull request for this feature. I plan to share an initial implementation soon and welcome any feedback or further ideas from the community before delving deeper into the work.

Thank you very much for considering this feature request! Looking forward to engaging with everyone on this.

geoand:
Thanks a lot for reporting!

I like the idea, feel free to open a Pull Request so we can discuss further.

cc @cescoffier @andreadimaio

andreadimaio:
If I understand correctly, the idea is to start from a list of tools that can be large and send to the LLM only the list that matches the user's request. 

For example, suppose I have a class that exposes two tools, one to perform a `sum` and another to perform a `web search`.
If the user's query is: **"Execute the sum of 1 + 1"**, the prompt of the LLM will contain only the first tool. 

This allows to streamline the prompt and to avoid hallucinations.
Is this correct?

cescoffier:
But how would you select the tools  for a specific call?

andreadimaio:
> But how would you select the tools for a specific call?

Good question... embedding? In any case, it should be something that can be customised to use the best approach, but yes, the doubt remains.

[...]

PR #989, opened 2024-10-16
[FEATURE] ToolProvider | Select tools dynamically on incoming message

MiggiV2:
In this PR, the idea is to implement the selection of tools dynamically on incoming message

# New Feature
Make ToolProvider from LangChain4J Quarkus ready

We can set our ToolProvider like this:
```JAVA
@RegisterAiService(toolProviderSupplier = MyCustomToolProviderSupplier.class)
interface MyServiceWithToolProvider {
     String chat(@UserMessage String msg, @MemoryId Object id);
}
```

Add a simple Supplier like this:

```JAVA
@ApplicationScoped
public class MyCustomToolProviderSupplier implements Supplier<ToolProvider> {
    @Inject
    MyCustomToolProvider myCustomToolProvider;

    @Override
    public ToolProvider get() {
        return myCustomToolProvider;
    }
}
``` 

Now, we can select tools dynamically in our ToolProvider like this:
```Java
@ApplicationScoped
public class MyCustomToolProvider implements ToolProvider {

    @Override
    public ToolProviderResult provideTools(ToolProviderRequest request) {
        boolean containsBooking = request.userMessage().singleText().contains("booking");
	if (!containsBooking) {
	      return ToolProviderResult.builder().build();
	}

        ToolSpecification toolSpecification = ToolSpecification.builder()
                .name("get_booking_advice")
                .description("Returns booking advice")
                .build();
        ToolExecutor toolExecutor = (t, m) -> "You should book your vacation now :)";
        return ToolProviderResult.builder()
                .add(toolSpecification, toolExecutor)
                .build();
    }
}
```

- Closes: #962

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/ToolProviderMetaBuildItem.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/ToolProviderMetaBuildItem.java
@@ -1,0 +1,21 @@
+package io.quarkiverse.langchain4j.deployment;
+
+import java.util.List;
+
+import io.quarkiverse.langchain4j.deployment.devui.ToolProviderInfo;
+import io.quarkus.builder.item.SimpleBuildItem;
+
+/**
+ * Holds metadata about toolProviders discovered at build time
+ */
+public final class ToolProviderMetaBuildItem extends SimpleBuildItem {
+    List<ToolProviderInfo> metadata;
+
+    public ToolProviderMetaBuildItem(List<ToolProviderInfo> metaData) {
+        this.metadata = metaData;
+    }
+
+    public List<ToolProviderInfo> getMetadata() {
+        return metadata;
+    }
+}
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ToolProviderTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ToolProviderTest.java
@@ -1,0 +1,135 @@
+package io.quarkiverse.langchain4j.test;
+
+import static dev.langchain4j.data.message.ChatMessageType.TOOL_EXECUTION_RESULT;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+import java.util.List;
+import java.util.function.Supplier;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.context.control.ActivateRequestContext;
+import jakarta.inject.Inject;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.data.message.ToolExecutionResultMessage;
+import dev.langchain4j.model.chat.ChatLanguageModel;
+import dev.langchain4j.model.output.FinishReason;
+import dev.langchain4j.model.output.Response;
+import dev.langchain4j.model.output.TokenUsage;
+import dev.langchain4j.service.MemoryId;
+import dev.langchain4j.service.UserMessage;
+import dev.langchain4j.service.tool.ToolExecutor;
+import dev.langchain4j.service.tool.ToolProvider;
+import dev.langchain4j.service.tool.ToolProviderRequest;
+import dev.langchain4j.service.tool.ToolProviderResult;
+import io.quarkiverse.langchain4j.RegisterAiService;
+import io.quarkus.test.QuarkusUnitTest;
+
+class ToolProviderTest {
+    @Inject
+    MyServiceWithCustomToolProvider myServiceWithTools;
+
+    @Inject
+    MyServiceWithDefaultToolProviderConfig myServiceWithoutTools;
+
+    @ApplicationScoped
+    public static class MyCustomToolProviderSupplier implements Supplier<ToolProvider> {
+        @Inject
+        MyCustomToolProvider myCustomToolProvider;
+
+        @Override
+        public ToolProvider get() {
+            return myCustomToolProvider;
+        }
+    }
+
+    @ApplicationScoped
+    public static class MyCustomToolProvider implements ToolProvider {
+        @Inject
+        MyServiceWithDefaultToolProviderConfig myServiceWithoutTools;
+
+        @Override
+        public ToolProviderResult provideTools(ToolProviderRequest request) {
+            assertNotNull(myServiceWithoutTools);
+
+            ToolSpecification toolSpecification = ToolSpecification.builder()
+                    .name("get_booking_details")
+                    .description("Returns booking details")
+                    .build();
+            ToolExecutor toolExecutor = (t, m) -> "0";
+            return ToolProviderResult.builder()
+                    .add(toolSpecification, toolExecutor)
+                    .build();
+        }
+    }
+
+    public static class TestAiSupplier implements Supplier<ChatLanguageModel> {
+        @Override
+        public ChatLanguageModel get() {
+            return new TestAiModel();
+        }
+    }
+
+    public static class TestAiModel implements ChatLanguageModel {
+        @Override
+        public Response<AiMessage> generate(List<ChatMessage> messages) {
+            return new Response<>(new AiMessage("42"));
+        }
+
+        @Override
+        public Response<AiMessage> generate(List<ChatMessage> messages, List<ToolSpecification> toolSpecifications) {
+            ChatMessage lastMsg = messages.get(messages.size() - 1);
+            boolean isLastMsgToolResponse = lastMsg.type().equals(TOOL_EXECUTION_RESULT);
+            if (isLastMsgToolResponse) {
+                ToolExecutionResultMessage msg = (ToolExecutionResultMessage) lastMsg;
+                return new Response<>(new AiMessage(msg.text()));
+            }
+            ToolSpecification toolSpecification = toolSpecifications.get(0);
+            ToolExecutionRequest toolExecutionRequest = ToolExecutionRequest.builder()
+                    .name(toolSpecification.name())
+                    .id(toolSpecification.name())
+                    .build();
+            TokenUsage usage = new TokenUsage(42, 42);
+            return new Response<>(AiMessage.from(toolExecutionRequest), usage, FinishReason.TOOL_EXECUTION);
+        }
+    }
+
+    @RegisterAiService(toolProviderSupplier = MyCustomToolProviderSupplier.class, chatLanguageModelSupplier = TestAiSupplier.class)
+    interface MyServiceWithCustomToolProvider {
+        String chat(@UserMessage String msg, @MemoryId Object id);
+    }
+
+    @RegisterAiService(chatLanguageModelSupplier = BlockingChatLanguageModelSupplierTest.MyModelSupplier.class, chatMemoryProviderSupplier = RegisterAiService.NoChatMemoryProviderSupplier.class)
+    interface MyServiceWithDefaultToolProviderConfig {
+        String chat(String msg);
+    }
+
+    @RegisterExtension
+    static final QuarkusUnitTest unitTest = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(MyServiceWithCustomToolProvider.class, MyCustomToolProvider.class,
+                            BlockingChatLanguageModelSupplierTest.MyModelSupplier.class));
+
+    @Test
+    @ActivateRequestContext
+    void testCall() {
+        String answer = myServiceWithTools.chat("hello", 1);
+        assertEquals("0", answer);
+    }
+
+    @Test
+    @ActivateRequestContext
+    void testCallNoTools() {
+        String answer = myServiceWithoutTools.chat("hello");
+        assertEquals("42", answer);
+    }
+}
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
@@ -19,6 +19,7 @@ import dev.langchain4j.model.moderation.ModerationModel;
 import dev.langchain4j.rag.RetrievalAugmentor;
 import dev.langchain4j.retriever.Retriever;
 import dev.langchain4j.service.AiServices;
+import dev.langchain4j.service.tool.ToolProvider;
 import dev.langchain4j.store.memory.chat.ChatMemoryStore;
 import dev.langchain4j.store.memory.chat.InMemoryChatMemoryStore;
 import io.quarkiverse.langchain4j.audit.AuditService;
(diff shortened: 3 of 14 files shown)


core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/ToolProviderMetaBuildItem.java
```diff
@@ -0,0 +1,18 @@
+package io.quarkiverse.langchain4j.deployment;
+
+import java.util.List;
+
+import io.quarkiverse.langchain4j.deployment.devui.ToolProviderInfo;
+import io.quarkus.builder.item.SimpleBuildItem;
+
+public final class ToolProviderMetaBuildItem extends SimpleBuildItem {
```
geoand, line 8: Can we please have a little Javadoc on what the purpose of this is?

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ToolProviderTest.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    @RegisterAiService(toolProvider = MyCustomToolProvider.class, chatLanguageModelSupplier = TestAiSupplier.class)
+    interface MyServiceWithToolProvider {
+        String chat(@UserMessage String msg, @MemoryId Object id);
+    }
+
+    @RegisterAiService(chatLanguageModelSupplier = BlockingChatLanguageModelSupplierTest.MyModelSupplier.class, chatMemoryProviderSupplier = RegisterAiService.NoChatMemoryProviderSupplier.class)
+    interface MyServiceWithoutToolProvider {
```
geoand, line 99: I think the name here is a little misleading. To me, it makes to call it  `MyServiceWithDefaultToolProvider`

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ToolProviderTest.java
```diff
@@ -0,0 +1,122 @@
...
+                    .name(toolSpecification.name())
+                    .id(toolSpecification.name())
+                    .build();
+            TokenUsage usage = new TokenUsage(42, 42);
+            return new Response<>(AiMessage.from(toolExecutionRequest), usage, FinishReason.TOOL_EXECUTION);
+        }
+    }
+
+    @RegisterAiService(toolProvider = MyCustomToolProvider.class, chatLanguageModelSupplier = TestAiSupplier.class)
+    interface MyServiceWithToolProvider {
```
geoand, line 94: I think the name here is a little misleading. To me, it makes to call it  `MyServiceWithCustomToolProvider`

core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
```diff
@@ -140,6 +145,11 @@
      */
     Class<? extends Supplier<ModerationModel>> moderationModelSupplier() default BeanIfExistsModerationModelSupplier.class;
 
+    /**
+     * Configures a toolProvider. Either a toolProvider or "normal" tools can be used, but not both
+     */
+    Class<? extends ToolProvider> toolProvider() default BeanIfExistsToolProviderSupplier.class;
```
geoand, line 151: Is there a reason this doesn't follow the supplier model of all the other configuration options (for example for `AuditService`)? 

The reason this model is nice is that it allows advanced use cases to create their own implementations, so my suggestion would be to follow it.

core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
```diff
@@ -262,4 +272,16 @@ public ImageModel get() {
             throw new UnsupportedOperationException("should never be called");
         }
     }
+
+    /**
+     * Default toolProvider that does not provide any tools
+     */
+    @ApplicationScoped
```
geoand, line 279: This should definitely not be added. If you follow my suggestion about using the supplier and the implementation does pretty much what is done for `AuditService`, then this problem will go away

geoand (changes requested):
This is very nice, thanks a lot!

I added a few small comments and one important one about how `@RegisterAiService` would configure this.

Also, I would like @jmartisk to review the DevUI part

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ToolProviderTest.java
MiggiV2: I understand your proposal, but in my implementation if there is no ToolProvider specified the ToolProvider is null.  
Why null?  
In LangChain4J, we specified, that there is either a ToolProvider or static (normal) tools, but not both.

geoand: Sure yeah, but that doesn't negate the fact that as I maintainer when I look at the test I get a misleading impression about what it does.

geoand: Sure, that's good

MiggiV2:
Thank you very much for your feedback, @geoand!   
I've made some changes, and I hope they align with your suggestions. 😁

geoand:
Thanks!

I'll have a look soon

core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
```diff
@@ -262,4 +268,15 @@ public ImageModel get() {
             throw new UnsupportedOperationException("should never be called");
         }
     }
+
+    /**
+     * Default toolProviderSupplier that does not provide any tools
```
geoand, line 273: This is not correct. Please use something like:

```
    /**
     * Marker that is used to tell Quarkus to use the {@link ToolProvider} that the user has configured as a CDI bean.
     * If no such bean exists, then no tool provider will be used.
     */
```

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -159,6 +161,19 @@ private static Object doImplement(AiServiceMethodCreateInfo methodCreateInfo, Ob
...
+            toolSpecifications = new ArrayList<>();
+            toolExecutors = new HashMap<>();
+            ToolProviderRequest request = new ToolProviderRequest(memoryId, userMessage);
+            ToolProviderResult result = context.toolProvider.provideTools(request);
+            for (ToolSpecification specification : result.tools().keySet()) {
+                toolSpecifications.add(specification);
+                toolExecutors.put(specification.name(), result.tools().get(specification));
+            }
+        }
+        List<ToolSpecification> finalToolSpecifications = toolSpecifications;
```
geoand, line 174: Instead of `final..`, please use `effective...`

geoand (commented):
This is great!

I only added some small comments.
When all are addressed, please squash the commits and force push.

Thanks!

MiggiV2:
Has the DevUI part been reviewed by @jmartisk and is it okay?

geoand:
Ah, I forgot about that! Good point :)

@jmatrisk?

jmartisk:
Sorry, been traveling now. I'll have a look ASAP

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/ChatJsonRPCService.java
```diff
@@ -225,12 +231,27 @@ public ChatResultPojo chat(String message, boolean ragEnabled) {
...
             if (toolSpecifications.isEmpty()) {
                 modelResponse = model.generate(memory.messages());
                 memory.add(modelResponse.content());
             } else {
-                modelResponse = executeWithTools(memory);
+                executeWithTools(memory);
+            }
+            // Remove toolProviderSupplier tools again
+            if (hasToolProvider) {
+                toolSpecifications.clear();
```
jmartisk, line 253: IIUC, if there's a class with static tools (with annotations) AND a dynamic tool provider, then when the `ChatJsonRPCService` is initialized, then information from the "static" tools is added and made available to the model, but then after handling the first request, you clear the `toolSpecifications` and `toolExecutors`, and on the second request you only fill it again with the dynamic tools, but not the static tools (these are lost for subsequent chatting). I'm not completely sure if we want to allow mixing both of these "tool sources" in the Dev UI (it's not allowed on AI services, right?), but maybe it wouldn't hurt to pick up and combine both in this case?! 
So, maybe these `toolSpecifications` and `toolExecutors` fields should only contain the static tools, and the dynamic ones should be retrieved and handled via local variables in the chat methods. WDYT?

jmartisk: Another thing is that you don't support streaming chats right now, so after splitting `toolSpecifications` and `toolExecutors` into separate dynamic and static ones, we could also retrieve and apply the dynamic ones in the `streamingChat` method

jmartisk:
Also I'd love to see an update of the documentation (`agent-and-tools.adoc`) with this :)

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/ChatJsonRPCService.java
```diff
@@ -62,14 +65,17 @@ public class ChatJsonRPCService {
...
+    private final ToolProvider toolProvider;
 
     public ChatJsonRPCService(@All List<ChatLanguageModel> models, // don't use ChatLanguageModel model because it results in the default model not being configured
             @All List<StreamingChatLanguageModel> streamingModels,
             @All List<Supplier<RetrievalAugmentor>> retrievalAugmentorSuppliers,
             @All List<RetrievalAugmentor> retrievalAugmentors,
             ChatMemoryProvider memoryProvider,
-            QuarkusToolExecutorFactory toolExecutorFactory) {
+            QuarkusToolExecutorFactory toolExecutorFactory,
+            @All List<ToolProvider> toolProviders) {
```
jmartisk, line 76: Maybe you should request the list of all `Supplier<ToolProvider>`s instead?
IIUC a `ToolProvider` isn't required to be a CDI bean, so if it's not, then it won't appear here. But the suppliers are required to be CDI beans.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/ChatJsonRPCService.java
MiggiV2: Thanks for the feedback.
I will work on streamingChat implementation, today then. 
About mixing static tools and dynamic (toolProvider) tools. In my opinion, it is more confusing to mix them in the DevUI `ChatJsonRPCService`. The `AiService` doesn't allow that, and it would be inconsistent.
I will try out if the DevUI starts with toolProvider and static tools. If the DevUI starts, I will need to fix that.

jmartisk: Yeah it should either use both sources, or always only one, but if it picks up both on the first request and only dynamic tools on subsequent requests, that's wrong.

MiggiV2: You are right. I will fix this :)

MiggiV2: The question is, should I ignore static tools when dynamic tools are found or throw an error?
Also, in the "Tools" tab, I only show a list of `ToolProvider`, if any are discovered.

jmartisk: Hmm, it depends on how we think users will use the application. There are two options

- A: There will be a set of static tools, and a dynamic provider that provides some *extra* methods that do not correspond to any of the static tools. In this case, it would make sense to let the application use both (also an AI service should be able to use both, and so should the Dev UI page)
- B: There will be a set of static tools, and the dynamic provider's main use will be to "filter" which tools to make available at a given moment (it will return references to some of the existing static tool methods). In this case, generally only the dynamic provider should be picked up (in the Dev UI), because by picking up both, you would get duplicate tools

I assume your main target when implementing this was to support case B, right?

jmartisk: Or maybe C, where the app should have either only static, or only dynamic tools. If that's the case, we should prefer dynamic tools only, and ignore the static ones.

MiggiV2: Option C sounds right. In Langchain4J, we specified, that [either static tools or the tool provider can be configured, but not both](https://github.com/langchain4j/langchain4j/pull/1644#discussion_r1732961279) for an AiService. My last [commit 1fb3761](https://github.com/quarkiverse/quarkus-langchain4j/pull/989/commits/1fb3761fe83544df025e9f2ef07ef598fe338ce2) follows this specification :)
If a toolProvider is found, we ignore static tools and only use the provider (dynamic tools).

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/ChatJsonRPCService.java
```diff
@@ -323,4 +346,25 @@ public void onError(Throwable error) {
...
+    private ToolProvider getToolProvider(List<Supplier<ToolProvider>> toolProviders) {
+        for (Supplier<ToolProvider> provider : toolProviders) {
+            if (provider.get() != null) {
+                return provider.get();
+            }
+        }
+        return null;
+    }
+
+    private boolean setToolsViaProviderIfAvaible(ChatMemory memory, UserMessage userMessage) {
```
jmartisk, line 358: typo "Avaible"

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/devui/ChatJsonRPCService.java
jmartisk: Ok then! Let's just make that clear in the quarkus-langchain4j docs too, because as you see, I understood it differently at first :)

MiggiV2:
I will work on the documentation for this feature now. Tomorrow, I will commit my first version of it.

docs/modules/ROOT/pages/agent-and-tools.adoc
```diff
@@ -90,6 +90,90 @@ If a method of an AI Service needs to use tools other than the ones configured o
...
+}
+----
+
+=== Alternative: Using `ToolSpecification` and `ToolExecutor`
+
+If you prefer more control, you can work directly with `ToolSpecification` and `ToolExecutor` to provide tools.
+
+=== Important Rule
+
+You can configure either static tools (via `tools`) or a dynamic `ToolProvider` (via `toolProviderSupplier`), but *not both* in the same service.
```
jmartisk, line 175:
```suggestion
You can configure either static tools (via `tools`) or a dynamic `ToolProvider` (via `toolProviderSupplier`), but *not both* in the same AI service.
```

just for clarification (I assume you can have an AI service with static tools, and one with a provider, right?)

MiggiV2: yes

jmartisk (approved):
Nice!

jmartisk:
Merged, thanks :)

andreadimaio:
I'm looking at the [ToolProvider](https://github.com/langchain4j/langchain4j/blob/40575058d93ef050c22f0d4735788e3564c57a62/langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProvider.java#L20) interface, which currently only provides two parameters in `ToolProviderRequest`: `chatMemoryId` and `UserMessage`. From my perspective, it might be useful to add access to the list of available tools associated with the method. This extension would allow developers to filter from the current list of tools, or to redefine the list entirely by specifying something new based on the `UserMessage`.

Is this something that could be addressed in a new issue? WDYT?

geoand:
I think it makes sense to, yeah

geoand:
@d135-1r43 are you planning on making an improvement here?

PR #1003, opened 2024-10-22
Move inner classes to top level classes for API concision

### LINKED ISSUE #926: Suggestion: Reduce repetition of words in Guardrails API signature?
I notice that

```
public interface InputGuardrail extends Guardrail<InputGuardrail.InputGuardrailParams, InputGuardrailResult> {
```

repeats `InputGuardrail` four times in the one line. 

Instead, could we have 

```
public interface InputGuardrail extends Guardrail<InputGuardrailParams, InputGuardrailResult> {
```
(no inner class)

or 

```
public interface InputGuardrail extends Guardrail<InputGuardrail.Params, InputGuardrailResult> {
```

(inner class, but without the repeated `InputGuardrail`?)

Similarly, 

```
public interface OutputGuardrail extends Guardrail<OutputGuardrail.OutputGuardrailParams, OutputGuardrailResult> {
```

could be 

```
public interface OutputGuardrail extends Guardrail<OutputGuardrailParams, OutputGuardrailResult> {
```

or 

```
public interface OutputGuardrail extends Guardrail<OutputGuardrail.Params, OutputGuardrailResult> {
```

The option without the inner class is perhaps better, since otherwise there's a puzzling asymmetry (why is the `Params` an inner class, but `OutputGuardrailResult` gets its own top-level class?)

holly-cummins:
Resolves #926


cescoffier (approved):

cescoffier:
Thanks!

Issue #993, opened 2024-10-18
Allow configuring the tls-configuration-name for the Model providers using the REST client


cescoffier:
We have a case where the remote model provider uses a self-signed certificate (yes, yes, I know.... 2024). It would be great to have the possibility to configure the `tls-configuration-name` and allow the loading of the TLS configuration for the Quarkus TLS registry. 

I mentioned in the title model providers, but anything remote should offer that possibility.

geoand:
Interesting. We can certainly do this

geoand:
@cescoffier I assume https://github.com/quarkiverse/quarkus-langchain4j/pull/1006 is what you are after?

cescoffier:
yes, exactly. 

We have a case where we need to connect to a model server using self-signed certificates.

geoand:
👍🏽

geoand:
You'll love the hack I had to come up with the make OpenAI work... 
It still needs more work, so I'll finish the PR tomorrow.

geoand:
PR is now ready to go

PR #1006, opened 2024-10-22
Add support for TLS configuration name

geoand:
This allows the TLS configuration name to
be applied like so:

```properties
quarkus.langchain4j.ollama.tls-configuration-name=foo
```

Where it's expected that the user would have also
done something like:

```properties
quarkus.tls.foo.trust-store.jks.path=truststore.jks
quarkus.tls.foo.trust-store.jks.password=pass
```

- Resolves: #993

### DIFF
--- a/model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorder.java
+++ b/model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorder.java
@@ -33,6 +33,7 @@ import io.quarkiverse.langchain4j.azure.openai.runtime.config.EmbeddingModelConf
 import io.quarkiverse.langchain4j.azure.openai.runtime.config.LangChain4jAzureOpenAiConfig;
 import io.quarkiverse.langchain4j.azure.openai.runtime.config.LangChain4jAzureOpenAiConfig.AzureAiConfig.EndpointType;
 import io.quarkiverse.langchain4j.openai.common.QuarkusOpenAiClient;
+import io.quarkiverse.langchain4j.openai.common.runtime.AdditionalPropertiesHack;
 import io.quarkiverse.langchain4j.runtime.NamedConfigUtil;
 import io.quarkus.arc.SyntheticCreationalContext;
 import io.quarkus.runtime.ShutdownContext;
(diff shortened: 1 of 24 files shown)


model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorder.java
```diff
@@ -317,6 +318,7 @@ private static ConfigValidationException.Problem createConfigProblem(String key,
     }
 
     public void cleanUp(ShutdownContext shutdown) {
+        AdditionalPropertiesHack.reset();
```
cescoffier, line 321: mouarf....

model-providers/openai/openai-vanilla/deployment/src/test/java/io/quarkiverse/langchain4j/openai/test/OpenAiChatLanguageModelTlsConfigurationTest.java
```diff
@@ -0,0 +1,169 @@
...
+import io.smallrye.certs.junit5.Certificates;
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.HttpServer;
+import io.vertx.core.http.HttpServerOptions;
+import io.vertx.core.net.KeyCertOptions;
+import io.vertx.ext.web.Router;
+
+@Certificates(baseDir = "target/certs", certificates = {
+        @Certificate(name = "lc4j", password = "secret", formats = Format.JKS),
```
cescoffier, line 35: it's not that I don't like JKS (actually, I don't), but I think testing with P12 would be much closer to what we do in 2024.

cescoffier (approved):
One small comment about the test using JKS.

model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorder.java
geoand: Yeah, it's horrific...

model-providers/openai/openai-vanilla/deployment/src/test/java/io/quarkiverse/langchain4j/openai/test/OpenAiChatLanguageModelTlsConfigurationTest.java
geoand: No problem, I'll update it

Issue #997, opened 2024-10-20
Add Llama3.java integration 


geoand:
Similar as Jlama

lordofthejars:
I am interested in this, too, as this will open the door to inference `gguf` models in Java (Quarkus).

I think the best option, due to the nature of the Llama3.java, would be to create a Maven profile that, when activated, updates the local Llama3.java, so when you want to do a release of the module with a new version of Llama3, instead of modifying the pom file with a new version you enable a profile which downloads the new Llama3 and copies locally.

I know it is not perfect but the project is what it is after all.

geoand:
We'll see how it plays out :)

lordofthejars:
Or we can add a man in the middle aka a standalone JAR that does all these things, so from Quarkus extension, it is a JAR, but we moved the "complexity" to another project managed by us.

PR #1008, opened 2024-10-23
Introduce llama3-java module

geoand:
- Resolves: #997


model-providers/llama3-java/pom.xml
```diff
@@ -0,0 +1,24 @@
...
+        <artifactId>quarkus-langchain4j-parent</artifactId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <artifactId>quarkus-langchain4j-llama3-java-parent</artifactId>
+    <name>Quarkus LangChain4j - Llama 3 - Java - Parent</name>
+    <packaging>pom</packaging>
+
+    <properties>
+        <maven.compiler.release>21</maven.compiler.release>
```
edeandrea, line 15: Really a Java 21 baseline?

geoand: What's the alternative?

edeandrea: 17?

geoand: Vector support on 17? Don't think so...

edeandrea: Ahhhhh....You are right!

I was only looking at this from a JVM baseline - where Quarkus core baselines on 17. But you are right.

geoand:
This still needs at least the following:

 * Download models at build time
 * Allow copying the models into the fast-jar like we did for Jlama
 * Fix the build to ensure that the preview features used does not prevent the extension from being used in JDKs other than the target version


Ideally we should also have some docs about how to use this and Jlama. @mariofusco IIRC you wanted to write something about that?

geoand:
Awesome, that would go a long way to helping people experiment with these two great technologies!

geoand:
This should now be good enough for basic usage. 

In JVM mode, one would build the application with:

```
mvn package
```

and run it like so:


```
java --enable-preview -jar integration-tests/llama3-java/target/quarkus-app/quarkus-run.jar
``` 

For native mode, best use the latest GraalVM from sdkman (I used `sdk install java 24.ea.18-graal`)

and build using:

```
mvn package -Dnative
```

but keeping it mind that you probably want to have:

```properties
quarkus.native.march=native
quarkus.native.additional-build-args=-O3
```

in `application.properties` for best performance.

After the application is built, it can simply be run via:

```
./target/whatever-1.0.0-SNAPSHOT-runner
```

As an extra thing to try, you can set:

```properties
quarkus.langchain4j.llama3.chat-model.pre-load-in-native=true
```

in `application.properties` which results in the GGUF model being pre-loaded in the native image. This will make inference start immediately when the application is run, but does result in an increase in the binary's size.

sebastienblanc:
@geoand sorry for the newbie question but in dev mode , at runtime I'm getting ` java.lang.NoClassDefFoundError: jdk/incubator/vector/FloatVector` I'm on `24.ea.18-graal` and I start the app ` MAVEN_OPTS='--enable-preview' mvn quarkus:dev` , what am I doing wrong ?

geoand:
For the time being, the dev-mode experience isn't great as you need to set:

 `<jvm.args>--add-modules=jdk.incubator.vector --enable-preview</jvm.args>`

in the `properties` of `pom.xml`.

That's not all however, unless you are using Quarkus from `main` and have also set the `<forceC2>true</forceC2>` property,
your inference will be excrutiatingly slow... 

For those reasons, for the time being I don't recommend using Quarkus Dev with Jlama and Llama3.java

sebastienblanc:
Thx ! Sticking to `java -jar` then for now and thx for `--add-modules=jdk.incubator.vector` I was missing that too.

sebastienblanc:
And also, native will only work on `amd64` right ? Getting a ` Support for the Foreign Function and Memory API is currently available only on the AMD64 architecture.` when trying to compile to native on my `m3`

geoand:
I assume, but that's a question for the GraalVM folks :)

PR #1014, opened 2024-10-28
Release 0.21.0.CR1

geoand:


geoand:
@gastaldi I tried to do a release with the new process but it failed: https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/11550856264/job/32146686766

Any idea what the problem is? It seems to not like the fact that our samples have fixed versions - but that is done on purpose and those versions are updated automatically when the release is out

gastaldi:
That is really strange, it failed in the `release:prepare`, and that step hasn't changed from the original process. Do you get the same error if you run it locally?

geoand:
Indeed it does. Weird, we never encountered that in previous releases

geoand:
Aha, it seems that we used to do `-Darguments=-DperformRelease -DperformRelease` which meant that the samples were disabled.

I can just add that back to the two releases executions. I don't see how it could cause problems with the new release workflows, do you?

geoand:
Ah, I can't add those now, can I?

gastaldi:
Using `performRelease` in the `release:prepare` looks wrong IMHO. This is what it does: https://github.com/quarkiverse/.github/blob/main/.github/workflows/prepare-release.yml#L117

geoand:
Okay, I just need a way to deactivate the `samples` profile when a release is done and using:

```xml
      <activation>
        <property>
          <name>performRelease</name>
          <value>!true</value>
        </property>
      </activation>
```

was the easiest way.

What do you propose that be changed to?

gastaldi:
I can think of a few options: 

- Remove the `samples/pom.xml` and the `<module>` entry in the parent pom.xml;
- I have a https://github.com/quarkiverse/.github/pull/7 ready that adds a `-DprepareRelease=true` and which you could add as an `<activation>` too (keep this flag because of https://github.com/quarkiverse/.github/blob/main/.github/workflows/perform-release.yml#L105)
- Have your own `prepare-release.yml` locally and keep `perform-release.yml` untouched;

geoand:
Thanks!

PR #1023, opened 2024-10-30
Enable Tools to Define Execution Model

cescoffier:
This commit introduces support for tools to define their execution model with annotations such as @Blocking, @NonBlocking, and @RunOnVirtualThread, or to rely on the method signature to specify the execution model.

Execution constraints are enforced based on the caller thread; for instance, blocking calls are disallowed on the event loop.

For reactive AiService method (streaming), a detection is done at build time to determine if the execution can be done on the event loop or need to be switched to a worker thread. The detection checks the execution model of the various tool methods the AI service method can invoke.

- [x] Write documentation
- [ ] Update sample to show blocking tools when streaming is used
- [x] Make sure the duplicated context is propagated


andreadimaio:
I was curious to test these changes in my workspace, but when I use the `@RunOnVirtualThread` annotation on a `@Tool` method, I encounter an exception during application startup. However, this issue doesn't occur when I run your tests in the core module, where "virtual thread" methods work fine. It is possible that this is an issue on my side :thinking: 

Moving on, I would like to ask about a piece of  code used within the `StreamingChatModel` in watsonx.ai. This [code](https://github.com/quarkiverse/quarkus-langchain4j/blob/c458c3f0a412343b7631fece3d8d2de8563545e8/model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxChatModel.java#L114-L118) can be removed once this PR is merged, right?

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -760,6 +763,58 @@ public void markUsedOutputGuardRailsUnremovable(List<AiServicesMethodBuildItem>
...
+     * @param method the AI method
+     * @param tools the tools
+     */
+    public boolean detectAiServiceMethodThanNeedToBeDispatchedOnWorkerThread(
+            MethodInfo method,
+            List<String> associatedTools,
+            List<ToolMethodBuildItem> tools) {
+        boolean reactive = method.returnType().name().equals(DotName.createSimple(Uni.class.getName()))
+                || method.returnType().name().equals(DotName.createSimple(CompletionStage.class.getName()))
+                || method.returnType().name().equals(DotName.createSimple(Multi.class.getName()));
```
geoand, line 783: Minor nitpick: I am pretty sure these `DotName` classes already exist as constants in `DotNames`

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/items/ToolMethodBuildItem.java
```diff
@@ -0,0 +1,51 @@
+package io.quarkiverse.langchain4j.deployment.items;
+
+import org.jboss.jandex.MethodInfo;
+
+import io.quarkiverse.langchain4j.deployment.DotNames;
+import io.quarkiverse.langchain4j.runtime.tool.ToolMethodCreateInfo;
+import io.quarkus.builder.item.MultiBuildItem;
+
+public final class ToolMethodBuildItem extends MultiBuildItem {
```
geoand, line 9: Can we have some Javadoc about what the purpose of this build item is?

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionModelTest.java
```diff
@@ -0,0 +1,340 @@
...
+            }
+        });
+
+        Awaitility.await().until(() -> failure.get() != null);
+        assertThat(failure.get()).hasMessageContaining("Cannot execute blocking tools on event loop thread");
+    }
+
+    @Test
+    @ActivateRequestContext
+    @EnabledForJreRange(min = JRE.JAVA_21)
```
geoand, line 91: +1

geoand (commented):
Awesome stuff!

I only added a few minor comments, but we should also see what is causing the problem that @andreadimaio mentions

cescoffier:
It should have been a draft - sorry.

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
cescoffier: +1

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionModelTest.java
cescoffier: that's the beauty of our model - it does not have a dependency on the JDK 21 API :-)

andreadimaio:
> Can you perhaps add a test into this PR that surfaces the problem?

Yes, I could create something in the `integration-tests` folder. Let me know if you'd like me to proceed, or if you'd prefer to add the test at a later time.

geoand:
Up to @cescoffier

cescoffier:
@andreadimaio would be great! 

I know a few cases that should not have been working but were working that are now rejected (calling an imperative method from the event loop without the `@NonBlocking` annotation). Still, most of these cases are detected (in the case of streaming), and a switch to a worker thread is automatically done. Because of this, I don't think the `emitOn` you add is required anymore. But I would need to see the full stack trace. 

Now there is still a bit of work do do, as I "forgot" to implement context propagation. It should not take long, but my next coding session will be next week :-(

andreadimaio:
Here you can find the test cescoffier/quarkus-langchain4j#1.
When I run the test, I get the exception:
```
The @Blocking, @NonBlocking and @RunOnVirtualThread annotations may only be used on "entrypoint" methods (methods invoked by various frameworks in Quarkus)
Using the @Blocking, @NonBlocking and @RunOnVirtualThread annotations on methods that can only be invoked by application code is invalid
        at io.quarkus.deployment.execannotations.ExecutionModelAnnotationsProcessor.check(ExecutionModelAnnotationsProcessor.java:55)
        at java.base/java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:733)
        at io.quarkus.deployment.ExtensionLoader$3.execute(ExtensionLoader.java:856)
        at io.quarkus.builder.BuildContext.run(BuildContext.java:256)
        at org.jboss.threads.ContextHandler$1.runWith(ContextHandler.java:18)
        at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2516)
        at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2495)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1521)
        at java.base/java.lang.Thread.run(Thread.java:1583)
        at org.jboss.threads.JBossThread.run(JBossThread.java:483)
```
If I remove the annotations: `@Blocking`, `@NonBlocking` and `@RunOnVirtualThread`, the tests are green.

cescoffier:
Oh oh… that may require a change in Quarkus. I’m wondering why it works in
core…

cescoffier:
There is a package name check… but I should be able to emit the right build
item to handle this.

cescoffier:
@andreadimaio I think I fixed it.

andreadimaio:
> @andreadimaio I think I fixed it.

Yes, it works!

andreadimaio:
About this [code](https://github.com/quarkiverse/quarkus-langchain4j/blob/c458c3f0a412343b7631fece3d8d2de8563545e8/model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxChatModel.java#L114-L118)? I think it can be removed and managed with annotations.

andreadimaio:
Uhm... if I remove these lines from the `WatsonxChatModel` class:
```java
if (tools != null) {
   // Today Langchain4j doesn't allow to use the async operation with tools.
   // One idea might be to give to the developer the possibility to use the VirtualThread.
   mutiny.emitOn(Infrastructure.getDefaultWorkerPool());
}
``` 
And I try to call a tool method annotated with `@Blocking`:
```java
@ToolBox(Calculator.class)
public Multi<String> message(@UserMessage String message);
```
```java
@ApplicationScoped
public class Calculator {

    @Inject
    EmbeddingModel model;

    @Tool
    @Blocking
    public int sum(int a, int b) {
        model.embed("test");
        return a + b;
    }
}
```
I am getting `java.lang.IllegalStateException: Cannot execute blocking tools on event loop thread`.
Looking at the [code](https://github.com/cescoffier/quarkus-langchain4j/blob/8e71a617165434d65b7a1390d908f36ce264b807/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java#L779), the `TokenStream` is executed in a worker-pool, but the exception is raised in any case.

andreadimaio:
```java
case BLOCKING:
    if (io.vertx.core.Context.isOnEventLoopThread()) {
        throw new IllegalStateException("Cannot execute blocking tools on event loop thread");
    }
```

cescoffier:
It should have done the switch automatically. I will add a few tests.

andreadimaio:
Yes, if the switch is the line `Infrastructure.getDefaultWorkerPool().execute(tokenStream::start);` it is executed, but the exception is raised in any case

cescoffier:
Interesting, can I reproduce it with your test?

cescoffier:
Ok, I know why it happens - I need to think about the proper way to fix it.

andreadimaio:
> Interesting, can I reproduce it with your test?

Not yet, but I can try to add it.

> Ok, I know why it happens - I need to think about the proper way to fix it.

What?

cescoffier:
Basically, each item is emitted on the event loop. In my tests, I simplified this a bit too much by sending all the items synchronously on subscription (on the event loop).

geoand:
@cescoffier I'd like to include this in the next release, but if it still needs plenty of work, we can have in a later one

cescoffier:
I have a plan to fix the remaining issue. 

When do you want to cut the release?

geoand:
This week was the plan, but it's not set in stone

cescoffier:
I just got 1 free hour this afternoon, that should be enough

geoand:
💪

cescoffier:
@geoand how do we modify samples? I need to bump the version of quarkus langchain4j to demonstrate the feature. If I switch to 999-SNAPSHOT would it breaks the world?

geoand (approved):

geoand:
It seems like some of the tests fail

cescoffier:
@geoand @andreadimaio Seems like Watsonx is doing something that now blocks the event loop - I'm having a look.

cescoffier:
Hopefully, it should be better now. The issue was that the initial call was not on a Vert.x context (neither event loop nor worker), which broke the switch (as I had nowhere to switch to).

geoand:
All green!

andreadimaio:
I could do some testing in my local environment if you think it would be useful. But I can do that in a bit.

andreadimaio:
It works!! :rocket: 
@cescoffier I have opened a PR to merge some changes in watsonx, I have removed:
```java
if (tools != null) {
   // Today Langchain4j doesn't allow to use the async operation with tools.
   // One idea might be to give to the developer the possibility to use the VirtualThread.
   mutiny.emitOn(Infrastructure.getDefaultWorkerPool());
}
```
If you don't want to add this change here, I can open another PR.

PR #1027, opened 2024-10-31
Easy-rag: allow to specify minScore through configuration

edeandrea:
Allowing to specify `minScore` through configuration.

Also a little refactoring to pass the whole `EasyRagConfig` into the `EasyRetrievalAugmentor` rather than just the `maxResults`


jmartisk (approved):

edeandrea:
Eric Deandrea: i was debating whether to set a default of 0.0 (the current LangChain4j default), or to make it an OptionalDouble and only set it if the user wants it

Eric Deandrea: that way if upstream LangChain4j changed the default then we would just inherit their default, rather than always say 0.0

edeandrea:
Just rebased based on https://github.com/quarkiverse/quarkus-langchain4j/pull/1027#issuecomment-2449698235

PR #1031, opened 2024-11-01
Release 0.21.0.CR3

geoand:


geoand:
@gastaldi I want to perform some post release steps so I did https://github.com/quarkiverse/quarkus-langchain4j/blob/main/.github/workflows/release-perform.yml#L22-L43

However I don't see those post release steps having taken place. Any idea what I'm doing wrong?

geoand:
Actually, the whole release probably failed...

gastaldi:
You need to put in a different job, not steps

geoand:
Makes sense. 

That leaves the question of how I can see the action logs when something goes wrong :)

gastaldi:
I see it fails here because the YAML is invalid: https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/11626863204

Issue #1044, opened 2024-11-04
cancelling ollama progress is "hard"


maxandersen:
im not sure what is going on but when ollama is fetching model it updates *very* rapidly and cancelling with Ctrl-c takes a while / requires multiple attempts.

I *think* its because the progress is updated/refreshed so often that it result in delayed queuing of the ctrl-c over the many % updates.

Reason I think that is because when my ctrl-c is actually registered I get this output:

```
Ollama model pull: 2024-11-04 15:11:28,735 INFO  [io.qua.lan.dep.dev.DevServicesOllamaProcessor] (HttpClient-2-Worker-2) Progress: 27.01%
Tests paused
Press [e] to edit command line args (currently ''), [r] to resume testing, [o] Toggle test output, [:] for the terminal, [h] for more options>
2024-11-04 15:11:28,796 INFO  [io.qua.lan.dep.dev.DevServicesOllamaProcessor] (HttpClient-2-Worker-2) Progress: 27.05%

2024-11-04 15:11:28,858 INFO  [io.qua.lan.dep.dev.DevServicesOllamaProcessor] (HttpClient-2-Worker-2) Progress: 27.09%

2024-11-04 15:11:28,916 INFO  [io.qua.lan.dep.dev.DevServicesOllamaProcessor] (HttpClient-2-Worker-2) Progress: 27.13%

2024-11-04 15:11:28,975 INFO  [io.qua.lan.dep.dev.DevServicesOllamaProcessor] (HttpClient-2-Worker-2) Progress: 27.16%

2024-11-04 15:11:29,035 INFO  [io.qua.lan.dep.dev.DevServicesOllamaProcessor] (HttpClient-2-Worker-2) Progress: 27.21%

2024-11-04 15:11:29,095 INFO  [io.qua.lan.dep.dev.DevServicesOllamaProcessor] (HttpClient-2-Worker-2) Progress: 27.24%

2024-11-04 15:11:29,155 INFO  [io.qua.lan.dep.dev.DevServicesOllamaProcessor] (HttpClient-2-Worker-2) Progress: 27.28%

2024-11-04 15:11:29,215 INFO  [io.qua.lan.dep.dev.DevServicesOllamaProcessor] (HttpClient-2-Worker-2) Progress: 27.33%

2024-11-04 15:11:29,275 INFO  [io.qua.lan.dep.dev.DevServicesOllamaProcessor] (HttpClient-2-Worker-2) Progress: 27.38%

[...]

PR #1045, opened 2024-11-04
Improve model download progress logging

geoand:
This change includes two things:

* Ration progress log updates
* Include model name in progress log

Originally started because of #1044, but it also handles: #1046 
- Closes: #1044
- Closes: #1046


maxandersen:
That makes it a bit better - instead of a full screen of progress lines I just get 4 so I get a bit faster stop.

geoand:
I suppose we could make it every second instead of half second.

geoand:
PR updated

maxandersen (approved):
looks good

PR #1050, opened 2024-11-05
Allow using generic ModelAuthProvider with named models

### LINKED ISSUE #1038: Generic ModelAuthProviders should work with named models
While working on #733, I've found that a generic `ModelAuthProvider` such as the one provided by `quarkus-langchain4j-oidc-model-auth-provider` does not work with the `azure-openai` chat model provider, because the model name is set to `openai` and it leads to `ModelAuthProvider.resolve` finding no matching `ModelAuthProvider`.

The current `ModelAuthProvider.resolve` logic is as follows:
* If the model name is not null - then check only `ModelAuthProvider` annotated with a matching `@ModelName` qualifier
* Otherwise, if the model name is null, `ModelAuthProvider` without the qualifier is collected

The problem is that `ModelAuthProvider` provided by `quarkus-langchain4j-oidc-model-auth-provider` does not know in advance any model names, all it can do is to propagate either the incoming bearer access token or the current authorization code flow access token, representing the user authentication. 

I've done a workaround in #733, but it will most likely need to be reworked... 

I have the following suggestion:
* `ModelAuthProvider` has a new method added, say `modelNames`, default implementation will return an empty set
* `quarkus-langchain4j-oidc-model-auth-provider` runtime configuration will have a `supported-models` Set property which will be returned from the OIDC  `ModelAuthProvider` provided by this extension
* `ModelAuthProvider.resolve`: if the model name is not null, it will check `@ModelName` qualifiers, then, if no matching `ModelAuthProvider` is found, it will check `ModelAuthProvider` without qualifiers and will check `modelNames` method...

I'm not 100% happy with this suggestion, but I'm not sure what else can be done...

Or, perhaps the simplest solution can be the workaround I did in #733 ? If the model name is not null - check the one with the matching qualifier, if none is found, why don't pick up a generic one ? 
It seems the decision about which model a given `ModelAuthProvider` wants to work with should be coming from the `ModelAuthProvider` ?

@csotiriou @geoand, what do you think ? I'll be happy to work on the PR once the solution is agreed upon

Thanks

sberyozkin:
Fixes #1038

@csotiriou @geoand Please review

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/auth/AllModelAuthProvidersTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/auth/AllModelAuthProvidersTest.java
@@ -1,0 +1,34 @@
+package io.quarkiverse.langchain4j.test.auth;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkiverse.langchain4j.auth.ModelAuthProvider;
+import io.quarkus.test.QuarkusUnitTest;
+
+public class AllModelAuthProvidersTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest unitTest = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(GeminiModelAuthProvider.class, OpenaiModelAuthProvider.class, GlobalModelAuthProvider.class));
+
+    @Test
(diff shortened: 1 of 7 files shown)


sberyozkin:
Sorry, something does not look right

sberyozkin:
Forgot to push the updates

sberyozkin:
Adding one more test

sberyozkin:
Now all the variations should be covered:
* Global and named providers are registered: `null` model gets a global provider, named models - named providers
* Only global provider is registered: `null` and named models get the global provider
* Only named providers are registered: `null` model gets no provider

sberyozkin:
@geoand Let me try to add JavaDocs right now in this PR, and I can open an enhancement request to document how `ModelAuthProvider` and the OIDC implementation one in particular

sberyozkin:
@geoand Have a look please at JavaDocs, I'll give more examples later in the docs; please polish the text there to make it read well :-)

geoand (approved):

sberyozkin:
Thanks @geoand

csotiriou:
beat me to it, saw the conversation, lgtm

PR #1060, opened 2024-11-08
Provide an abstract output guardrails for json data extraction

mariofusco:
This is a first implementation of @lordofthejars 's idea discussed here https://github.com/quarkiverse/quarkus-langchain4j/discussions/1022

I believe this could be basically used for each `AiService` trying to extract a POJO from the user prompt like

```
@RegisterAiService
@Singleton
public interface CustomerExtractor {

    @UserMessage("Extract information about a customer from this text '{text}'. The response must contain only the JSON with customer's data and without any other sentence.")
    @OutputGuardrails(CustomerExtractionOutputGuardrail.class)
    Customer extractData(String text);
}
```

implementing the output guardrail simply extending the provided `AbstractJsonExtractorOutputGuardrail` as it follows

```
@ApplicationScoped
public class CustomerExtractionOutputGuardrail extends AbstractJsonExtractorOutputGuardrail {

    @Override
    protected Class<?> getOutputClass() {
        return Customer.class;
    }
}
```

/cc @geoand @cescoffier


lordofthejars:
Oh, wow, that is really nice. I have a list of guardrails that we could provide. I basically followed the LLM-guard project to check what guardrails they are offering and which of them we could implement.

mariofusco:
> Oh, wow, that is really nice. 

This is only the more straightforward things that it comes me to mind. We could add more features and implementations incrementally with other commits.

> I have a list of guardrails that we could provide. I basically followed the LLM-guard project to check what guardrails they are offering and which of them we could implement.

If you already have a list of guardrails that we could implement and provide out-of-the-box please feel free to share it, I'd be glad to collaborate on this.

lordofthejars:
Mario are you up next week for a quick meeting and I explain my idea?

geoand:
@gsmet when you have some time, I would love to chat with you about setting up a bot for this repo :)
A good starting point would be providing the awesome CI summary we have in the main Quarkus repo

gsmet:
I think we could install the Quarkus one here and only enable some of its features (that's something we can do).

geoand:
That would be awesome, thanks!

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/AbstractJsonExtractorOutputGuardrail.java
```diff
@@ -0,0 +1,38 @@
...
+import dev.langchain4j.data.message.AiMessage;
+import jakarta.inject.Inject;
+import org.jboss.logging.Logger;
+
+import static io.quarkiverse.langchain4j.guardrails.GuardrailsUtils.trimNonJson;
+import static io.quarkiverse.langchain4j.guardrails.GuardrailsUtils.validateJson;
+
+public abstract class AbstractJsonExtractorOutputGuardrail implements OutputGuardrail {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
```
cescoffier, line 13: I'm not sure we want to do that.
Mapper are huge, you probably want to reuse an existing one, maybe even the Quarkus one.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailsUtils.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+class GuardrailsUtils {
+    private GuardrailsUtils() {
+    }
+
+    private static class MapperHolder {
+        private static final ObjectMapper MAPPER = new ObjectMapper();
```
cescoffier, line 11: Another mapper? As I sid, we should avoid instantiating these.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailsUtils.java
```diff
@@ -0,0 +1,31 @@
...
+        int jsonEnd = llmResponse.lastIndexOf("}");
+        if (jsonStart >= 0 && jsonEnd >= 0 && jsonStart < jsonEnd) {
+            return llmResponse.substring(jsonStart, jsonEnd + 1);
+        }
+        return null;
+    }
+
+    static boolean validateJson(String json, Class<?> expectedOutputClass) {
+        try {
+            MapperHolder.MAPPER.readValue(json, expectedOutputClass);
```
cescoffier, line 25: You are immediately going into Databind. There are various issues:

1. I believe you need to verify it's valid JSON before mapping (see the example from the doc)
2. You pass a Class instead of a Type/TypeReference, so you won't handle generics and collections 
3. Mapping can be super expensive, and Quarkus is going to do it anyway latter anyway

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/AbstractJsonExtractorOutputGuardrail.java
lordofthejars: `@Inject ObjectMapper MAPPER`

cescoffier: Yes, that would be better (and it will also handle the modules the user has registered)

mariofusco: Sorry, this is a leftover of my original implementation, this mapper is totally useless now and I will remove it.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailsUtils.java
mariofusco: I will try to replace it with the one retrieved through Quarkus injection.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailsUtils.java
```diff
@@ -0,0 +1,31 @@
...
+class GuardrailsUtils {
+    private GuardrailsUtils() {
+    }
+
+    private static class MapperHolder {
+        private static final ObjectMapper MAPPER = new ObjectMapper();
+    }
+
+    static String trimNonJson(String llmResponse) {
+        int jsonStart = llmResponse.indexOf("{");
```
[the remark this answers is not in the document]
mariofusco: Good point, I will fix it.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailsUtils.java
mariofusco: > You are immediately going into Databind. There are various issues:
> 
>     1. I believe you need to verify it's valid JSON before mapping (see the example from the doc)

Mapping is a stronger constraint and then a super set of a valid JSON. I was thinking that in this way I'm doing both things in one shoot, but I'm no longer sure if this is a good idea. However if the JSON is valid (and I guess that in most cases it will be) I will have to also perform a mapping anyway.

>     2. You pass a Class instead of a Type/TypeReference, so you won't handle generics and collections

Good point, I will fix this.
 
>     3. Mapping can be super expensive, and Quarkus is going to do it anyway latter anyway

That's another good point, but with more complex solution. Ideally if the mapping is successful we may want to pass downstream the deserialized object instead of having to redo the mapping once again, but at the moment output guardrails don't allow to do this. I will check if I can change them a bit to make this possible.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailsUtils.java
cescoffier: You can use programmatic lookup and only create one if none are available.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailsUtils.java
cescoffier: Totally agree with your last point - if the mapping is successful we should pass that object as response.

mariofusco:
@cescoffier I should have addressed all your remarks. For now I'm keeping the 2 commits separated only to make it easier for you to review my latest change. In particular now an `OutputGuardrail` can not only rewrite the textual LLM response, but also provide an explicit object result for the whole AiService invocation. In this way the object eventually deserialized from the json response can be added to the result and returned as it is instead of requiring a second deserialization.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailsUtils.java
mariofusco: > yes that would be great specially in terms of performance

I already implemented this with my latest commit, give a look.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/AbstractJsonExtractorOutputGuardrail.java
```diff
@@ -0,0 +1,60 @@
...
+                return successWith(json, result);
+            }
+        }
+
+        return reprompt("Invalid JSON",
+                "Make sure you return a valid JSON object following "
+                        + "the specified format");
+    }
+
+    private Object deserialize(String llmResponse) {
```
cescoffier, line 48: Should we make this one protected?

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/AbstractJsonExtractorOutputGuardrail.java
```diff
@@ -0,0 +1,60 @@
...
+        return getOutputClass() != null ? guardrailsUtils.deserialize(llmResponse, getOutputClass())
+                : guardrailsUtils.deserialize(llmResponse, getOutputType());
+    }
+
+    protected Class<?> getOutputClass() {
+        return null;
+    }
+
+    protected TypeReference<?> getOutputType() {
```
cescoffier, line 57: Wondering if we can use `Type` to merge both the output class and output type?

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailsUtils.java
```diff
@@ -0,0 +1,47 @@
...
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+@ApplicationScoped
+class GuardrailsUtils {
```
cescoffier, line 11: Should it be named `JsonGuardrailsUtils`?

cescoffier (changes requested):
LGTM, I've just a few minor comments.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/AbstractJsonExtractorOutputGuardrail.java
mariofusco: Done.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/GuardrailsUtils.java
mariofusco: Done.

core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/AbstractJsonExtractorOutputGuardrail.java
mariofusco: I tried to, but `TypeReference` doesn't extends `Type`, but wraps it so I didn't find a reasonable way to do this.

mariofusco:
I implemented all the requested changes and squashed all commits.

cescoffier (approved):

cescoffier:
We should extend a bit the guardrail documentation to remove the json example and mention this one (with an example)

mariofusco:
> We should extend a bit the guardrail documentation to remove the json example and mention this one (with an example)

Agreed. I will rework docs as soon as this will be merged.

Issue #1059, opened 2024-11-07
Change the way the request models are collected


mkouba:
Currently, only "specific" injection points, such as `@Inject @ModelName ChatLanguageModel`, are considered. This means that programmatic lookup, such as `@Inject @Any Instance<ChatLanguageModel>` or `@Inject @All List<ChatLanguageModel>`, does not work unless a specific model is injected somewhere else.

See also zulip discussion: https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/named.20ai.20models.3F

geoand:
Thanks for opening the issue. I'll have a look

geoand:
I don't see a good way of handling these to be honest...

manovotn:
> I don't see a good way of handling these to be honest...

@geoand What's the reason behind only creating those beans if you find the injection point for them?

To implement this, we'd need to scan the config and pull out all of the models from there.
But looking at the code, there is quite some effort to *not* do that.

geoand:
The idea was to provide great error messages in order to guide users when configuring various models.

Your idea makes sense, but we'll have to move the configuration to build time (which I want to do anyway at some point)

manovotn:
> The idea was to provide great error messages in order to guide users when configuring various models.
> 
> Your idea makes sense, but we'll have to move the configuration to build time (which I want to do anyway at some point)

I see.
You can still keep that and do some best effort in registering all the other models for which you have no injection points.
Here's a draft of what I mean - https://github.com/quarkiverse/quarkus-langchain4j/pull/1061
I am not saying the PR is nice; just sharing it to show what I had in mind :)

PR #1061, opened 2024-11-08
Register all models from configuration as beans regardless of the existance of injection point

manovotn:
- Closes: #1059 

Just a draft to show how to enable injection even for models that do not have explicit injection points.

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/BeansProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/BeansProcessor.java
@@ -11,6 +11,7 @@ import static io.quarkiverse.langchain4j.deployment.LangChain4jDotNames.TOKEN_CO
 
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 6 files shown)


geoand:
Thanks I'll have a look!

geoand:
Ah, now I see what this does. I was under the impression that it was doing something more expansive, but what it does makes perfect sense

geoand:
@manovotn whenever you think the PR is ready, please mark it as such

manovotn:
I've expanded the tests and fix another bug that was causing `@Inject Instance<T>` to be treated as default model

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/BeansProcessor.java
```diff
@@ -298,6 +476,10 @@ private String determineModelName(InjectionPointInfo ip) {
                 return value;
             }
         }
+        // @Inject @Any Instance<Foo> should not be treated as default name
+        if (modelNameInstance == null && ip.isProgrammaticLookup()) {
```
geoand, line 480: Nice, I was not aware of `isProgrammaticLookup`!

PR #1068, opened 2024-11-11
Introduce a global temperature property

geoand:


jmartisk (changes requested):
The changes in `samples/review-triage/pom.xml` look accidental :)

geoand:
Good catch!

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/config/LangChain4jConfig.java
```diff
@@ -31,6 +32,11 @@ public interface LangChain4jConfig {
     @ConfigDocDefault("10s")
     Optional<Duration> timeout();
 
+    /**
+     * Global temperature for LLM APIs
```
jmartisk, line 36: Should we say what the allowed range is?

(And here we are, as I expressed my concern about creating similar abstractions - are we sure that all models and model providers will have the same allowed range for temperature?)

jmartisk: Then it's questionable whether to have this abstraction at all, but ok, a different number range isn't that hardcore, I could imagine worse :)

jmartisk (approved):

Issue #685, opened 2024-06-16
Random MismatchedInputException when using streaming with ollama


Malandril:
# Context
When using Ollama with an AiService with streaming enabled i randomly get an exception, which causes the `Multi` to never close.
```java
HTTP 400 Bad Request
        at io.quarkus.rest.client.reactive.jackson.runtime.serialisers.ClientJacksonMessageBodyReader.readFrom(ClientJacksonMessageBodyReader.java:59)
        at org.jboss.resteasy.reactive.client.impl.ClientReaderInterceptorContextImpl.proceed(ClientReaderInterceptorContextImpl.java:86)
......
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of `io.quarkiverse.langchain4j.ollama.ChatResponse` (although at least one Creator exists): no boolean/Boolean-argument constructor/factory method to deserialize from boolean value (true)
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 1]
        at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:63)
        at com.fasterxml.jackson.databind.DeserializationContext.reportInputMismatch(DeserializationContext.java:1754)
        at com.fasterxml.jackson.databind.DeserializationContext.handleMissingInstantiator(DeserializationContext.java:1379)
        at com.fasterxml.jackson.databind.deser.ValueInstantiator.createFromBoolean(ValueInstantiator.java:371)
        at com.fasterxml.jackson.databind.deser.std.StdValueInstantiator.createFromBoolean(StdValueInstantiator.java:505)
        at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.deserializeFromBoolean(BeanDeserializerBase.java:1651)
        at com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeOther(BeanDeserializer.java:206)
        at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:187)
        at com.fasterxml.jackson.databind.deser.DefaultDeserializationContext.readRootValue(DefaultDeserializationContext.java:342)
        at com.fasterxml.jackson.databind.ObjectReader._bindAndClose(ObjectReader.java:2125)
        at com.fasterxml.jackson.databind.ObjectReader.readValue(ObjectReader.java:1501)
[...]

PR #1070, opened 2024-11-12
Properly handle Ollama streaming

geoand:
We also catch `MismatchedInputException` when the
results are being streamed back as it indicates
we are still in an intermediate state

- Fixes: #685


cescoffier (approved):

benDeMtp:
@geoand , do you think this is possible to have this fix in the next release ?

geoand:
Yes, I will likely get a new release out this week

PR #1072, opened 2024-11-12
Fix MismatchedInputException in Ollama streaming

jpohlmeyer:
Handle MismatchedInputException in catch-clause.
It is either wrapped in ClientWebApplicationException with Status 400 or wrapped in ProcessingException depending on the Quarkus version.

See #685.

The diff display is a bit misleading, i only changed the catch and if clause + indentation.

Handling both cases in a single catch makes the if-clause a bit unwieldy, but two catch-clauses and extraction of the body to avoid code duplication was not really practical either.

### DIFF
--- a/model-providers/ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/OllamaRestApi.java
+++ b/model-providers/ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/OllamaRestApi.java
@@ -12,6 +12,7 @@ import java.util.List;
 import jakarta.ws.rs.Consumes;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
+import jakarta.ws.rs.ProcessingException;
 import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.WebApplicationException;
 import jakarta.ws.rs.client.ClientRequestContext;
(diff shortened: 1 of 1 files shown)


geoand (approved):
Thanks!

model-providers/ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/OllamaRestApi.java
```diff
@@ -85,40 +86,41 @@ class OllamaRestApiReaderInterceptor implements ReaderInterceptor {
...
-                                final var ctx = Vertx.currentContext();
-                                if (ctx == null) {
-                                    throw e;
-                                }
+            } catch (ClientWebApplicationException | ProcessingException e) {
+                if ((e instanceof ProcessingException pe && pe.getCause() instanceof MismatchedInputException) ||
+                        (e instanceof WebApplicationException wae
+                                && ((wae.getCause() instanceof JsonParseException && wae.getResponse().getStatus() == 200) ||
+                                        (wae.getCause() instanceof MismatchedInputException
+                                                && wae.getResponse().getStatus() == 400)))) {
```
geoand, line 94: Actually we should probably have a comment on why this is needed (i.e. the change in Quarkus).

jpohlmeyer: Yes, i will add a comment

geoand: 🙏🏽

geoand (approved):

PR #1073, opened 2024-11-12
Add a Secure SQL ChatBot demo

### LINKED ISSUE #609: Add secure-sql-chatbot demo
It can follow a pattern similar to #539, but will depend on `quarkus-langchain4j` upgrading to Quarkus 3.11 or later, which contains an initial WebSockets Next security integration done by @mkouba.

sberyozkin:
Fixes #609.

The flow will look like this:

* Step 1:

 Screenshot From 2024-11-12 12-39-59

* Step 2 (this page is only available to authenticated users, and the security identity is propagated to the WS connection once the user chooses a chatbot icon):

Screenshot From 2024-11-12 12-39-08

`ContentRetriever` can only be called if the user is authenticated:

Screenshot From 2024-11-12 13-10-25

I'd just like to look at some hardening updates at the Quarkus level with Martin and Michal to have some more security binding done, but overall it is looking not bad with the WS-Next security support.

Here are [application.properties](https://github.com/quarkiverse/quarkus-langchain4j/pull/1073/files#diff-3cb1fadce1645ccc4443e33c8a81848b386f43c3e49118d45687f1d9bfc90091).

It is only `HTTPS` and `WSS` secure protocols. 

I'll look at adding a custom WS Next `HttpUpgradeCheck`.

Perhaps an output Quardrail can be added as well

CC @geoand @cescoffier


jmartisk:
Gonna try to run it today

samples/secure-sql-chatbot/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+Then, simply run the project in Dev mode:
+
+```
+mvn quarkus:dev
+```
+
+## Using the example
+
+Open your browser and navigate to `https://localhost:8080`, accept the demo certificate.
```
jmartisk, line 35:
```suggestion
Open your browser and navigate to `https://localhost:8443`, accept the demo certificate.
```

geoand:
Very cool @sberyozkin!

samples/secure-sql-chatbot/README.md
```diff
@@ -0,0 +1,55 @@
...
+The chatbot is available at the secure `wss:` scheme.
+
+It uses a SQL database with information about movies with their
+basic metadata (the database is populated with data from
+`src/main/resources/data/movies.csv` at startup). When you ask a question, an
+LLM is used to generate SQL queries necessary for answering your question.
+Check the application's log, the SQL queries and the retrieved data will be
+printed there.
+
+The chatbot will refer to you by your name during the conversation.
```
jmartisk, line 48: The app just prepends the name before the bot's messages, which is not very elegant... I'm thinking how we could make this better. Maybe we could pass the name of the user as part of the system message?! And instead of the hardcoded `"Hello, " + idToken.getName() + ", I'm MovieMuse, how can I help you?"` at start, we could tell the LLM to greet the user...
I don't know - but it would probably make it look better.

sberyozkin: Hi @jmartisk 

>The app just prepends the name before the bot's messages, which is not very elegant... 

Yeah, I just wanted to confirm it was visible, without thinking how to make it look nicer

>Maybe we could pass the name of the user as part of the system message?! 

Sure, but I'm not sure how we can capture it in a system message ? 

>And instead of the hardcoded "Hello, " + idToken.getName() + ", I'm MovieMuse, how can I help you?" at start, we could tell the LLM to greet the user...

It would be great if we can get it do it, I'm just not sure how to teach LLM that it can get a user name from `idToken.getName()`, what do you think ?

sberyozkin: Hi @jmartisk , I guess it would not be much of an AI then, which would take a user name as a parameter and echo it back :-).
May be we get that user name used in a more interesting way.... The purpose of the demo is to show the authenticated user identity is visible to AI content retriever,  etc...

sberyozkin: @jmartisk  We already demo it in the secure fraud detection demo, the demo is kind of similar but the difference is that WebSockets are involved... But yeah, something more interesting with the user name should be shown, somehow it should impact the RAG process output, we can do something simple, just to highlight the point, if you have some ideas, let me know please

jmartisk: Yeah I don't have any better idea how to utilize the username other than for greeting the user :/

jmartisk: Unless we go really deep and make it into some kind of personalized adviser that knows some specific data about the user, which is out of scope here.

sberyozkin: Sure @jmartisk, I'll see at least if I can get the user identity impacting the RAG output, thanks

sberyozkin:
@geoand @jmartisk I've added a custom WebSockets ticket scheme just to show what users can do to make what are already secure WSS upgrade requests even more secure, and updated README with a a more complete list of security considerations.

Next, I'll to tune some of the RAG code to take the user identity into account and the demo should be ready by review once it is done

sberyozkin:
@jmartisk Jan, I did something similar to what is done in the secure fraud detection demo.

So, in addition to the Movies data support which was already available, a `MovieWatcher` DB is generated, where a registered movie watcher has a random preferred genre allocated.

Then, the Movie content retriever passes the genre preferred by the current user to the MovieSupport AI service which is asked to use this property to sort the movies...

Hopefully it is good enough to show how AI can use the user identity to support its work :-), have a look please

Thanks

sberyozkin:
Also added a Logout support, missing movie watcher exception support, similarly to how it is done in the secure fraud detection demo...

I'm happy enough now with what this demo can show...

samples/secure-sql-chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/MovieMuseSupport.java
```diff
@@ -0,0 +1,33 @@
...
+            Always include `movie_name` in the SELECT clause.
+
+            The user might have not provided the movie name exactly, in that case
+            try to correct it to the official movie name, or match it using a LIKE clause.
+
+            The table has the following columns:
+            {schemaStr}
+            
+            Sort the list of movies by the genre prefered by the movie watcher: 
+            {genre}
```
jmartisk, line 27:
```suggestion
            {preferredGenre}
```

This has to match the argument name, otherwise, this throws an exception that is swallowed and the bot just doesn't respond, without any error indication...

samples/secure-sql-chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/MovieWatcherRepository.java
```diff
@@ -0,0 +1,16 @@
...
+import io.quarkus.hibernate.orm.panache.PanacheRepository;
+import jakarta.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+public class MovieWatcherRepository implements PanacheRepository<MovieWatcher> {
+	
+	public String findPreferredGenre(String movieWatcherName, String movieWatcherEmail) {
+		MovieWatcher movieWatcher = find("name = ?1 and email = ?2", movieWatcherName, movieWatcherEmail).firstResult();
+        if (movieWatcher == null) {
+            throw new MissingMovieWatcherException();
```
jmartisk, line 12: The problem with this is that this exception is swallowed and the bot just stops responding without error indication... I think this will need some better handling for content retriever errors in langchain4j, but until we have a proper fix, could we employ some workaround to make sure that the user gets an error? At least print the exception into the log here

samples/secure-sql-chatbot/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+The setup is defined in the [Setup.java](./src/main/java/io/quarkiverse/langchain4j/samples/chatbot/Setup.java) class.
+
+The registered movie watchers stored in a PostgreSQL database. When running the demo in dev mode (recommended), the database is automatically created and populated.
+
+The genre preferred by the registered movie watcher is used by a Movie `ContentRetriever` to sort the list of movies.
+
+## Running the example
+
+Users must authenticate with Google before they can start working with the Movie Muse and register the OpenAI API key.
```
jmartisk, line 24: Please add some examples of suitable prompts that the users can try out
(for example, `Give me movies with a rating higher than 8` seems fine to me)

samples/secure-sql-chatbot/README.md
```diff
@@ -0,0 +1,100 @@
...
+mvn quarkus:dev -Dname="Firstname Familyname" -Demail=someuser@gmail.com
+```
+
+Note, you should use double quotes to register your Google account's full name.
+
+## Using the example
+
+Open your browser and navigate to `https://localhost:8443`, accept the demo certificate.
+Now, choose a `Login with Google` option.
+Once logged in, click the orange robot in the bottom right corner to open the chat window.
```
jmartisk, line 52: Please add some examples of suitable prompts that the users can try...
`Give me movies with a rating higher than 8` seems like a good example to me

samples/secure-sql-chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/MovieMuseSupport.java
```diff
@@ -0,0 +1,33 @@
...
+            The dialect is PostgreSQL and the relevant table is called 'movie'.
+            Always include `movie_name` in the SELECT clause.
+
+            The user might have not provided the movie name exactly, in that case
+            try to correct it to the official movie name, or match it using a LIKE clause.
+
+            The table has the following columns:
+            {schemaStr}
+            
+            Sort the list of movies by the genre prefered by the movie watcher:
```
jmartisk, line 26:
```suggestion
            Sort the list of movies by the genre preferred by the movie watcher: 
```

samples/secure-sql-chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/MovieWatcherRepository.java
jmartisk: Ok I dug into it a bit more, and I think a solid solution would be to update the `ChatBotWebSocket.onMessage` method to catch and log an exception that is thrown from the `bot.chat(message)` call.

jmartisk: Because if we don't catch the exception here, it's just swallowed and the UI will forever keep waiting for a response that never comes

sberyozkin: @jmartisk Jan, have a look please, I added 2 exception catch blocks

jmartisk (approved):
LGTM

samples/secure-sql-chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/MovieMuseSupport.java
gsmet: @jmartisk isn't ^ something that would need fixing?

jmartisk: @gsmet It is fixed in the current state, maybe you're looking at an outdated commit

sberyozkin:
@jmartisk Let me do a bit of formatting as I've reset the workspace and lost the formatting configuration, so a few sources have tabs impacting them. And a few typos in README

sberyozkin:
Should be better now...

samples/secure-sql-chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/MovieMuseSupport.java
gsmet: I was more thinking of this behavior: `otherwise, this throws an exception that is swallowed and the bot just doesn't respond, without any error indication...` but I think I misunderstood and it was all related to this PR not to a Quarkus LangChain4j misbehavior.

jmartisk: That we fixed by catching the exception in the `ChatbotWebSocket` class

PR #1092, opened 2024-11-21
Prevent Jlama inference to block vertx event loop

mariofusco:
@cescoffier @geoand Please review.


cescoffier (approved):
Thanks! Looks good!

geoand (commented):
I'm +0.5, so let's here from @cescoffier.

If we do agree, we'll need to do the same thing for Llama3 as well

mariofusco:
> > we'll need to do the same thing for Llama3 as well
> 
> Can you please take of that @mariofusco ?

Sure. Should I do this in this same PR or a different one?

mariofusco:
> Let's just do it now and get it over with 🙂

Done.

geoand (approved):
Thanks!

Issue #1075, opened 2024-11-14
Update to LangChain4j 0.36


geoand:
cc @jmartisk @edeandrea

edeandrea:
So by cc'ing us, is that more for "FYI", or are you trying to cattle-prod one of us to do the work? :D

jmartisk:
While we're at this, I'm willing to help with it (hopefully I'm not catching myself into a big trap here)

jmartisk:
We've disabled Dependabot for the `dev.langchain4j` groupId as it kinda never turns out to work without additional changes

geoand:
> So by cc'ing us, is that more for "FYI", or are you trying to cattle-prod one of us to do the work? :D

Trying? Demanding is more like it 😆

edeandrea:
> > So by cc'ing us, is that more for "FYI", or are you trying to cattle-prod one of us to do the work? :D
> 
> Trying? Demanding is more like it 😆

You forget, I don't work in engineering :D 

> 
> > it kinda never turns out to work without additional changes
> 
> I think it only worked once in all the updates we've done

Good to know!

geoand:
> You forget, I don't work in engineering :D

Do you want to keep it that way? 😆 😆 😆

geoand:
Have you guys by any chance started any work on this?

geoand:
If you want to start, it would be great as I have a conference tomorrow so I won't be able to do any work

jmartisk:
Ok I'll get to it tomorrow probably

geoand:
Thanks a lot!

edeandrea:
> Have you guys by any chance started any work on this?

I don't work in engineering :)

Issue #1095, opened 2024-11-22
Integrate Mistral moderation model


jmartisk:
This is now supported in lc4j upstream since 0.36, but we don't have it integrated yet

PR #1097, opened 2024-11-22
Upgrade to LangChain4j 0.36.2

jmartisk:
- Closes: #1075

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodCreateInfo.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodCreateInfo.java
@@ -1,7 +1,5 @@
 package io.quarkiverse.langchain4j.runtime.aiservice;
 
-import static org.apache.commons.lang3.StringUtils.EMPTY;
-
 import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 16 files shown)


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodCreateInfo.java
```diff
@@ -1,7 +1,5 @@
 package io.quarkiverse.langchain4j.runtime.aiservice;
 
-import static org.apache.commons.lang3.StringUtils.EMPTY;
```
jmartisk, line 3: This dependency has somehow disappeared, but I don't see why we should re-add it just for a constant of an empty string

model-providers/mistral/runtime/src/main/java/io/quarkiverse/langchain4j/mistralai/QuarkusMistralAiClient.java
```diff
@@ -116,6 +118,12 @@ public MistralAiEmbeddingResponse embedding(MistralAiEmbeddingRequest request) {
         return restApi.embedding(request, apiKey);
     }
 
+    // TODO: we don't provide support for MistralAiModerationModel yet
```
geoand, line 121: We don't?

geoand (commented):
Great stuff @jmartisk!

model-providers/mistral/runtime/src/main/java/io/quarkiverse/langchain4j/mistralai/QuarkusMistralAiClient.java
jmartisk: Well, users can instantiate it directly, but we don't have a declarative way via config properties for it yet. I plan to add that via https://github.com/quarkiverse/quarkus-langchain4j/issues/1095

geoand: Gotcha, thanks

PR #1100, opened 2024-11-22
Migrate to the JsonSchemaElement API

### LINKED ISSUE #1054: Migrate to the `JsonSchemaElement` API
When defining tools, use the new `JsonSchemaElement` API instead of `JsonSchemaProperty`.

See https://github.com/quarkiverse/quarkus-langchain4j/issues/1036#issuecomment-2457557006 for some details. This needs to be put on hold a bit until the next upstream LangChain4j release, as it contains some things that are needed.

edeandrea:
Migrate to the `JsonSchemaElement` API.

Closes #1054


edeandrea:
I'm pretty sure I got everything! Let's make sure CI passes

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/JsonSchemaElementObjectSubstitution.java
```diff
@@ -0,0 +1,86 @@
...
+        JsonEnumSchemaObjectSubstitution,
+        JsonIntegerSchemaObjectSubstitution,
+        JsonNumberSchemaObjectSubstitution,
+        JsonObjectSchemaObjectSubstitution,
+        JsonReferenceSchemaObjectSubstitution,
+        JsonStringSchemaObjectSubstitution {
+
+    // Using ConcurrentHashMap in case multiple threads are using this class at the same time
+    // Not sure if this will ever happen
+    private final Map<Class<?>, JsonSchemaElementObjectSubstitution> substitutions = new ConcurrentHashMap<>(8);
```
edeandrea, line 30: @geoand I'm not too familiar with substitutions/etc. I wasn't sure if there was any kind of a chance of multi-threading here, so I used a `ConcurrentHashMap`.

If there's no chance of multiple threads calling this at the same time I can switch it to a `HashMap`.

edeandrea:
Looks like I need to fix

https://github.com/quarkiverse/quarkus-langchain4j/blob/2103f0748aed983262a41dd08d1ea6c6e0aa06c3/model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/bean/TextChatMessage.java#L225-L232

andreadimaio:
Ops, I'm starting to work on the same topic. You can see my changes in my [repo](https://github.com/quarkiverse/quarkus-langchain4j/commit/cf642ad61b9b0895879e619ce93d21d49bbbdc6b).

edeandrea:
I'm sorry about crossing each other. There was already an issue I had created a few weeks ago (#1054) that I had assigned to myself. 

I started working on this a few weeks ago but then ran into a wall and needed to wait for a LangChain4j upgrade.

andreadimaio:
Yeah, I noticed it after I started. I wasn't paying attention 😄 .

edeandrea:
From what I can tell (only looking at this on my phone) we came up with similar solutions :)

Good sanity check!

geoand:
Very nice, thanks @edeandrea! I added a small comment on this

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/JsonSchemaElementObjectSubstitution.java
geoand: There is no need for this complexity. I prefer to use what @andreadimaio has done

edeandrea: I will fix this. One difference between what I and @andreadimaio did was that he removed all traces of `toolParameters` from existence within Quarkus LangChain4j.

Is that something we can/should do? Or leave it as deprecated for now? If we remove it then that is a breaking change to other people who may be using/extending from us.

geoand: Good point. Let's just deprecate for now

geoand: The general idea is that Quarkus needs to be able to take an object passed to a recorder and create bytecode instructions out of it. For POJOs (or stuff like records), it's straighforward (as it is for example when serializing to JSON). But for more complex types, an extension needs to essentially instruct Quarkus on how to serialize / deserialize these types

> I'm also not clear on what a substitution actually is.

It's a way to allow Quarkus serialize (and deserialize) a complex object where normal POJO rules don't apply. 

> Do I need an ObjectSubstitution class specific for the generic JsonSchemaElement interface? Or does having an ObjectSubstitution for every implementing class suffice?

The latter.

geoand (approved):
Thanks a lot!

geoand:
Once this is in, I'll release a `0.22.0.CR1`

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/JsonSchemaElementObjectSubstitution.java
andreadimaio: > I will fix this. One difference between what I and @andreadimaio did was that he removed all traces of `toolParameters` from existence within Quarkus LangChain4j.
> 
> Is that something we can/should do? Or leave it as deprecated for now? If we remove it then that is a breaking change to other people who may be using/extending from us.

I removed all traces of `toolParameters` because it is not possible to use both the `parameters` and `toolParameters` fields simultaneously.
```java
private ToolSpecification(Builder builder) {
        this.name = builder.name;
        this.description = builder.description;
        if (builder.parameters != null && builder.toolParameters != null) {
            throw new IllegalArgumentException("Both (new) JsonObjectSchema and (old) ToolParameters " +
                    "are used to specify tool parameters. Please use only (new) JsonObjectSchema.");
        }
        this.parameters = builder.parameters;
        this.toolParameters = builder.toolParameters;
    }
```
Since the `ToolProcessor` class now exclusively uses the parameters field, the `toolParameters` field will always be null. As a result, I decided to remove all traces of `toolParameters`.

Issue #1087, opened 2024-11-19
Enable resolution of AI services by EL name


aldettinger:
Enabling AI services bean resolution by EL name may improve user experience in Apache Camel.

Like we could define an AI service like below:
```
@RegisterAiService
@ApplicationScoped
@Named("customPojoExtractionService")
public interface CustomPojoExtractionService {...}
```
And then use it from a Camel route like that:
```
from("...").bean("customPojoExtractionService");
// Under the hood, camel uses jakarta.enterprise.inject.spi.BeanManager.getBeans("customPojoExtractionService");
```

One way to implement could be to copy the `@Named` annotation value when quarkus-langchain4j generates the AI service implementation.

Does it sound a valid idea to start working on a pr ?
I guess it is valid to assume that there will forever be only a single AI service implementation class per AI service interface ?
Maybe later on, it makes sense to also implement resolution by interface type ? and by @Identifier ?

geoand:
We generally try to avoid named beans as much as possible. 

In the specific case of AiService, why use the name of the bean and not the type?

aldettinger:
Along the years, it's common habit for camel users to reference beans by name and to a less extent by type. 
In the specific case of AiService, there is camel bug or maybe limitations preventing the resolution by interface.
This could be investigated for sure but that's actually another topic.

Back to resolution by name, if we consider it's a bad design in the specific case of AiService, then maybe we prefer to close this ticket ? What do you think ?

geoand:
If you want to add opt-in support for `@Named` like you propose, that's fine - we can have it

aldettinger:
It could be worth exploring, just need a bit more info about opt-in support.
How is opt-in usually implemented ? Maybe we have an example somewhere please ?

geoand:
I don't have any examples, but in this case we likely just need to read the annotation and pass the value to `ExtendedBeanConfigurator`.

https://github.com/quarkiverse/quarkus-langchain4j/blob/bc7d6c05f942f94de0cd052292b0e3d456ab4f16/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java#L469 is where the bean is built, so if you include (an optional) bean name in `DeclarativeAiServiceBuildItem`, then you can pass it to the bean creation process.

aldettinger:
Hi @geoand, I wonder if the fix is right actually. Looking at the resolved bean, the class is as below:

```
bean.getClass(): class 
[...]

PR #1110, opened 2024-11-26
Enable resolution of AI services by bean name

aldettinger:
- Closes: #1087


geoand (approved):
LGTM, thanks!

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -381,6 +382,12 @@ public void findDeclarativeServices(CombinedIndexBuildItem indexBuildItem,
 
             String imageModelName = chatModelName; // TODO: should we have a separate setting for this?
 
+            AnnotationInstance namedAnno = declarativeAiServiceClassInfo.annotation(NAMED);
+            Optional<String> beanName = Optional.empty();
```
cescoffier, line 386: Is this the @Named annotation? We had an issue with the nonbinding aspect of that annotation in the past.

geoand: It is yeah

aldettinger: Yes, this is **@ Named** annotation. So the non binding aspect is that this annotation can't be implemented well in Quarkus ? Does it apply in this specific case ? Or maybe, it's just not recommended to use it ?

geoand: It shouldn't be a problem in this case as it is entirely opt-in

PR #1123, opened 2024-12-01
Use user.home as default for models-path for jlama and llama3 docs

rsvoboda:
Use `user.home` as default for models-path for jlama and llama3 docs

Code is using `user.home`:
 * https://github.com/quarkiverse/quarkus-langchain4j/blob/main/model-providers/jlama/runtime/src/main/java/io/quarkiverse/langchain4j/jlama/JlamaModelRegistry.java#L20
 * https://github.com/quarkiverse/quarkus-langchain4j/blob/main/model-providers/llama3-java/runtime/src/main/java/io/quarkiverse/langchain4j/llama3/Llama3ModelRegistry.java#L41

But the config docs default was using `user.name`


geoand (approved):
Thanks!

Issue #1127, opened 2024-12-02
isAuthProviderAvailable throws instead of returning false


dastrobu:
With the current version `0.22.0` existing code broke with

`: jakarta.enterprise.inject.CreationException: Error creating synthetic bean [gqAPD1NX8Mh_nvvnixhSvId3n_A]: java.lang.IllegalArgumentException: A synthetic injection point was not declared for required type [jakarta.enterprise.inject.Instance<io.quarkiverse.langchain4j.auth.ModelAuthProvider> and qualifiers: [@jakarta.enterprise.inject.Default()]
`

Version `0.21.0` did not have any issue. 

The exception is thrown from 

```java
    private static <T> boolean isAuthProviderAvailable(SyntheticCreationalContext<T> context, String configName) {
        return context.getInjectedReference(MODEL_AUTH_PROVIDER_TYPE_LITERAL).isResolvable();
    }
```

where `getReference` is called: 

```java
    private <R> R getReference(Type requiredType, Annotation... qualifiers) {
        if (qualifiers.length == 0) {
            qualifiers = new Annotation[]{Literal.INSTANCE};
        }

        R ref = (R)this.injectedReferences.get(new TypeAndQualifiers(requiredType, qualifiers));
        if (ref == null) {
            throw new IllegalArgumentException("A synthetic injection point was not declared for required type [" + requiredType + " and qualifiers: " + Arrays.toString(qualifiers));
        } else {
            return ref;
        }
    }
```

I have set a qualifier in my case, but did not implement a `ModelAuthProvider`. Instead, the `api-key` is set. 

There might be something wrong with the implementation of `isAuthProviderAvailable`, as I'd expect it to return `false`.

geoand:
Thanks for reporting!

Can you please attach a sample project we can use to see the problem in action?

dastrobu:
@geoand I created a draft PR #1130 reproducing the issue in a test. 

The test is most likely not clean enough to be merged, but it is what I could boil down my project to. Hope that helps, let me know if I can assist further.

geoand:
Cool!

I'll have a look tomorrow

PR #1130, opened 2024-12-02
add test to reproduce isAuthProviderAvailable throws instead of retur…

dastrobu:
…ning false #1127


geoand:
Thanks for this.

Closing in favor of #1131

PR #1132, opened 2024-12-03
Support no authorization header being set when null is returned from ModelAuthProvider

### LINKED ISSUE #1128: Allow `NoAuthModelAuthProvider` for proxy/sidecar base authentication
I like the new `ModelAuthProvider` interface introduced in version `0.22.0`, allowing the implementation of custom logic for setting the Authorization header.

It would be nice if the interface allowed the implementation of a `NoAuthModelAuthProvider`:

```java
@ApplicationScoped
public class NoAuthModelAuthProvider implements ModelAuthProvider {
    @Override
        return null;
    }
}
```

This should simply prevent setting any authorization header. Currently, it fails with:

```
Caused by: java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "obj" is null
        at org.jboss.resteasy.reactive.common.headers.HeaderUtil.headerToString(HeaderUtil.java:48)
```

The use case is that I have a deployment where the generation of the authorization header is offloaded to a sidecar, which injects the authorization header.

Prior to version `0.22.0`, I was setting a dummy API key in the config, which is simply ignored. This is a workaround but not very nice.

```
quarkus.langchain4j.azure-openai.api-key=dummy-key
```

I can create a pull request.

dastrobu:
A typical use case is when the Authorization header is injected by a proxy (sidecar container).

Closes #1128

### DIFF
--- a/model-providers/ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/OllamaRestApi.java
+++ b/model-providers/ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/OllamaRestApi.java
@@ -170,10 +170,13 @@ public interface OllamaRestApi {
 
         @Override
         public void filter(ClientRequestContext context) {
-            context.getHeaders().putSingle(
-                    "Authorization",
-                    authorizer
-                            .getAuthorization(new AuthInputImpl(context.getMethod(), context.getUri(), context.getHeaders())));
+            String authValue = authorizer.getAuthorization(new AuthInputImpl(
+                    context.getMethod(),
+                    context.getUri(),
+                    context.getHeaders()));
+            if (authValue != null) {
+                context.getHeaders().putSingle("Authorization", authValue);
+            }
         }
 
         private record AuthInputImpl(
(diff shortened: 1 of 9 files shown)


model-providers/ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/OllamaRestApi.java
```diff
@@ -9,12 +9,7 @@
 import java.net.URI;
 import java.util.List;
 
-import jakarta.ws.rs.Consumes;
-import jakarta.ws.rs.POST;
-import jakarta.ws.rs.Path;
-import jakarta.ws.rs.ProcessingException;
-import jakarta.ws.rs.Produces;
-import jakarta.ws.rs.WebApplicationException;
+import jakarta.ws.rs.*;
```
geoand, line 12: Please don't use star imports. It's likely the IDE did this automatically (and it can be configured not to do so)

model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/common/OpenAiRestApi.java
```diff
@@ -14,24 +14,12 @@
...
-import jakarta.ws.rs.BeanParam;
-import jakarta.ws.rs.Consumes;
-import jakarta.ws.rs.HeaderParam;
-import jakarta.ws.rs.POST;
-import jakarta.ws.rs.Path;
-import jakarta.ws.rs.Priorities;
-import jakarta.ws.rs.Produces;
-import jakarta.ws.rs.QueryParam;
-import jakarta.ws.rs.WebApplicationException;
+import jakarta.ws.rs.*;
```
geoand, line 17: Same

geoand (commented):
Thanks!

I have added some comments about star imports (there are more instances than I flagged)

geoand (approved):
Thanks!

Issue #1129, opened 2024-12-02
Upgrading 0.21.0 -> 0.22.0 causes failures


edeandrea:
I have an app where it is [trying to upgrade](https://github.com/rh-rad-ai-roadshow/parasol-insurance/pull/175) from `0.21.0` to `0.22.0` and it fails with the following exception(s) (see https://github.com/rh-rad-ai-roadshow/parasol-insurance/actions/runs/12091388401/job/33800581449 & https://github.com/rh-rad-ai-roadshow/parasol-insurance/actions/runs/12091388401/job/33800581949

To reproduce, check out that PR and do `cd app && ./mvnw clean verify -Pollama` or `cd app && ./mvnw clean verify -Pjlama`

```
[ERROR] Tests run: 4, Failures: 0, Errors: 1, Skipped: 3, Time elapsed: 6.503 s <<< FAILURE! -- in org.parasol.resources.ClaimResourceTests
[ERROR] org.parasol.resources.ClaimResourceTests.getAllNoneFound -- Time elapsed: 0.004 s <<< ERROR!
java.lang.RuntimeException: 
java.lang.RuntimeException: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
        [error]: Build step io.quarkus.arc.deployment.SyntheticBeansProcessor#initRuntime threw an exception: java.lang.IllegalStateException: A synthetic bean with identifier OfFfdXYI2wr9Ymzsx9IXE32kX_M is already registered: SYNTHETIC bean [types=[dev.langchain4j.model.embedding.EmbeddingModel, java.lang.Object], qualifiers=[@Default, @Any], target=n/a]
        at io.quarkus.arc.processor.BeanDeployment.addSyntheticBean(BeanDeployment.java:1536)
        at io.quarkus.arc.processor.BeanDeployment$BeanRegistrationContextImpl.accept(BeanDeployment.java:1891)
        at io.quarkus.arc.processor.BeanDeployment$BeanRegistrationContextImpl.accept(BeanDeployment.java:1860)
        at io.quarkus.arc.processor.BeanConfigurator.done(BeanConfigurator.java:115)
        at io.quarkus.arc.deployment.SyntheticBeansProcessor.configureSyntheticBean(SyntheticBeansProcessor.java:101)
        at io.quarkus.arc.deployment.SyntheticBeansProcessor.initRuntime(SyntheticBeansProcessor.java:60)
[...]

PR #1135, opened 2024-12-03
Fix registration of embedding model for Easy RAG

jmartisk:
Fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/1129


jmartisk:
@manovotn may want to have a look too, as it is related to something that started failing after https://github.com/quarkiverse/quarkus-langchain4j/commit/b1c52cbbb667ed86fc31c7478c53015d6fd99b6f

This tries to avoid a duplicate `SelectedEmbeddingModelCandidateBuildItem` when Easy RAG is present and the user has chosen an explicit provider for the embedding model

edeandrea:
I'm building this locally now

edeandrea (approved):
I can confirm this indeed fixes #1129.

geoand (approved):

manovotn (approved):
> @manovotn may want to have a look too, as it is related to something that started failing after https://github.com/quarkiverse/quarkus-langchain4j/commit/b1c52cbbb667ed86fc31c7478c53015d6fd99b6f

Right, I see the bit that I missed there :+1:

PR #1138, opened 2024-12-05
Make response handling more generic. Resulting in generic cost handling.

alesj:


brunobat:
Is this still relevant?

PR #1150, opened 2024-12-11
Add support for structured output in Ollama

andreadimaio:
This PR enables the use of structured output with json_schema in Ollama.

My idea is to force the use of the structured output + json_schema every time an AIService returns a POJO and the developer doesn't fill the "format" property. In this second case, the value chosen by the developer will have the priority.

### DIFF
--- a/model-providers/ollama/deployment/src/test/java/io/quarkiverse/langchain4j/ollama/deployment/OllamaJsonOutputTest.java
+++ b/model-providers/ollama/deployment/src/test/java/io/quarkiverse/langchain4j/ollama/deployment/OllamaJsonOutputTest.java
@@ -1,0 +1,93 @@
+package io.quarkiverse.langchain4j.ollama.deployment;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.equalToJson;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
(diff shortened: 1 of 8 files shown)


geoand (approved):
Thanks a lot!

andreadimaio:
I realize that I can improve the `format` checks a bit at build time, but before I start adding those checks, let me know what you think.

Today, the format property can take the value `json` or a json object that must conform to a json-schema `{...}`.

Suppose we have an AIService that returns a `Person` object and the `format` property has the value

- `null` | `empty`: The json schema is automatically generated and used by quarkus.
- `json`: The structured output is not used and the string "You must..." is appended to the prompt.
- `{...}` (a manual json schema is provided. I don't know how often this can happen, but it could):
   - `json-schema` invalid: raises an exception.
   - `json-schema` matches the `Person` object: continue without exception.
   - `json-schema` doesn't match the `Person` object: throws an exception, or we have to use whatever the developer wrote?

What do you think about these checks.. do they make sense or I'll increase the lines of code without giving a value?

My doubt is what to do when the developer writes a correct json-schema in the `format` property?

geoand:
The way things are now with my initial commit, both the json_schema and the old `You must...` are captured at build time and available at runtime to the AiService implementation.

I believe that initially we should default to the latter for now, while allowing the users to opt-in to the new structured output (like the test I had added does). In the future we can flip that default. 
To be honest, I don't see a scenario where the user writes the schema themselves.

andreadimaio:
> To be honest, I don't see a scenario where the user writes the schema themselves.

I think so too :+1: . Let's see if some future issues show us if other developers think differently.

geoand:
We'll find out soon enough :)

geoand:
@dliubarskyi @mariofusco I have [started working](#1308 on moving to the upstream LangChain4j OllamaChatModel, but IIUC, this feature hasn't been implemented there, correct?

dliubarskyi:
@geoand structured outputs with JSON schema are supported in Ollama: https://github.com/langchain4j/langchain4j/blob/1a8a70e76118f7f501561c9b77212b35a77688ca/langchain4j-ollama/src/test/java/dev/langchain4j/model/ollama/OllamaAiServicesWithJsonSchemaIT.java#L19

geoand:
I will take a look and get back to you, thanks.

PR #1187, opened 2025-01-03
Some fixes to the build

jmartisk:


testing/scorer/scorer-strategies/semantic-similarity/pom.xml
```diff
@@ -17,12 +17,13 @@
...
             <groupId>dev.langchain4j</groupId>
             <artifactId>langchain4j-embeddings-bge-small-en-v15</artifactId>
+            <version>${langchain4j-embeddings.version}</version>
             <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>io.quarkiverse.langchain4j</groupId>
             <artifactId>quarkus-langchain4j-testing-scorer-core</artifactId>
-            <version>999-SNAPSHOT</version>
+            <version>${project.version}</version>
```
jmartisk, line 26: I assume this would be a problem when releasing and changing the `project.version` to something else than `999-SNAPSHOT`

testing/scorer/scorer-strategies/semantic-similarity/pom.xml
```diff
@@ -17,12 +17,13 @@
         <dependency>
             <groupId>dev.langchain4j</groupId>
             <artifactId>langchain4j-embeddings-bge-small-en-v15</artifactId>
+            <version>${langchain4j-embeddings.version}</version>
```
jmartisk, line 20: This is a problem when I bump `langchain4j.version` to a version different than `langchain4j-embeddings.version` (for development purposes) - because then the version of this artifact is taken from the langchain4j bom instead of the `langchain4j-embeddings.version` property, making it inconsistent and forcing me to compile the `embeddings` repo as well to be able to build

geoand (approved):

PR #1189, opened 2025-01-03
Replace the SQL chatbot Data Ingestion with Camel Route & fix the GPT…

zbendhiba:
… default model name in the doc


samples/sql-chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/CsvIngestor.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.apache.camel.builder.endpoint.EndpointRouteBuilder;
+import org.apache.camel.model.dataformat.BindyType;
+
+public class CsvIngestor extends EndpointRouteBuilder {
+    @Override
+    public void configure() throws Exception {
+        from(file("src/main/resources/data?fileName=movies.csv").noop(true))
+                .unmarshal().bindy(BindyType.Csv, Movie.class)
+                .to(jpa(Movie.class.getName()+"?entityType=java.util.List"));
```
geoand, line 11: I think comments for each one of the camel methods here as I assume most of our users won't know what this does (I certainly don't)

samples/sql-chatbot/pom.xml
```diff
@@ -15,9 +15,10 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
         <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
-        <quarkus.platform.version>3.15.2</quarkus.platform.version>
+        <quarkus.platform.version>3.17.5</quarkus.platform.version>
```
geoand, line 18: I prefer not use Quarlis LTS unless there is something the sample needs for the latest versions

geoand (commented):
Thanks!

I added a couple of comments

samples/sql-chatbot/src/main/java/io/quarkiverse/langchain4j/sample/chatbot/Movie.java
```diff
@@ -6,51 +6,63 @@
...
 @Table(name = "movie", schema = "public")
+@CsvRecord(separator = ",", skipFirstLine = true )
 public class Movie {
 
     @Id
     @GeneratedValue
     private int id;
 
     @Column(name = "index")
+    @DataField(pos = 1)
```
maxandersen, line 25: Does camel not support csv column names ? Seems redundant having to specify order here ?

zbendhiba: According to the [doc](https://camel.apache.org/components/4.8.x/dataformats/bindy-dataformat.html) : Camel Bindy doesn't. And it's not related to persistence so I don't know exactly if we can improve that

Issue #1220, opened 2025-01-17
embed-model for OpenAI can not have its own base-url / api key


pilhuhn:
I am trying to access 2 different remote OpenAI-API-compatible services for the chat and embed-models.
This fails, as the embedding model does not have its own base-url and api-key

Same applies to imageModel and  moderationModel.

I think having those overrides is important as the OpenAI-API is kind of becoming a default API for many LLM-servers (InstructLab , Ollama in OpenAI mode, ..)

The following patch illustrates a potential fix, that works on my laptop :)

```
---
Index: model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/EmbeddingModelConfig.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/EmbeddingModelConfig.java b/model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/EmbeddingModelConfig.java
--- a/model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/EmbeddingModelConfig.java	(revision a8864444916d915b178eda65ee7665934e0ac7e8)
+++ b/model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/config/EmbeddingModelConfig.java	(date 1737111305494)
@@ -31,4 +31,14 @@
      * A unique identifier representing your end-user, which can help OpenAI to monitor and detect abuse.
      */
     Optional<String> user();
+
+    /**
+     * Base URL override for the embedding model
+     */
+    Optional<String> baseUrl();
+
+    /**
+     * Api key override for the embedding model
+     */
+    Optional<String> apiKey();
 }
Index: model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/OpenAiRecorder.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
[...]

PR #1225, opened 2025-01-20
Clarify usage of multiple models in one process

pilhuhn:
Clarify a bit the usage of more than one model of the same kind of provider (e.g. chat and embedd), that sit on different URLs. Also mention the usage of `@ModelName`, which was not listed in the docs.

This is a result of #1220


geoand (approved):
Thanks!

Issue #1091, opened 2024-11-21
Simplify dev-mode experience for Jlama and Llama3.java


geoand:
See https://github.com/quarkusio/quarkus/pull/44305

geoand:
We need to wait for Quarkus 3.18 to implement this

PR #1235, opened 2025-01-22
Make Jlama and Llama3.java work properly in dev mode without any use configuration

geoand:
- Closes: #1091


geoand:
I'll do the same for Llama3.java and then take this out of draft

PR #1260, opened 2025-02-03
draft: add extended-span-chat-model-listener

flyinfish:
enabling adding more/custom attributes to span, namely  before span is closed

for `onRequest` this would already be possible just using another `ChatModelListener`. not so unfortunately for `onResponse` as span is ended in `SpanChatModelListener.onResponse()` and so additional attributes land on parent-span overwriting each other with every new request.

this change is just missing some `Processor` Magic producing a default-`ExtendedSpanChatModelListener` and some docs may-be.


geoand:
Interesting!

I will have a closer look in a few days as I am traveling

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/ExtendedSpanChatModelListener.java
```diff
@@ -0,0 +1,53 @@
...
+import dev.langchain4j.model.chat.listener.ChatModelRequest;
+import dev.langchain4j.model.chat.listener.ChatModelRequestContext;
+import dev.langchain4j.model.chat.listener.ChatModelResponse;
+import dev.langchain4j.model.chat.listener.ChatModelResponseContext;
+import io.opentelemetry.api.trace.Span;
+
+/** 
+ * extended ChatModelListener adding possibility to add custom attributes to span
+ */
+public interface ExtendedSpanChatModelListener {
```
geoand, line 13: I would probably call this `ChatModelSpanEnhancer`

flyinfish: if you like it i can rename and make it merge-ready

geoand: I want to take a closer look in a few days.

But the PR will also need to make sure that the beans are picked up (if they exist) and work regardless of whether such beans are present

geoand: I thought about this a little more and I like the approach a lot.
So let's make sure it works under all circumstances (ideally with some tests) and also document it.

flyinfish:
@geoand @brunobat
code should be ok like this!?, still missing docs but let's agree first if you like to have it merged

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/listeners/ListenersProcessorOnlySpanChatModelListenerTest.java
```diff
@@ -0,0 +1,88 @@
...
+import io.quarkus.maven.dependency.Dependency;
+import io.quarkus.test.QuarkusUnitTest;
+
+class ListenersProcessorOnlySpanChatModelListenerTest {
+    @RegisterExtension
+    static final QuarkusUnitTest unitTest = new QuarkusUnitTest()
+            .setArchiveProducer(
+                    () -> ShrinkWrap.create(JavaArchive.class)
+                            .addClasses(AiService.class, EchoChatLanguageModelSupplier.class))
+            .setForcedDependencies(List.of(Dependency.of("io.quarkus", "quarkus-opentelemetry", "3.15.2")));
```
flyinfish, line 38: @geoand 
`"3.15.2"` any idea where this is supposed to come from?
or is there a simpler solution to it adding otel?

geoand: I'll need to externalize the version, but just go ahead and copy it for now if you need to add a new test.

geoand:
How about using `@All ListChatModelSpanContributor>`

so we can support multiple contributors? That way we also won't need the default bean.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/SpanChatModelListener.java
```diff
@@ -78,6 +82,7 @@ public void onResponse(ChatModelResponseContext responseContext) {
                     span.setAttribute("gen_ai.client.estimated_cost", costEstimate.toString());
                 }
             }
+            chatModelSpanContributor.onResponse(responseContext, span);
```
brunobat, line 85: This is not a safe operation. We don't control the code being executed there, meaning it can throw an exception and the span will never close.
This should be wrapped on a try/catch...
The whole span processing should be on a try/catch

brunobat (changes requested):

brunobat:
By principal, we should own the list of attributes to be placed in a span, to make sure we have all the important attributes covered.
Custom ad-hock attribute names, out of conventions, will never be considered by APM tools and users will have to handle their schema themselves.
Saying this, if ppl really want the feature, we should allow it with a warning.

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/listeners/ListenersProcessorOnlySpanChatModelListenerTest.java
```diff
@@ -0,0 +1,125 @@
...
+        assertThat(actual).isEqualTo("test:echo");
+    }
+
+    @Test
+    @ActivateRequestContext
+    void shouldCreateSpanOnSuccess() {
+        aiService.test();
+
+        await().untilAsserted(() -> assertThat(exporter.getFinishedSpanItems()).hasSize(1));
+        assertThat(exporter.getFinishedSpanItems()).extracting(SpanData::getName)
```
flyinfish, line 90: @geoand this will get nasty
i have yet to understand how spans are built in test - do you wanna go this road?
or may be a more stable / Mockito approach

geoand: I guess for the purposes of this test, we can go with a mock approach (although I'm not a huge fan)

flyinfish: code should be ready, still missing version to be externalized
there is one "bug" i think i fixed that span was not ended in `onError`

flyinfish:
> By principal, we should own the list of attributes to be placed in a span, to make sure we have all the important attributes covered. Custom ad-hock attribute names, out of conventions, will never be considered by APM tools and users will have to handle their schema themselves. Saying this, if ppl really want the feature, we should allow it with a warning.

@brunobat this solution enables different attribute-sets for every "APM worth having" to be collected. they could be enabled by config

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/SpanChatModelListener.java
```diff
@@ -53,6 +58,7 @@ public void onRequest(ChatModelRequestContext requestContext) {
         var attributes = requestContext.attributes();
         attributes.put(OTEL_SCOPE_KEY_NAME, scope);
         attributes.put(OTEL_SPAN_KEY_NAME, span);
+        chatModelSpanContributors.forEach(c -> invokeSafeWithRecordException(span, () -> c.onRequest(requestContext, span)));
```
geoand, line 61: Let's please avoid lambdas in this code. We generally strive for this principle in code that is executed at runtime

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/SpanChatModelListener.java
```diff
@@ -78,6 +84,8 @@ public void onResponse(ChatModelResponseContext responseContext) {
                     span.setAttribute("gen_ai.client.estimated_cost", costEstimate.toString());
                 }
             }
+            chatModelSpanContributors
+                    .forEach(c -> invokeSafeWithRecordException(span, () -> c.onResponse(responseContext, span)));
```
geoand, line 88: Same here

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/SpanChatModelListener.java
```diff
@@ -92,6 +100,8 @@ public void onError(ChatModelErrorContext errorContext) {
         Span span = (Span) attributes.get(OTEL_SPAN_KEY_NAME);
         if (span != null) {
             span.recordException(errorContext.error());
+            chatModelSpanContributors.forEach(c -> invokeSafeWithRecordException(span, () -> c.onError(errorContext, span)));
```
geoand, line 103: And here

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/ChatModelSpanContributor.java
```diff
@@ -0,0 +1,50 @@
...
+import dev.langchain4j.model.chat.listener.ChatModelResponse;
+import dev.langchain4j.model.chat.listener.ChatModelResponseContext;
+import io.opentelemetry.api.trace.Span;
+
+/**
+ * Contributes custom attributes, events or other data to the spans created by {@link SpanChatModelListener}.
+ */
+public interface ChatModelSpanContributor {
+    /**
+     * allows custom data to be added to the span
```
geoand, line 15: allows -> Allows

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/ChatModelSpanContributor.java
```diff
@@ -0,0 +1,50 @@
...
+     * @param requestContext The request context. It contains the {@link ChatModelRequest} and attributes.
+     *        The attributes can be used to pass data between methods of this listener
+     *        or between multiple listeners.
+     * @param currentSpan Span opened by {@link SpanChatModelListener}.
+     */
+    default void onRequest(ChatModelRequestContext requestContext, Span currentSpan) {
+    }
+
+    /**
+     * allows custom data to be added to the span
```
geoand, line 26: Same here

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/ChatModelSpanContributor.java
```diff
@@ -0,0 +1,50 @@
...
+     *        It contains {@link ChatModelResponse}, corresponding {@link ChatModelRequest} and attributes.
+     *        The attributes can be used to pass data between methods of this listener
+     *        or between multiple listeners.
+     * @param currentSpan Span opened by {@link SpanChatModelListener}.
+     */
+    default void onResponse(ChatModelResponseContext responseContext, Span currentSpan) {
+    }
+
+    /**
+     * allows custom data to be added to the span
```
geoand, line 38: And here

geoand (commented):
Very nice! I only added a few very minor comments

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/SpanChatModelListener.java
flyinfish: np... - but how comes?

docs/modules/ROOT/pages/ai-services.adoc
```diff
@@ -32,9 +32,8 @@ Once registered, you can inject the _AI Service_ into your application:
 The beans created by `@RegisterAiService` are `@RequestScoped` by default. The reason for this is that it enables removing chat <<memory,memory>> objects.
 This is a good default when a service is used during when handling an HTTP request, but it's inappropriate in CLIs or in WebSockets (WebSocket support is expected to improve in the near future).
 For example when using a service in a CLI, it makes sense to have the service be `@ApplicationScoped` and the extension allows this simply if the service is annotated with `@ApplicationScoped`.
-====
 
-== AI method declaration
+=== AI method declaration
```
flyinfish, line 36: @geoand current doc broken here?

flyinfish:
hope i am done!?

docs/modules/ROOT/pages/ai-services.adoc
geoand: Looks like it!

PR #1276, opened 2025-02-11
Fix testing doc typo

anunnakian:
I've fixed a small typo related to sample files.

The correct way to declare a sample is to use `expected-output` not `expectedOutput` :   

```yaml
- name: Sample1
  parameters:
    - "parameter1"
  expected-output: "expected1"
  tags: ["tag1"]
```


geoand (approved):
Great catch, thanks!

cescoffier (approved):

PR #1280, opened 2025-02-13
Rewrite auditing to use CDI events

### LINKED ISSUE #1217: Auditing is clunky and needs to be revisited
I was starting to implement some of the [auditing features in Quarkus LC4j](https://docs.quarkiverse.io/quarkus-langchain4j/dev/ai-services.html#_auditing) but quickly realized that it is very clunky.

The `Audit` class, which is intended to be extended, is the class that contains all of the event receiver methods (`initialMessages`, `addRelevantDocument` , `addLLMToApplicationMessage`, etc).  This class is not a CDI bean. Instead, its created by the AuditService.

If I wanted to, say, store those audit events into a database (or publish them to a Kafka topic), I really can't because its not a CDI bean. I can't inject an `Emitter` or a panache repository.

Wouldn't it be better that the pojo which holds the event data be separate from the listener class which handles the events, that way the event handler could be `ApplicationScoped` and could receive generic events?

## Design idea
I was thinking of using CDI events instead. We could create `record`s for each of the various event types which would capture the necessary data for each kind of event. Then applications could register to observe whichever events they wanted.

@geoand / @cescoffier/ @jmartisk thoughts?

If we agree on the approach this is something I could take up, although it will be a bit before I could get to it. Probably not until February sometime.

edeandrea:
Re-write of the auditing stuff.

**NOTE:** I did **NOT** remove the current auditing stuff. I only introduced the new stuff. I figured we could let both exist for a few releases to give people a chance to move (if anyone is even using it?)

If you'd prefer I get rid of the existing stuff completely as part of this just let me know & I'll do that too. I also added some additional deprecations that were missed in #1219, so I'll take that out too if you'd prefer I nuke the existing auditing stuff.

Also, initially I had the `AuditSourceInfo` class I created as a record, but I wanted to add a unique identifier to each instance that was set upon creation, so that made me have to build it as a class instead of a record.

Closes #1217


edeandrea:
@geoand / @cescoffier FYI

edeandrea:
Also after this bakes a little bit here in Quarkus I think ideally it should be moved upstream to @langchain4j

edeandrea:
Looks like I have some build failures to fix. I'll work on those in the morning.

geoand:
Thanks for this!

CI doesn't seem to like it :)

edeandrea:
> Thanks for this!
> 
> CI doesn't seem to like it :)

Yeah I noticed last night. I probably missed something I'm sure - i didn't run ALL the tests locally. I'll fix it.

edeandrea:
@geoand what do you want me to do about the existing auditing stuff? I left it as-is, albeit with some additional deprecations that I missed in #1219.

Do you want to keep that stuff around for a few releases and have both in parallel for a few releases? Or do you want me to nuke the old stuff?

geoand:
Let's get rid of it

edeandrea:
> Let's get rid of it

I'll work on that and fixing what's broken today.

geoand:
This will likely need a rebase onto `main`

geoand:
When you're done, we'll need a rebase to get rid of the merge commits  and commits that are a result of merge conflicts

edeandrea:
> This will likely need a rebase onto `main`

Yeah I just noticed that. GitHub forced it when I pushed (I think we pushed at the same time).

I'm not sure how to get rid of the merge commits....

edeandrea:
i'll do a rebase from main now and see if that fixes things

edeandrea:
There that should do it

edeandrea:
Please take a good look at this - first time I've ever done anything with CDI events, so I'm not 100% sure what I did was the correct way to do things.

core/runtime/src/main/java/io/quarkiverse/langchain4j/audit/AuditSourceInfo.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+
+import io.quarkiverse.langchain4j.runtime.aiservice.AiServiceMethodCreateInfo;
+
+/**
+ * Contains information about the source of an audit event
+ */
+public final class AuditSourceInfo {
```
geoand, line 13: As this is something users are onlt going to be consuming, I'd rather it be an interface

edeandrea: Good thought.

edeandrea: done

core/runtime/src/main/java/io/quarkiverse/langchain4j/audit/AuditSourceInfoImpl.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+
+import io.quarkiverse.langchain4j.runtime.aiservice.AiServiceMethodCreateInfo;
+
+/**
+ * Contains information about the source of an audit event
+ */
+public final class AuditSourceInfoImpl implements AuditSourceInfo {
```
geoand, line 13: Does this have to be public?

edeandrea: I supposed I could move the implementation class elsewhere (to `io.quarkiverse.langchain4j.runtime.aiservice`)

geoand: Yeah, I would much prefer that

edeandrea:
The Millvus failure I don't think is at all related. Some docker pull error.

core/runtime/src/main/java/io/quarkiverse/langchain4j/audit/AuditSourceInfoImpl.java
geoand: I don't want the class to be visible

edeandrea: Let me know what you'd like me to do. We either leave it alone, or we move the class to the `io.quarkiverse.langchain4j.runtime.aiservice` package and remove `public` from it.

My $0.02 I think we leave it as is. I think it makes more sense to keep it packaged with the rest of the auditing stuff rather than moving some random auditing class into a package dealing with AiService stuff. But I'll do whatever you prefer.

geoand: Let's move it, I don't want it to be public

edeandrea: weird it took forever for your comments to show up for me

PR #1288, opened 2025-02-15
Update watsonx.ai module to Langchain4j 1.0.0-beta1

andreadimaio:
I implemented the new `chat` method that was added in the latest version of lanchain4j


geoand (approved):
Thanks a lot!

PR #1215, opened 2025-01-13
Draft: refactor ModelAuthProvider to async

flyinfish:
ok this is an absolute draft - but for my case/azure it seems to work

https://github.com/flyinfish/quarkus-langchain4j--samples-azure-openai/pull/1/files#diff-e491064562ea5e8da3dde65d50fbdd527d9e1a058fb4031c78fa85ea43294677R39

- Closes: #1214

PR #1291, opened 2025-02-16
Suspend Azure OpenAI request when ModelAuthProvider is called

### LINKED ISSUE #1214: ModelAuthProvider restricted to @Blocking
## ISSUE
actually i am not 100% shure if bug,issue or i just missed something...
i am using `ModelAuthProvider` as described in https://docs.quarkiverse.io/quarkus-langchain4j/dev/openai.html#_using_authprovider.

everything works fine except that it forces me to use `@Blocking` when using SSE

```
  @GET
  @Path("/stream")
  @Produces(MediaType.SERVER_SENT_EVENTS)
  //@Blocking
  public Multi<String> helloStream(@QueryParam("message") Optional<String> message) {
    return helloAI.chatStream(message.orElse("Hi"));
  }

@ApplicationScoped
public class AzureOpenAiTokenProvider implements ModelAuthProvider {
  ...

  @Override
    var token = tokenCache.getTokenSync(TOKEN_REQUEST_CONTEXT, false).getToken();
    return format("Bearer %s", token);
  }
}

: io.vertx.core.VertxException: Thread blocked
        at java.base/java.io.FileInputStream.readBytes(Native Method)
....
        at com.azure.identity.implementation.IdentityClientBase.getTokenFromAzureCLIAuthentication(IdentityClientBase.java:681)
        at com.azure.identity.implementation.IdentitySyncClient.authenticateWithAzureCli(IdentitySyncClient.java:393)
        at com.azure.identity.AzureCliCredential.getTokenSync(AzureCliCredential.java:97)
        at com.azure.identity.ChainedTokenCredential.getTokenSync(ChainedTokenCredential.java:145)
        at com.azure.core.implementation.AccessTokenCache.lambda$new$2(AccessTokenCache.java:65)
        at com.azure.core.implementation.AccessTokenCache.lambda$retrieveTokenSync$11(AccessTokenCache.java:230)
        at com.azure.core.implementation.AccessTokenCache.getTokenSync(AccessTokenCache.java:93)
        at io.quarkiverse.langchain4j.sample.hello.AzureOpenAiTokenProvider.getAuthorization(AzureOpenAiTokenProvider.java:30)
```
## Expected Behaviour
wouldn't it be nice to use non-blocking auth which at least azure would provide
## Workaround
using blocking communication for whole llm-call
## Reproduce
```
git clone https://github.com/flyinfish/quarkus-langchain4j--samples-azure-openai.git
cd hellomodelauthprovider/
quarkus dev
http://localhost:8080/hello/stream
```

sberyozkin:
Fixes #1214.

This PR replaces #1215 - please close #1215 if this PR is approved.

I've confirmed it works as expected with the secure-poem-multiple-models demo


model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/common/OpenAiRestApi.java
```diff
@@ -184,14 +185,36 @@ public boolean test(SseEvent<String> event) {
...
+                            setAuthorization(requestContext);
+                            requestContext.resume();
+                        } catch (Exception e) {
+                            requestContext.resume(e);
+                        }
+                    }
+                });
+            } else {
+                setAuthorization(requestContext);
+            }
```
geoand, line 213: I would prefer we do something like:

```java
            Context context = vertx.getOrCreateContext();
            return new Executor() {
                @Override
                public void execute(Runnable command) {
                    context.runOnContext(v -> command.run());
                }
            };
```

where `vertx` can be obtained from CDI.

sberyozkin: Hi @geoand I can try a bit later as I'll need to verify it against Azure OpenAI.

In meantime, out of curiosity, is it a better pattern because of the way Quarkus LangChain4j deals with the blocking LangChain4j calls ? I'm also not sure if a Vertx instance is available at the filter initialization time, if not it would have done at the request time...

geoand: 👋🏽 

Vertx should be available at that point (and there are multiple ways to obtain it).

The pattern is the same one we use elsewhere in Quarkus code, so it's best to use it here too.

sberyozkin:
Hi @geoand How is it looking now ? I still added a vertx null check, just in case

geoand:
`OpenAIRestAPIFilterTest` will need to be updated to be a proper Quarkus test

sberyozkin:
@geoand I looked at it and actually tried to move that test to `openai-common-deployment`, but then found out `asure-openai-deployment` already has a test, and I thought, given that currently this filter run is only relevant in scope of Azure OpenAI, that updating Azure OpenAI test would be better, and it can be pushed down to `openai-common-deployment` if OpenAI starts supporting tokens directly.

And I removed the unit test... What is not tested as a result is that if `ModelAuthProvider` returns null then the filter sets nothing either, but I think this is actually an exceptional case that may have to be treated with another dedicated PR, since the `ModelAuthProvider` is used when no api keys are available, and if no token is also available then may be a failure should be reported

geoand (approved):

PR #1300, opened 2025-02-19
Update vertex-ai-gemini to depend on gemini-common

sberyozkin:
@geoand @lordofthejars Have a look please


geoand (approved):
Thanks!

lordofthejars (approved):

geoand:
@lordofthejars if you can add some docs at some point, that would be great as I am eager to try out this new integration.

Tanks!

lordofthejars:
Yes I will do it, also I think I updated the yml file too so it is shown in code.quarkus.io, but if you can take a look I did correctly I think it would be better.

PR #1320, opened 2025-02-25
Fix AI services as tools lookup using @ToolBox annotation

mariofusco:
Fixes bug discussed here https://github.com/mariofusco/quarkus-agentic-ai/commit/0516a93f1313ee78a1609c5d96b476d5ebba8991#commitcomment-152919896

Adding the AI service tool among the unremoveable beans using its class name is wrong because that actual bean that shouldn't be removed is the generated one that implements the AI service but has a different class name. Conversely, registering its bean type using its class fixes the problem.

/cc @geoand

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -1120,8 +1120,10 @@ public class AiServicesProcessor {
                                 ignoredPredicates,
                                 tools, toolQualifierProviderItems);
                         if (!methodCreateInfo.getToolClassInfo().isEmpty()) {
-                            unremovableBeanProducer.produce(UnremovableBeanBuildItem
-                                    .beanClassNames(methodCreateInfo.getToolClassInfo().keySet().toArray(EMPTY_STRING_ARRAY)));
+                            methodCreateInfo.getToolClassInfo().keySet().stream()
+                                    .map(DotName::createSimple)
+                                    .map(UnremovableBeanBuildItem::beanTypes)
+                                    .forEach(unremovableBeanProducer::produce);
                         }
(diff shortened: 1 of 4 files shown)


core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -1120,8 +1120,16 @@ public void handleAiServices(
...
-                            unremovableBeanProducer.produce(UnremovableBeanBuildItem
-                                    .beanClassNames(methodCreateInfo.getToolClassInfo().keySet().toArray(EMPTY_STRING_ARRAY)));
+                            methodCreateInfo.getToolClassInfo().keySet().stream()
+                                    .map(toolClassName -> {
+                                        try {
+                                            return Thread.currentThread().getContextClassLoader().loadClass(toolClassName);
+                                        } catch (ClassNotFoundException e) {
+                                            throw new IllegalStateException(e);
+                                        }
+                                    })
```
geoand, line 1130: Do we really need to load the class?

mariofusco: I also don't like this. Alternatively I'd need the `DotName` of the class. I tried to carry it inside the  `AiServiceMethodCreateInfo` instead of the `String` as it does now, only to realize that I cannot do that because that class is in the runtime which doesn't depend on jandex. Is there a way to recreate the `DotName` of the class here starting from the `String` class name?

geoand: `DotName.createSimple(toolClassName);` is what you are looking for

mariofusco: Thanks for suggestion, fixed.

geoand (approved):
Thanks a lot for tracking this down!

PR #1334, opened 2025-03-02
Support EmbeddingModel for VertexAi Gemini

sberyozkin:
Aligning AiGemini with VertexAiGemini to support EmbeddingModel, based on the work of @lordofthejars.

I think we'll have more opportunities to push to the common gemini module later :-), since both providers are nearly identical, with some configuration differences.

@lordofthejars @geoand, please review


sberyozkin:
By the way, I know it is a bit of hassle, but I'd like to propose that any code that impacts AI Gemini is synced with the Vertex AI Gemini so that we can keep them in sync and eventually avoid all the duplication. Otherwise I won't be able to keep up with syncing Vertex AI Gemini and eventually the Vertex AI provider option will become stale.

FYI, I'm going to work on adding ChatModelListener support to both providers.

geoand:
I think that makes sense.
We definitely should try to keep the modules in sync.

geoand (approved):
Thanks a lot @sberyozkin !

sberyozkin:
@geoand Can you please merge ? Once this one and the upcoming ChatModelListener for Gemini PR is in, I'll look at pushing down some of the common chat/embedding model code as well...

PR #1335, opened 2025-03-02
Update watsonx.ai test classes

andreadimaio:
With the beta2 release of langchain4j the `generate` method in the `ChatLanguageModel` and `StreamingChatLanguageModel` interfaces will no longer be available. I have updated the tests in the watsonx.ai module to be ready for the new release.

### DIFF
--- a/model-providers/watsonx/deployment/src/test/java/io/quarkiverse/langchain4j/watsonx/deployment/AiChatCacheTokenTest.java
+++ b/model-providers/watsonx/deployment/src/test/java/io/quarkiverse/langchain4j/watsonx/deployment/AiChatCacheTokenTest.java
@@ -1,8 +1,11 @@
 package io.quarkiverse.langchain4j.watsonx.deployment;
 
-import static io.quarkiverse.langchain4j.watsonx.deployment.WireMockUtil.streamingResponseHandler;
+import static org.awaitility.Awaitility.await;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.fail;
 
+import java.time.Duration;
 import java.time.Instant;
 import java.util.Date;
 import java.util.Map;
@@ -19,9 +22,10 @@ import org.junit.jupiter.api.extension.RegisterExtension;
 
 import com.github.tomakehurst.wiremock.stubbing.Scenario;
 
-import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.model.chat.ChatLanguageModel;
 import dev.langchain4j.model.chat.StreamingChatLanguageModel;
+import dev.langchain4j.model.chat.response.ChatResponse;
+import io.quarkiverse.langchain4j.watsonx.deployment.WireMockUtil.WatsonxBuilder;
 import io.quarkus.test.QuarkusUnitTest;
 
 public class AiChatCacheTokenTest extends WireMockAbstract {
@@ -55,6 +59,11 @@ public class AiChatCacheTokenTest extends WireMockAbstract {
     @Inject
     StreamingChatLanguageModel streamingChatModel;
 
+    @Override
+    void handlerBeforeEach() throws Exception {
+        Thread.sleep(cacheTimeout);
+    }
+
     @Test
     void try_token_cache() throws InterruptedException {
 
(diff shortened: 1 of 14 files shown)


model-providers/watsonx/deployment/src/test/java/io/quarkiverse/langchain4j/watsonx/deployment/AiChatCacheTokenTest.java
```diff
@@ -43,9 +47,9 @@ public class AiChatCacheTokenTest extends WireMockAbstract {
 
     @RegisterExtension
     static QuarkusUnitTest unitTest = new QuarkusUnitTest()
-            .overrideRuntimeConfigKey("quarkus.langchain4j.watsonx.base-url", WireMockUtil.URL_WATSONX_SERVER)
-            .overrideRuntimeConfigKey("quarkus.langchain4j.watsonx.iam.base-url", WireMockUtil.URL_IAM_SERVER)
-            .overrideRuntimeConfigKey("quarkus.langchain4j.watsonx.api-key", WireMockUtil.API_KEY)
+            .overrideConfigKey("quarkus.langchain4j.watsonx.base-url", WireMockUtil.URL_WATSONX_SERVER)
+            .overrideConfigKey("quarkus.langchain4j.watsonx.iam.base-url", WireMockUtil.URL_IAM_SERVER)
+            .overrideConfigKey("quarkus.langchain4j.watsonx.api-key", WireMockUtil.API_KEY)
```
geoand, line 52: Why this change? Same for the other instances where the method was changed.

model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/runtime/config/LangChain4jWatsonxFixedRuntimeConfig.java
```diff
@@ -29,9 +30,31 @@ public interface LangChain4jWatsonxFixedRuntimeConfig {
...
+         * Base URL for prompts and agent tools in the watsonx.ai API.
+         * <p>
+         * If empty, this URL will be automatically calculated based on the {@code base-url}.
+         */
+        Optional<String> wxBaseUrl();
+
+        /**
+         * IBM Cloud API key.
+         */
+        Optional<String> apiKey();
```
geoand, line 52: This configuration should probably **not**  be runtime fixed, because that means that a user can't change it at runtime

geoand (commented):
Thanks for this!

I have doubts about the config

model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/runtime/config/LangChain4jWatsonxFixedRuntimeConfig.java
andreadimaio: Are you suggesting to use `ConfigPhase.BUILD_TIME`?

model-providers/watsonx/deployment/src/test/java/io/quarkiverse/langchain4j/watsonx/deployment/AiChatCacheTokenTest.java
andreadimaio: I made these changes because I am working on some new features in watsonx.ai that are still in beta for now, so I did not open a PR. Looking at the changes I am working on, I only need to check the baseUrl property during the build phase to perform some checks, the rest of the parameters can be moved back to runtime.

model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/runtime/config/LangChain4jWatsonxFixedRuntimeConfig.java
geoand: I'm suggesting RUNTIME config.

andreadimaio: Could you please take a look at this [link](https://github.com/quarkiverse/quarkus-langchain4j/commit/e2a0271be793e846753568aec0faa71f06dbfe90#diff-ace47a21ef0a381dd3789481f0affd2a6f089fd8d08425c32bbaea154d5296bcR87)? If the `baseUrl` is a RUNTIME config I have to move this check into a record, is it correct?

andreadimaio: > because that means that a user can't change it at runtime

Another curiosity, why should the user change the baseUrl property at runtime? Once specified, could it change while the application is being used?

geoand: If you make it fixed, then you are locking users into using that target. For example, it seems like the URLs in your commit seem to be geographic specific, which means that you need to rebuild the entire application if you want a different region.

For API keys, you absolutely can't have it fixed at build time.

andreadimaio: Yes, I was thinking about it and I was coming to your conclusion. This means that the check I'm doing needs to be moved to runtime, thanks! :)

geoand: 👍🏽

geoand (approved):
Thanks!

PR #1337, opened 2025-03-02
Response type support for Vertex AI Gemini

sberyozkin:
I just copied some code from #1306 to keep it in sync, so that I can easily update both providers to handle ChatModelListener


sberyozkin:
I hope, eventually, all of this code will be in gemini-common, users would only choose how to deal with Gemini (with devcentric API keys or scale it up in vertex ai with regions, locations, tokens)

geoand (approved):
Thanks!

PR #1341, opened 2025-03-04
Create common Gemini ChatModel and EmbeddingModel code

sberyozkin:
This PR attempts to minimize the duplication between AiGemini and VertexAiGemini by pushing the common `ChatModel` and `EmbeddingModel` code down to the `gemini-common` module.

Some duplication will still remain and can be minimized during some follow-up iterations but, hopefully, with the `ChatModel` and `EmbeddingModel` code becoming shared, it will be much easier to deal with fixing the model or content related issues for both providers at the same time.

CC @geoand @lordofthejars


geoand (approved):
Thanks!

Issue #1342, opened 2025-03-04
Problem with some Ai Gemini tool parameters


sberyozkin:
I'm testing using access tokens (as opposed to keys) with  Ai Gemini, and noticed it fails with the invalid  argument error when one of the complex tool parameters has an empty properties object, for example

```
{"name":"updateEvent","description":"Update or move event","parameters":{"type":"object","properties":{"calendarId":{"type":"string"},"eventId":{"type":"string"},"event":{"type":"object","properties":{"description":{"type":"string"},

"end":{"type":"object","properties":{"date":{"type":"string"},

"dateTime":{"type":"object","properties":{},"required":[]},

...
```

here, `dateTime` is `ZonedDateTime` in the Java tool method, and the error is returned

```
GenerateContentRequest.tools[0].function_declarations[2].parameters.properties[event].properties[end].properties[dateTime].properties: should be non-empty for OBJECT type\n* GenerateContentRequest.tools[0].function_declarations[2].parameters.properties[event].properties[start].properties[dateTime].properties: should be non-empty for OBJECT type\n",
```

This works for VertexAiGemini but looks like AiGemini has a stricter tool argument validation.

@geoand I recall you were talking about problems for the LocalDate, but I don't remember in what context. I wonder if it is the same type of issue

geoand:
Is

```
"dateTime":{"type":"object","properties":{},"required":[]},
```

what we are sending?

sberyozkin:
Yes. With Vertex AI Gemini, the actual tool invocation succeeds, the agent can update the calendar correctly, despite the fact that the tool specification for the complex ZonedDateTime parameter is not complete.
But AiGemini is stricter so it wants individual `dateTime` object properties listed

sberyozkin:
The transformation happens [here](https://github.com/quarkiverse/quarkus-langchain4j/blob/main/model-providers/google/gemini/gemini-common/runtime/src/main/java/io/quarkiverse/langchain4j/gemini/common/ContentMapper.java#L176). Looks like `ToolSpecification` does not give a correct `JsonSchemaElement` for the complex parameter like `ZoneDateTime`

geoand:
Thanks. We'll have to fix that.

geoand:
I'll take a look tomorrow hopefully

geoand:
I am sure I have a Vertex AI Gemini access, I'll have to check.

sberyozkin:
@geoand Only AI Gemini is enough

[...]

PR #1343, opened 2025-03-04
Support accessing AiGemini with tokens

### LINKED ISSUE #1301: Allow access to AI Gemini with access tokens
See https://ai.google.dev/gemini-api/docs/oauth

I can have a look a bit later. Looks like the vertex-ai-gemini client request filter (which supplies either the default Google credential or ModelAuthProvider response as  a token) should be re-used between vertex-ai-gemini and ai-gemini and run unless an API key is available

sberyozkin:
Fixes #1301.

This PR:

* Moves the VertexAiGemini `TokenFilter` to `gemini-common`, renames to `ModelAuthProviderFilter`, uses the same approach as the Azure OpenAI filter does.
* Updated AiGemini `ChatModel` and `EmbeddingModel` to register TokenFilter if the api key is not available, after confirming earlier that either api key or `ModelAuthProvider` is available
* Added tests, the AiGemini smoke tests were depending on the default dummy key property, so I fixed that too

I've also confirmed with my VertexAiGemini demo, that with this PR, dropping the `ai.gemini.api-key` configuration and adding an OIDC ModelAuthProvider works, finding along the way a problem related to the tool parameters (#1342) which is not related to this PR though


geoand:
Great works @sberyozkin!

I will definitely try this when I'm back

model-providers/google/gemini/gemini-common/runtime/pom.xml
```diff
@@ -22,6 +22,11 @@
             <artifactId>quarkus-langchain4j-core</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.auth</groupId>
+            <artifactId>google-auth-library-oauth2-http</artifactId>
+            <version>1.33.1</version>
```
geoand, line 28: I think we already have this version in a property, no?

sberyozkin: @geoand I did not find it, though this must be also be available the LC4J level and see we import the bom. For now I added the version property to the parent pom

geoand (approved):
Great!

PR #1357, opened 2025-03-12
Bump wc-chatbot from 0.2.0 to 0.2.1 and increase wc-chatbot window size

rsvoboda:
Bump wc-chatbot from 0.2.0 to 0.2.1 and increase wc-chatbot window size

Window size is adjustable since https://github.com/yishiashia/wc-chatbot/pull/7

Changing from the default 360 x 500 to 500 x 600 px to have more space for the interaction and returned content.


geoand (approved):
Nice, thanks!

PR #1383, opened 2025-03-25
Dev UI: always produce JsonRPCProvidersBuildItem

Ladicek:
This is due to an upcoming change in Execution Model Validation [1] where we need the `JsonRPCProvidersBuildItem` produced always, not only in dev mode. JSON RPC providers can use execution model affecting annotations, so we need to know about them, otherwise a non-dev build would fail with an incorrect validation error.

[1] https://github.com/quarkusio/quarkus/pull/46965


geoand (approved):

geoand:
Thanks!

geoand:
Merging this as the failure is unrelated to the PR (and I have fixed the sample in `main`).

PR #1393, opened 2025-03-27
McpClients bean 

### LINKED ISSUE #1345: have ability to get name for mcp's
as discussed in [#dev > mcp client](https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/mcp.20client)

idea was to use a map but it have some issue thus suggestion is to have like WS with OpenConnections.

jmartisk:
Fixes #1345 

Accepting suggestions on how to make the name a bit more distinguished.. 

Also moves the `@McpClientName` annotation out of the `runtime` package as it is meant to be public API, hence the `breaking-change` label.

### DIFF
--- a/mcp/deployment/src/main/java/io/quarkiverse/langchain4j/mcp/deployment/McpProcessor.java
+++ b/mcp/deployment/src/main/java/io/quarkiverse/langchain4j/mcp/deployment/McpProcessor.java
@@ -5,7 +5,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import io.quarkiverse.langchain4j.mcp.McpClients;
 import jakarta.enterprise.context.ApplicationScoped;
 
 import org.jboss.jandex.AnnotationInstance;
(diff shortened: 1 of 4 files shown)


geoand:
I personally don't find it terribly useful, but I'm dead set against it either :)

jmartisk:
Honestly my position is similar, but @maxandersen requested it so.. ;)

geoand:
@maxandersen why do you need this instead of just using standard CDI features?

maxandersen:
see details of discussion in https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/mcp.20client

issue is that to use built-in CDI I either have to 

1) know the names of tools upfront; since with mcp the usecase is to be able to add additional tools in configuration that is not an option.
2) use injection of all tools, which give me all tools but does not give me their name

Hence the ask for tools to know their name OR be able to get a map from names to the tools to use.

does that make sense?

(same issue exists for  when wanting to support multiple models in quarkus l4j)

geoand:
> use injection of all tools, which give me all tools but does not give me their name

Are you sure? You can extract the qualifiers from the instance, no?

maxandersen:
Hmm. Not something I found ? Got example ?

geoand:
Something like:

```java
boolean found = false;
        for (InstanceHandle<McpClient> handle : Arc.container().select(McpClient.class).handles()) {
            InjectableBean<McpClient> bean = handle.getBean();
            for (Annotation qualifier : bean.getQualifiers()) {
                if (qualifier instanceof McpClient mcpClient) }
                    if ("whatever".equals(name)) {
                        found = true;
                        break; 
                    }
                }
            }
        }
```

PR #1399, opened 2025-04-01
Fix native registration of user-provided model classes

xstefank:
Reproducer - https://github.com/xstefank/quarkus-reproducers/tree/main/lc4j-custom-model-native. Just run `-Dnative` with tests.


geoand (approved):
Thanks!

Issue #1397, opened 2025-03-28
Add support for MCP Claude configuration file


lordofthejars:
Add support for MCP Claude configuration file, so for example we can create the definition as:

```json
{
  "mcpServers": {
    "postgres": {
      "command": "npx",
      "args": [
        "-y",
        "@modelcontextprotocol/server-postgres",
        "postgresql://localhost/mydb"
      ]
    }
  }
}
```

And automatically (by storing them in a default directory) or by setting the path to the file as a configuration property in the Quarkus app.
This way, we are more aligned with what the user usually does when configuring MCP.

jmartisk:
I'm not so sure about supporting a vendor-specific file that is not defined in the MCP spec and other vendors beyond Claude will have different file formats. For example, Goose already uses a different configuration file for the same (see https://block.github.io/goose/docs/getting-started/using-extensions#config-entry). And there will be more and more over time. 

Also you won't be able to configure quarkus-specific properties like timeouts and logging (unless you combine this with `application.properties`, which makes it convoluted)

It may save some time, but it feels quite hacky and given that one of the main selling points of Quarkus is 'unified configuration in a single file', this doesn't fit there very well IMO...

jmartisk:
> I don't see why it would be an issue to generate JSON/YAML file of known MCP consumers.

I'm not following, what JSON/YAML file?

geoand:
Maybe I am understanding this issue incorrectly. I thought it was asking for us to generate snippet that users could just plug into Claude or Goose or whatever. Did I get that wrong?

jmartisk:
Hm, maybe it is me who understood it incorrectly. I thought it was about Quarkus understanding this Claude config file and using it instead of `application.properties`, but maybe it's rather about generating this file from what is known from `application.properties`. Both "kinda" make sense.

lordofthejars:
my idea is that Quarkus can understand Claude config as most of the examples out there are with this format

geoand:
Ah okay. That I indeed am not sure of...

maxandersen:
the "claude config" is being used beyond just claude. mcp.json in vscode and cursor follows same format.

So I see a value in being able to read that same format in to configure lets say a cli written with quarkus that could share config with these. 

But how to mix it best with config/defaults/timeout I'm not sure ... having to tell users to write a quarkus properties file isn't great either :)

geoand:
[...]

PR #1415, opened 2025-04-08
Support Claude Desktop config file

geoand:
- Closes: #1397

### DIFF
--- a/mcp/deployment/src/test/java/io/quarkiverse/langchain4j/mcp/test/McpClaudeConfigTest.java
+++ b/mcp/deployment/src/test/java/io/quarkiverse/langchain4j/mcp/test/McpClaudeConfigTest.java
@@ -1,0 +1,133 @@
+package io.quarkiverse.langchain4j.mcp.test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.entry;
+
+import java.util.List;
+
+import jakarta.enterprise.inject.Instance;
+import jakarta.inject.Inject;
+
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import dev.langchain4j.mcp.client.McpClient;
+import io.quarkiverse.langchain4j.mcp.runtime.McpClientName;
+import io.quarkiverse.langchain4j.mcp.runtime.config.McpClientRuntimeConfig;
+import io.quarkiverse.langchain4j.mcp.runtime.config.McpRuntimeConfiguration;
+import io.quarkus.test.QuarkusUnitTest;
+import io.smallrye.config.SmallRyeConfig;
+
+public class McpClaudeConfigTest {
+
+    @RegisterExtension
+    static QuarkusUnitTest unitTest = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClass(MockHttpMcpServer.class)
+                    .addAsResource(new StringAsset("""
+                            {
+                              "mcpServers": {
+                                "gemini": {
+                                  "command": "npx",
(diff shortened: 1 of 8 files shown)


mcp/deployment/src/test/java/io/quarkiverse/langchain4j/mcp/test/McpClaudeConfigTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @Test
+    public void testConfig() {
+        SmallRyeConfig smallRyeConfig = ConfigProvider.getConfig().unwrap(SmallRyeConfig.class);
+        McpRuntimeConfiguration mcpRuntimeConfiguration = smallRyeConfig.getConfigMapping(McpRuntimeConfiguration.class);
+        assertThat(mcpRuntimeConfiguration).isNotNull();
+        assertThat(mcpRuntimeConfiguration.clients()).containsOnlyKeys("client1", "filesystem", "gemini");
+
+        assertThat(mcpRuntimeConfiguration.clients().get("filesystem").command()).contains(List.of("npx", "-y",
+                "@modelcontextprotocol/server-filesystem", "/Users/username/Desktop", "/Users/username/Downloads"));
+        assertThat(mcpRuntimeConfiguration.clients().get("filesystem").environment()).isEmpty();
```
jmartisk, line 95: Could we also test the combination of Claude config with regular properties (for a single MCP client), for example you also add  a `quarkus.langchain4j.mcp.filesystem.log-requests` property and check that it was applied?

geoand: Sure thing

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/config/ClaudeDesktopConfigSourceProvider.java
```diff
@@ -0,0 +1,40 @@
...
+        if (McpRecorder.claudeConfigContents.isEmpty()) {
+            return List.of();
+        }
+
+        Map<String, String> configMap = new HashMap<>();
+        McpRecorder.claudeConfigContents.forEach(new BiConsumer<>() {
+            @Override
+            public void accept(String clientName, LocalLaunchParams localLaunchParams) {
+                if (!localLaunchParams.command().isEmpty()) {
+                    configMap.put(String.format("quarkus.langchain4j.mcp.%s.command", clientName),
```
jmartisk, line 26: I would assume the client name might contain dots, so should we wrap `%s` in quotes?

geoand: Yeah, that probably makes sense

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/config/McpBuildTimeConfiguration.java
```diff
@@ -31,4 +31,9 @@ public interface McpBuildTimeConfiguration {
     @ConfigDocDefault("true")
     Optional<Boolean> generateToolProvider();
 
+    /**
+     * File containing the MCP servers configuration in the Claude Desktop format.
+     * This configuration can only be used to configure {@code stdio} transport type MCP servers
```
jmartisk, line 36: So do I get it right that we read this file at build time, and it can't be overridden at runtime? This should be clearly stated in the javadoc then I guess

jmartisk: Yeah but the javadoc IMO should then make clear that this file is read during the build (so, consequently, neither changing the path to the file nor changing its contents at runtime is possible)

jmartisk: I mean, the user might not expect that the values can be overridden by config properties, but not by changing the claude file.

geoand: How does it look now?

jmartisk (approved):
As I mentioned earlier, I'm not super convinced we need this, but technically, LGTM.

Issue #1414, opened 2025-04-08
Integrate health checks


jmartisk:
Health checks were added in upstream via https://github.com/langchain4j/langchain4j/pull/2769. On the Quarkus side, after upgrading to beta3, we need to properly integrate them using MP Health

jmartisk:
Also, add the Quarkus-specific pinging in `QuarkusHttpMcpTransport.checkHealth()`

PR #1433, opened 2025-04-15
MP Health check for MCP clients, plus configurable ping timeout

jmartisk:
- Closes: https://github.com/quarkiverse/quarkus-langchain4j/issues/1414

### DIFF
--- a/mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/McpClientHealthCheck.java
+++ b/mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/McpClientHealthCheck.java
@@ -1,0 +1,60 @@
+package io.quarkiverse.langchain4j.mcp.runtime;
+
+import java.lang.annotation.Annotation;
+import java.util.HashMap;
+import java.util.Map;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Any;
+
+import org.eclipse.microprofile.health.HealthCheck;
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
+import org.eclipse.microprofile.health.Readiness;
+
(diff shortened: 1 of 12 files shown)


mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/McpClientHealthCheck.java
```diff
@@ -0,0 +1,60 @@
...
+            } catch (Exception e) {
+                builder.down().withData(name, e.getMessage());
+            }
+        }
+        return builder.build();
+    }
+
+    private Map<String, McpClient> getClientMap() {
+        Map<String, McpClient> map = new HashMap<>();
+        for (InstanceHandle<McpClient> handle : Arc.container().select(McpClient.class, Any.Literal.INSTANCE).handles()) {
```
geoand, line 49: Minor gripe: Can't we make this a proper injection point instead of doing a programmatic lookup?

geoand (commented):
Glad to see this! I just had one minor complaint

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/McpClientHealthCheck.java
jmartisk: Only if we merge https://github.com/quarkiverse/quarkus-langchain4j/pull/1393 first - I did here exactly what you [suggested](https://github.com/quarkiverse/quarkus-langchain4j/pull/1393#issuecomment-2765165445) the user should do instead :)

geoand: kk, makes sense :)

jmartisk: We need to access the name of the client.

Issue #1409, opened 2025-04-04
Add fine-grained control for MCP server(s) provided tools


mariofusco:
At the moment all tools exposed by all MCP servers configured in an application are automatically made available to all AI services in that application. This can potentially provide an AI service with tens of tools that it doesn't need and it isn't supposed to use, thus increasing the possibility of hallucinations.

It would be nice to have a chance to selectively decide if a AI service is supposed to use MCP servers and eventually which ones, in the same way to what we currently do for internal tools through the `@ToolBox` annotation. For instance we could have a similar `@McpToolBox` annotation taking an array of Strings representing the names of the MCP servers configured in the `application.properties`, something like

`@McpToolBox({"wolfram-llm", "weather"})`

where if no server name is provided, it means that we want to use all of them.

jmartisk:
The current idea is that if you need an AI service to use specific MCP clients, you use a tool provider.
```
@ApplicationScoped
public class MyToolProvider implements Supplier<ToolProvider> {

  private ToolProvider toolProvider;

  @Inject 
  @McpClientName("client1")
  McpClient client1;

  @Inject
  @McpClientName("client2")
  McpClient client2;
   
  @Override
  public ToolProvider get() {
      if(toolProvider == null) {          
            toolProvider = McpToolProvider.builder()
                    .mcpClients(mcpClient1, mcpClient2)
                    .build();
      }
      return toolProvider;
  }
}

@RegisterAiService(toolProviderSupplier = MyToolProvider.class)
public interface MyAiService {
  // ...
}
```

Right, that's much more verbose, but it gives the user more control. One can also develop a custom `ToolProvider` that provides MCP clients as well as custom tools on top. Isn't this enough?

PR #1434, opened 2025-04-15
Add fine-grained control for MCP server(s) provided tools

mariofusco:
This is an implementation for https://github.com/quarkiverse/quarkus-langchain4j/issues/1409

At the moment the `McpToolBox` annotation is only used to limit the MCP servers that a AI service can use. In other words if you have configured the MCP servers A, B and C, the following service:

```
@RegisterAiService
public interface MyService {

    @McpToolBox({"A", "C"})
    String chat(String query);
}
```

will be provided only for tools in the server A and C, but not B. If the `McpToolBox` annotation is present with no values or is not present at all it will behave in the same way, automatically providing all the tools for all the configured MCP servers. I did this to keep backward compatibility, but in reality what I would like to do is giving access to the MCP provided tools only to method annotated with `McpToolBox`, only the named ones or all if used without values. Any feedback on this is welcome.

I implemented this feature purely on the extension side and for this reason I had add to add a `QuarkusToolProviderRequest` and a `QuarkusMcpToolProvider`, but maybe we may this also in LangChain4j, so I could move the changes that I introduced with those classes directly there? /cc @dliubarskyi 

Finally I'm struggling to find a way to properly test this feature, so any suggestion about this will be also very welcome.

/cc @jmartisk @geoand

### DIFF
--- a/mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/McpToolBox.java
+++ b/mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/McpToolBox.java
@@ -1,0 +1,30 @@
+package io.quarkiverse.langchain4j.mcp.runtime;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import dev.langchain4j.service.tool.ToolProvider;
+import io.quarkiverse.langchain4j.RegisterAiService;
+
+/**
+ * When used on a method of an AiService annotated with {@link RegisterAiService}, the method will use the tools
+ * provided by the MCP servers named in {@code value}. If no name is provided than the method will automatically
+ * use all the MCP servers available.
+ * </p>
+ * Note that the filtering of the named MCP servers is possible only if the MCP extension is allowed to automatically
+ * generate a {@link ToolProvider} that is wired up to all the configured MCP clients,
+ * i.e. the {@code quarkus.langchain4j.mcp.generate-tool-provider} property is set to true (which is the default value).
+ * Conversely, if the AI service uses a custom {@link ToolProvider} than this annotation will have no effect and the
+ * wiring of specific MCP clients will have to be encoded in the {@link ToolProvider} itself.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD })
+public @interface McpToolBox {
+
+    /**
+     * MCP servers to use. In case no {@code value} is provided it will use all the MCP servers available.
+     */
+    String[] value() default {};
+}
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -1375,6 +1375,8 @@ public class AiServicesProcessor {
                 toolQualifierProviders.stream().map(
                         ToolQualifierProvider.BuildItem::getProvider).toList());
 
+        List<String> methodMcpClientNames = gatherMethodMcpClientNames(method);
+
         List<String> outputGuardrails = AiServicesMethodBuildItem.gatherGuardrails(method, OUTPUT_GUARDRAILS);
         List<String> inputGuardrails = AiServicesMethodBuildItem.gatherGuardrails(method, INPUT_GUARDRAILS);
 
@@ -1390,8 +1392,8 @@ public class AiServicesProcessor {
                 userMessageInfo, memoryIdParamPosition, requiresModeration,
                 returnTypeSignature(method.returnType(), new TypeArgMapper(method.declaringClass(), index)),
                 overrideChatModelParamPosition, metricsTimedInfo, metricsCountedInfo, spanInfo, responseSchemaInfo,
-                methodToolClassInfo, switchToWorkerThreadForToolExecution, inputGuardrails, outputGuardrails,
-                accumulatorClassName, responseAugmenterClassName);
+                methodToolClassInfo, methodMcpClientNames, switchToWorkerThreadForToolExecution, inputGuardrails,
+                outputGuardrails, accumulatorClassName, responseAugmenterClassName);
     }
 
     private Optional<JsonSchema> jsonSchemaFrom(java.lang.reflect.Type returnType) {
@@ -1837,6 +1839,26 @@ public class AiServicesProcessor {
         }
 
         return Arrays.stream(toolClasses).map(t -> t.name().toString()).collect(Collectors.toList());
+    }
+
+    private List<String> gatherMethodMcpClientNames(MethodInfo method) {
+        // Using the class name to keep the McpToolBox annotation in the mcp module
+        AnnotationInstance mcpToolBoxInstance = method.declaredAnnotation("io.quarkiverse.langchain4j.mcp.runtime.McpToolBox");
+        if (mcpToolBoxInstance == null) {
+            return null;
+        }
+
+        AnnotationValue mcpToolBoxValue = mcpToolBoxInstance.value();
+        if (mcpToolBoxValue == null) {
+            return Collections.emptyList();
+        }
+
+        String[] mcpClientNames = mcpToolBoxValue.asStringArray();
+        if (mcpClientNames.length == 0) {
+            return Collections.emptyList();
+        }
+
+        return Arrays.asList(mcpClientNames);
     }
 
     private DotName determineChatMemorySeeder(ClassInfo iface, ClassOutput classOutput) {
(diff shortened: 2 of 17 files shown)


geoand:
This is very interesting! I can definitely this being useful

jmartisk:
Yea I'd say we could change the upstream `ToolProviderRequest` to allow some filtering of the 'allowed' MCP clients instead of creating a new Quarkus-specific class, no? Like, let its constructor accept a `Predicate<McpClient>` and Quarkus would implement this predicate that performs the filtering. It would seem cleaner than further straying away from the upstream.

mariofusco:
> Yea I'd say we could change the upstream `McpToolProvider` to allow some filtering of the 'allowed' MCP clients instead of creating a new Quarkus-specific class, no? Like, let its constructor accept a `Predicate<McpClient>` and Quarkus would implement this predicate that performs the filtering. It would seem cleaner than further straying away from the upstream. We wouldn't need `QuarkusToolProviderRequest` then either.

@jmartisk @dliubarskyi I'm trying to follow this suggestion and more in general to move (in a more general way, e.g. introducing that predicate suggested by Jan) the logic that I implemented in the `QuarkusMcpToolProvider` back in langchain4j in order to get rid of it and of the `QuarkusToolProviderRequest` too, but I'm having hard time doing so and at this point I believe that this feature is entirely quarkus-specific and doesn't fit langchain4j use cases at all, for a twofold reason:

1. In reality I think that the problem doesn't exist at all in langchain4j. There the user programmatically creates a `McpToolProvider` with the list of MCP clients he wants to use. He normally has a different tool provider for each AI service, so it is already possible to define the list of clients that should be used for a specific service. Conversely in quarkus-langchain4j there is only one single tool provider automatically created [here](https://github.com/quarkiverse/quarkus-langchain4j/blob/adaaea839b5a7dfd67cc1a1861a1b410cf10c0f9/mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/McpRecorder.java#L87) and shared among all AI services, which is also the reason why I decided to put the list of the names of the MCP clients used by a specific service in the `QuarkusToolProviderRequest`.
2. Filtering the `McpClient` s to be used for a specific request, for instance through a `Predicate<McpClient>`, won't be enough to cover this use case for the simple fact that they don't have a name in langchain4j. The `McpClient`'s name is purely a quarkus feature, mostly deriving from how they are configured in the `application.properties`. Adding a (optional?) name to the langchain4j `McpClient` interface is also doable, but it seems to me a bit artificial and not very necessary to langchain4j needs alone.

WDYT? Is there anything that I could implement in langchain4j anyway or do we want to keep this only on the quarkus extension side? /cc @geoand

dliubarskyi:
My bad, I have misread the PR description:

```java
@RegisterAiService
public interface MyService {

    @McpToolBox({"A", "C"})
    String chat(String query);
}
```

I wrongly assumed that `A` and `C` are names of tools...

mariofusco:
> My bad, I have misread the PR description:
> 
> ```java
> @RegisterAiService
> public interface MyService {
> 
>     @McpToolBox({"A", "C"})
>     String chat(String query);
> }
> ```
> 
> I wrongly assumed that `A` and `C` are names of tools...

`A` and `C` in my example are the name of the MCP clients as they have been configured in the quarkus `application.properties` file. As I wrote MCP clients don't have a name in langchain4j so it is a bit difficult to map this feature there (and probably it doesn't make much sense).

Filtering the tools provided by a single MCP server also makes lot's of sense tbh, I'm also seeing some of them providing too many tools, but I believe this could be something orthogonal e not strictly dependent on what proposed here. @dliubarskyi I can also start working on this second feature if you agree and see some value in this.

mariofusco:
@jmartisk @geoand According to what also discussed with @dliubarskyi this feature doesn't make much sense in vanilla langchain4j, but I still believe it is useful here in the extension, so I would keep developing it here. Please let me know if you agree and in this case I will try to add at least one test for this.

geoand:
I believe this is a useful feature

dliubarskyi:
@mariofusco seems that users want to filter MCP tools as well as modify their descriptions: https://github.com/devoxx/DevoxxGenieIDEAPlugin/issues/670

mariofusco:
> The testing side will be a bit tricky. On the upstream langchain4j side, we have full-fledged MCP servers running using JBang, but here, we only have a mock MCP server that is exposed in the form of a REST endpoint - we might want to switch to doing something more similar to the upstream in the future... But with the current approach, I could imagine adding two more mock MCP servers that are simpler than `MockHttpMcpServer` and only offer a specific tool. But I guess you would also need a mock `ChatLanguageModel` (unless we want the tests to use an actual one) that receives the prompt along with a list of tools and it would verify that only the 'allowed' tools were passed to it. It's not super simple but I don't know if there's anything that is simpler yet provides a lot of value.

@jmartisk I added tests trying to follow your suggestions, thanks a lot for them. Please give it a look.

mariofusco:
> @mariofusco seems that users want to filter MCP tools as well as modify their descriptions: [devoxx/DevoxxGenieIDEAPlugin#670](https://github.com/devoxx/DevoxxGenieIDEAPlugin/issues/670)

Nice, but that's a filter of tools in the single MCP server. As discussed this is an orthogonal (and complementary) feature and can be implemented directly in langchain4j. I will give this a try.

mariofusco:
The last doubt I have (and the reason why I'm still leaving this as a draft for now) is how to express the fact that I may NOT want any MCP client for a specific AI service. At the moment annotating a service with `@McpToolBox` without any value or not annotating it all has the same effect of providing the service with all configured MCP clients. I did this to preserve backward compatibility, but I'm not sure this is a good idea. The 2 alternatives I see are:

1. By default a AI service doesn't use any MCP clients. You need to annotate it with `@McpToolBox`, specifying which clients you want to use, or not providing any value if you want to use them all.
2. By default a AI service uses all available MCP clients as it happens now. We could have some special value for the `@McpToolBox` annotation to say that we don't want any of them, something like `@McpToolBox("none")`.

Personally I found option 1. much clearer and less error prone, wdyt? Do you see any other possibility?

geoand:
Thanks a lot @mariofusco. I'm actually torn on this.

On one hand option `1` is closer to what the experience with Claude Desktop or Goose is like, but on the other hand option `2` is in line with what `@AiService(tools)` and `@ToolBox` do.

mariofusco:
> My point is that I don't really have a strong opinion, but I guess if I were forced to choose, I would go with `1`

Agreed, I will go with 1.

geoand:
Thanks!

I'll let @jmartisk review this as it directly touches his work

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/McpToolBox.java
```diff
@@ -0,0 +1,22 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import io.quarkiverse.langchain4j.RegisterAiService;
+
+/**
+ * When used on a method of an AiService annotated with {@link RegisterAiService}, the method will use the tools
+ * provided by the MCP servers named in {@code value}
```
jmartisk, line 12: Maybe we should clarify what relationship this has to the [quarkus.langchain4j.mcp.generate-tool-provider](https://docs.quarkiverse.io/quarkus-langchain4j/dev/mcp.html#quarkus-langchain4j-mcp_quarkus-langchain4j-mcp-generate-tool-provider) property. 

IIUC: 
If `generate-tool-provider` is on and the AI service/method is using that generated tool provider, then it works as a filter on top all configured MCP servers.  
But, if `generate-tool-provider` is false or the service/method is declared to use a custom tool provider, then it will filter the MCP clients that are part of that custom provider. 

... Or will it? From looking at the code, I think that if the user creates their own `ToolProvider` bean that implements `dev.langchain4j.service.tool.ToolProvider`, AND they apply a `@McpToolBox` on an Ai service that uses that provider, then the filtering won't be applied, no? If that is intended, the javadoc should warn about it

mariofusco: I must admit that I wasn't thinking to this scenario, but your understanding of what happens when the user create his own `ToolProvider` is totally correct. I tried to clarify it in the JavaDoc of this class. Please give it one more look.

jmartisk:
Not completely sure whether it wouldn't be cleaner to just let the user to create their own tool providers  for this purpose (if they want to allow just specific MCPs for a specific AI service/method, create a specific tool provider that uses only the desired clients). An annotation looks simpler, but look at how much complexity it is adding. Anyway, I don't object, if others find it useful, then so be it.

jmartisk (approved):

mariofusco:
> Not completely sure whether it wouldn't be cleaner to just let the user to create their own tool providers for this purpose (if they want to allow just specific MCPs for a specific AI service/method, create a specific tool provider that uses only the desired clients). An annotation looks simpler, but look at how much complexity it is adding. Anyway, I don't object, if others find it useful, then so be it.

I totally see your point about the added complexity, but under a users point of view I also believe that a very small fraction of them will decide to create their own `ToolProvider` and they would prefer the simplicity of this annotation. Also what I don't like of the current situation is that once you have configured one or more MCP server, each and every AI service will use all of them by default.

Anyway I'm also not strongly opinionated on this and I will let @geoand and @dliubarskyi to take the final decision.

geoand:
Let's add this for now and see how it goes

PR #1438, opened 2025-04-20
Prevent infinite loop when using tool-choice in watsonx.ai

andreadimaio:
This commit avoids an infinite loop when a developer uses the `tool-choice` option.  The fix is for the `watsonx.ai` module, but maybe this can also affect other models in `quarkus-langchain4j` and could be something to move in the core part.

I've done a quick fix in `watsonx.ai` because I'm working on a project and I need to use the `tool-choice` in the next few days.
Let's understand if this is something that can be moved to the core or if it is a problem specific to the watsonx.ai module.

### DIFF
--- a/model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxChatModel.java
+++ b/model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxChatModel.java
@@ -1,6 +1,7 @@
 package io.quarkiverse.langchain4j.watsonx;
 
 import static io.quarkiverse.langchain4j.watsonx.WatsonxUtils.retryOn;
+import static java.util.Objects.nonNull;
 
 import java.util.List;
 import java.util.Set;
@@ -8,6 +9,8 @@ import java.util.concurrent.Callable;
 
 import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.data.message.ToolExecutionResultMessage;
 import dev.langchain4j.exception.UnsupportedFeatureException;
 import dev.langchain4j.model.chat.Capability;
 import dev.langchain4j.model.chat.ChatLanguageModel;
@@ -83,8 +86,29 @@ public class WatsonxChatModel extends Watsonx implements ChatLanguageModel {
                 ? toolSpecifications.stream().map(TextChatParameterTool::of).toList()
                 : null;
 
-        TextChatRequest request = new TextChatRequest(modelId, spaceId, projectId, messages, tools,
-                TextChatParameters.convert(parameters));
+        TextChatParameters textChatParameters = TextChatParameters.convert(parameters);
+
+        if (nonNull(parameters.toolChoice()) && parameters.toolChoice().equals(ToolChoice.REQUIRED) && messages.size() > 0) {
+            // This code is needed to avoid a infinite-loop when using the AiService
+            // in combination with the tool-choice option.
+            // If the tool-choice option is not removed after calling the tool,
+            // the model may continuously reselect the same tool in subsequent responses,
+            // even though the tool has already been invoked. This leads to an infinite loop
+            // where the assistant keeps generating tool calls without progressing the conversation.
+            // By explicitly removing the tool-choice field after the tool has been executed,
+            // we allow the assistant to resume normal message generation and provide a response
+            // based on the tool output instead of redundantly triggering it again.
+
+            int LAST_MESSAGE = chatRequest.messages().size() - 1;
+            ChatMessage lastMessage = chatRequest.messages().get(LAST_MESSAGE);
+            if (lastMessage instanceof ToolExecutionResultMessage) {
+                tools = null;
+                textChatParameters.cleanToolChoice();
+            }
+        }
+
+        TextChatRequest request = new TextChatRequest(modelId, spaceId, projectId, messages, tools, textChatParameters);
+
         TextChatResponse response = retryOn(new Callable<TextChatResponse>() {
             @Override
             public TextChatResponse call() throws Exception {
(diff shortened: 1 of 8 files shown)


geoand:
Very nice @andreadimaio!

@dliubarskyi is this something that can happen generally?

dliubarskyi:
@geoand yes, but it seems like a misconfiguration to me. There is a low-level `ChatModel` that has a specific tool-calling configuration (in this case "always call a tool X"). Then there is a high-level AI Service that uses that `ChatModel` but has it's own tool calling configuration (in this case "you can use tools X, Y"). So perhaps AI service should always override tool calling configuration of `ChatModel` (both `toolSpecifications` and `toolChoice`). `toolChoice=REQUIRED` configured on a `ChatModel` does not seem compatible with AI Service.

geoand:
Thanks makes sense @dliubarskyi . @andreadimaio what exactly is your use case here?

andreadimaio:
> @andreadimaio what exactly is your use case here?

What I want to achieve is an `AiService` that **always calls a tool**.  
To do this, I have annotated an interface with `@RegisterAiService` and configured it using the `tool-choice` option:

```java
@ApplicationScoped
@RegisterAiService(
    tools = ToolsClass.class,
    modelName = "my-agent"
)
public interface MyAgent {
    ...
}
```

```properties
quarkus.langchain4j.watsonx.my-agent.chat-model.model-id=...
quarkus.langchain4j.watsonx.my-agent.chat-model.tool-choice=...
...
```

However, because `tool-choice` is always configured, the result is that the model continues to call the tool indefinitely without ever returning a final answer.

andreadimaio:
> @geoand yes, but it seems like a misconfiguration to me. There is a low-level `ChatModel` that has a specific tool-calling configuration (in this case "always call a tool X"). Then there is a high-level AI Service that uses that `ChatModel` but has it's own tool calling configuration (in this case "you can use tools X, Y"). So perhaps AI service should always override tool calling configuration of `ChatModel` (both `toolSpecifications` and `toolChoice`). `toolChoice=REQUIRED` configured on a `ChatModel` does not seem compatible with AI Service.

I'm not quite following your consideration, why exactly do you think there's a misconfiguration?  

The `AiService` is configured to always call a tool, using the `tool-choice` option.  
The expected behavior is that the model calls the tool, receives the result, and then elaborates that result to generate a final answer for the user.

The issue is that if `tool-choice` is set to `REQUIRED` for every request, then after the tool returns its response, the model does not proceed to generate an answer. Instead, it uses the tool's output as input to call the tool again, leading to a loop.

That's why I had to temporarily remove the `tool-choice` option from the second request, so the model can generate a proper response instead of re-invoking the tool.

dliubarskyi:
> The AiService is configured to always call a tool, using the tool-choice option.

The `tool-choice` option is applicable to the `ChatModel`, not AI service. There is currently no way to enforce tool call on the AI Service level of API. Perhaps we need to introduce a configuration on the AI Service level to be able to achieve what you have described. In this case we would need to automatically add `toolChoice=REQUIRED` to the first LLM call (instead of removing it as it is now). WDYT?

andreadimaio:
> The `tool-choice` option is applicable to the `ChatModel`, not AI service. There is currently no way to enforce tool call on the AI Service level of API. 

Now I undernstand your point, but the `@RegisterAiService` inheritance the properties configuration of the `ChatModel` that will be automatically created by quarkus. So, this is the reason why I encountered this problem. From my undestanding the problem in langchain4j is not present because the AiService "clean" the `tool-choice` option.

> Perhaps we need to introduce a configuration on the AI Service level to be able to achieve what you have described. In this case we would need to automatically add `toolChoice=REQUIRED` to the first LLM call (instead of removing it as it is now). WDYT?

Yes, this could be a more elegant solution, but be careful, each model can manage `tool-choice` in a different way, so how this happens will need to be configured by the provider.

model-providers/watsonx/runtime/src/main/java/io/quarkiverse/langchain4j/watsonx/WatsonxChatModel.java
```diff
@@ -83,8 +86,29 @@ public ChatResponse doChat(ChatRequest chatRequest) {
...
+            // where the assistant keeps generating tool calls without progressing the conversation.
+            // By explicitly removing the tool-choice field after the tool has been executed,
+            // we allow the assistant to resume normal message generation and provide a response
+            // based on the tool output instead of redundantly triggering it again.
+
+            int LAST_MESSAGE = chatRequest.messages().size() - 1;
+            ChatMessage lastMessage = chatRequest.messages().get(LAST_MESSAGE);
+            if (lastMessage instanceof ToolExecutionResultMessage) {
+                tools = null;
+                textChatParameters.cleanToolChoice();
```
dliubarskyi, line 106: In this place (`WatsonxChatModel`) we are on the lowest level of LC4j API, so user expects to have complete freedom of what they are doing. If they want to call LLM twice with `ToolChoice.REQUIRED`, they should be able to do so. IMHO this is not the right place to "fix" this issue for multiple reasons, also because the same fix should be duplicated for each `ChatModel` implementation

andreadimaio: I agree with what you say, and that's one of the reasons why I said that this fix should probably be ported to the core module of `quarkus-langchain4j`, perhaps the ideal class is `AiServiceMethodImplementationSupport`.

That way, the developer can use the AIService which will take care of everything, or if they want more control instead, they can go straight to using the `ChatLanguageModel`.

andreadimaio: For now, this quick fix allows me to move forward in the scenario described above using the AIService (with all the benefits) and I considered it valid because it is true that in this case I do not allow the developer to use the `tool-choice` as it prefers, but it is also true that today this always leads to an infinite loop and I do not see scenarios in which this could be useful.

But we're on the same page, if we take it outside of the ChatLanguageModel, it'll be good for all the provides.

geoand: Okay, let's merge this for now so we can unblock you and you can take up the general solution subsequently

andreadimaio: These days I will be busy with other things, but in the following weeks I can propose a draft PR that will allow this solution to be implemented in the `AiServiceMethodImplementationSupport' class.

geoand: 🙏🏽

andreadimaio:
I'm thinking about a possible way to solve the looping problem directly in the `AiServiceMethodImplementationSupport` class. One idea could be to use the `toolChoice` parameter of the `ChatRequestParameters` class.

The idea could be to automatically switch `toolChoice` from `REQUIRED` to `AUTO` when the last message from the LLM is a `ToolExecutionResultMessage`, to prevent the model from repeatedly calling the same tool.

However, the challenge is that `AiServiceMethodImplementationSupport` doesn't currently have access to the `toolChoice` configuration at runtime. To proceed with this approach, we would probably need to define a common configuration mechanism that exposes the `toolChoice` setting for all `ChatLanguageModel` and `StreamingChatLanguageModel` implementations.

Does this direction make sense? Or do you have any other ideas?

@geoand

geoand:
It sounds reasonable to me

andreadimaio:
> It sounds reasonable to me

Ok, I'll try to implement it this way

PR #1453, opened 2025-05-02
Prevent infinite loop when using tool-choice in AiService

andreadimaio:
A few days ago I did a commit to prevent an infinite loop when using `tool-choice` in watsonx.ai (more details at #1438). The idea with this commit is to move the fix to the AiService level (`AiServiceMethodImplementationSupport`), so that all providers in quarkus-langchain4j will benefit from this fix.

Note: The fix won't work if the AiService is used  with `TokenStream` as return type, in this case the fix must come from langchain4j.


geoand (approved):
Thanks a lot!

PR #1454, opened 2025-05-03
Make Gemini API version configurable

sberyozkin:
This is a draft PR as `v1` has does not recognize `systemInstruction`, `tools` and `responseMediaType` JSON properties

### DIFF
--- a/integration-tests/ai-gemini/src/main/java/org/acme/example/gemini/aiservices/GeminiResource.java
+++ b/integration-tests/ai-gemini/src/main/java/org/acme/example/gemini/aiservices/GeminiResource.java
@@ -16,7 +16,7 @@ import io.vertx.core.json.JsonObject;
 public class GeminiResource {
 
     @POST
-    @Path("v1beta/models/gemini-1.5-flash:generateContent")
+    @Path("v1/models/gemini-1.5-flash:generateContent")
     @Produces("application/json")
     @Consumes("application/json")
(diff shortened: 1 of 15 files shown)


geoand:
👍🏽

geoand:
@sberyozkin is there something more that needs to be done on this one?

sberyozkin:
Hi @geoand, I thought it would be an easy update (As far as I recall, Gemini 2.5 was no available with `v1Beta`), but after I did it, I noticed those tags like `tools` were not recognized with `v1`, and then I started actually  looking for the docs, and found:
https://ai.google.dev/gemini-api/docs/api-versions

According to that `v1`  does not support function calls yet. I think I saw somewhere that `tools` were renamed but I can't find that information again anywhere.

So I'm not sure right now how to evolve this PR. I checked with Guillaume as well. I guess we need to wait until some docs about how exactly function calls will be represented in `v1` and then we'd have version specific generate content request/response support.

sberyozkin:
@lordofthejars Hi Alex, have you seen some `v1` specific migration guide somewhere, how `tools` are going to be represented, etc ?

lordofthejars:
> @lordofthejars Hi Alex, have you seen some `v1` specific migration guide somewhere, how `tools` are going to be represented, etc ?

I have not looked at it. I know that tools is not working nowadays, but I have not checked.

PR #1470, opened 2025-05-08
Add McpClientAuthProvider

### LINKED ISSUE #1455: Support token propagation for MCP SSE client
Apply the `ModelAuthProvider` idea, introduce `McpAuthProvider` for users be able to pick up tokens from various sources, and ship an OIDC implementation that will allow user authentication token propagation to MCP servers

sberyozkin:
Fixes #1455.

The demo works but tests have to be added, demo README fixed.

I'm also hesitating if it should be `McpClientAuthProvider` or `McpAuthProvider` to keep it a bit simpler, as well as with this interface package


geoand:
Thanks @sberyozkin!

I vote for  `McpClientAuthProvider` as you've done

sberyozkin:
@geoand Thanks, ouch, I missed your comment and renamed it to `McpAuthProvider`, but I'll sort it out once I'm ready to open for review.

The other thing I did, I moved it to `io.quarkiverse.langchain4j.mcp.auth` to try to keep it similar  to how `ModelAuthProvider` is packaged, even though it is mainly for HTTP. Are you OK with that ?

geoand:
👍🏽

sberyozkin:
It is getting closer, with some cleanup pending but I'm thinking of restructuring the demo a bit to run it similarly to how the secure-mcp-sse-server in quarkus-mcp-server is run, so that in this case, we could show both dev and prod modes for Quarkus LangChain4j MCP client, and possibly show something more interesting that just `@Authenticated` control, may be `@PermissionAllowed`...

geoand:
@sberyozkin should this be taken out of draft?

sberyozkin:
Hi @geoand Sure, I did get the point yesterday where I was happy enough with the demo, I'd like to add MCP docs, to describe the provider, and also doc `ModelAuthProvider` along the way.

Do you think we can keep `McpAuthProvider` name, or prefer to restore the original `McpClientAuthProvider` ? THat would be fine, my only reservation is how to deal with duplicate `-client-` in the `quarkus-oidc-client` based MCP auth provider that I'll add after the PR is done, to demo it without the user login :-)

sberyozkin:
The only other thing is that it may be worth updating README to explain how to use GitHub if users would like to use MCP client against what we demoed in `quarkus-mcp-sever` as implied in #1476.

I need to think a bit more how to make it easier

geoand:
> Do you think we can keep McpAuthProvider name, or prefer to restore the original McpClientAuthProvider

I vote for the original `McpClientAuthProvider`

sberyozkin:
@geoand Hi Georgios, I'm not sure why the compilation failure occurred, do I have to add an extra dependency ? I don't see health check related dependency in `samples/mcp-sse-client-server/mcp-client`

geoand:
Have you rebased onto the latest?

sberyozkin:
I think so, I did it yesterday, `main` builds ok, `samples/mcp-sse/client-server` is also good but it may be getting the healthcheck api from one of its dependencies, I'll have a look

sberyozkin:
Thanks, I added the optional `smallrye-health` dependency for the moment, let me finalize the PR, may be it will be easier to check once it is reviewed and merged.

geoand:
Sure. Once you do that I'll have a look at the health module shouldn't be necessary.

sberyozkin:
I just had to revert some vertex ai doc updates which I picked up earlier, need to doc McpClientAuthProvider, and it should be ready

sberyozkin:
@geoand @jmartisk This PR is ready for review now.

My initial plan was use Google authentication, because I really wanted to show the case where a single access token acquired after Google login can be used to access not only MCP server but also AI Gemini (with no API key required).

However, after I saw #1476 opened by @zbendhiba, I realized, given that users have already invested time into setting up GitHub OAuth2 applications with the Quarkus MCP server demo, asking them now to create Google applications would not be really great, so I reworked it to use `GitHub` only, thus also requiring AI Gemini API key which is easy to get.

I'll start working on a blog entry quite soon, to analyze the demo deeper.

Going forward, my plan is, in addition to `quarkus-langchain4j-oidc-mcp-auth-provider`, add `quarkus-langchain4j-oidc-client-mcp-auth-provider` that we'd be able to use without the user login with MCP client. 

The interface is called `McpClientAuthProvider` as agreed with Georgios, but I thought I'd avoid adding `-client` to the extension names, as with the next planned extension it would two `-client` in the name :-)

jmartisk:
Thanks! I'll try to review tomorrow

sberyozkin:
Compared to the Quarkus MCP server demo, here a bit more realistic augmentation scenario is introduced, to show how Quarkus MCP server demo can augment tokens with extra permissions.

geoand:
https://github.com/quarkiverse/quarkus-langchain4j/pull/1484 fixes the issue with the health check

geoand:
Very nice work @sberyozkin!

sberyozkin:
Thanks @geoand, and sure, I'll rebase once #1484 is merged

geoand:
I just merged it

samples/secure-mcp-sse-client-server/secure-mcp-client/src/main/resources/application.properties
```diff
@@ -0,0 +1,15 @@
+quarkus.langchain4j.mcp.generate-tool-provider.transport-type=http
```
jmartisk, line 1: See the same comment that I made at https://github.com/quarkusio/quarkus-workshop-langchain4j/pull/73#discussion_r2083916912

Where did you get the name `generate-tool-provider` from? The fact that two people did this mistake at the same time makes me think that there is something wrong with our documentation or some other samples.

samples/secure-mcp-sse-client-server/README.md
```diff
@@ -0,0 +1,89 @@
...
+with the following two properties instead:
+
+```properties
+%quarkus.oidc.credentials.client-secret.method=post
+%quarkus.oidc.credentials.client-secret.value=${github.client.secret}
+```
+
+# AI Gemini API key
+
+Get [AI Gemini API key](https://aistudio.google.com/app/apikey) and set the `quarkus.langchain4j.ai.gemini.api-key` property.
```
jmartisk, line 27: To some users it might not be clear whether this is on the server or the client, and how exactly they are supposed to set it. The client's properties file says

```
quarkus.langchain4j.ai.gemini.api-key=${ai.gemini.api.key}
```
, so maybe we should tell the user to set the `AI_GEMINI_API_KEY` env property for the client? 
The same applies above to  `quarkus.oidc.client-id` and `quarkus.oidc.credentials.secret`, it could use a bit more clarity

samples/secure-mcp-sse-client-server/README.md
```diff
@@ -0,0 +1,89 @@
...
+Package and run the mcp client component:
+
+```shell
+mvn clean package
+java -jar target/quarkus-app/quarkus-run.jar
+```
+
+Go to `http://localhost:8080`, login with GitHub, request a poem and confirm that your name is returned alongside the poem about Java.
+
+# Securiy
```
jmartisk, line 82: typo

samples/secure-mcp-sse-client-server/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+Both Quarkus MCP server main SSE and tool endpoints require authentication.
```
jmartisk, line 84: "tool endpoint" may be a bit misleading term (I usually say "POST" endpoint), and will be rendered obsolete when we move to the streamable HTTP transport. Could we rather say "All HTTP requests to the MCP server require authentication"? That sounds safer to me.

Then below, you could say that "the endpoint that serves for tool calling" requires the `read:name` permission.

jmartisk (commented):
Nice, I managed to make it work. A few docs/usability comments below

samples/secure-mcp-sse-client-server/secure-mcp-client/src/main/resources/application.properties
sberyozkin: @jmartisk I think I copied it from the `mcp-sse-client-server/mcp-client` :-), sure,  I'll update

jmartisk: I see, ok, I'll change it there, thanks :)

sberyozkin: I can update that demo too along the way

samples/secure-mcp-sse-client-server/secure-mcp-client/src/main/java/io/quarkiverse/langchain4j/sample/PoemResource.java
```diff
@@ -0,0 +1,36 @@
...
+public class PoemResource {
+
+    static final String USER_MESSAGE = """
+            Write a short 1 paragraph poem about a Java programming language.
+            Please start by greeting the currently logged in user by name and asking to enjoy reading the poem.""";
+
+    @RegisterAiService
+    public interface GeminiPoemService {
+        @UserMessage(USER_MESSAGE)
+        @McpToolBox
```
jmartisk, line 24: This is unnecessary and only bloats the code. When you don't specify a `value` for `@McpToolBox`, it just takes all the available mcp clients.

geoand: I am pretty sure that's **not** the case anymore, right @mariofusco ?

jmartisk: Ok it seems it is necessary (not that I agree with this change, but that's for another discussion)

samples/secure-mcp-sse-client-server/README.md
sberyozkin: Hi @jmartisk I added notes about the env properties, but also explained how to update `application.properties` if preferred by some users. When I was testing setting env properties, it felt a little bit unusual to me exporting all theses keys as env props, and with `application.properties` one only needs to replace the parameterized values...

Not sure if it makes the README clearer :-), but I tried, have a look please

jmartisk (approved):

geoand (approved):

samples/secure-mcp-sse-client-server/secure-mcp-client/src/main/java/io/quarkiverse/langchain4j/sample/PoemResource.java
mariofusco: Yes, I confirm it is now necessary to add that annotation to use the configured MCP clients. This is in line with what the `@ToolBox` annotation does and also I don't see how you could say that you don't want that a service could use any MCP tool. Anyway, I'm open to discuss this, so please let me know if you have any proposal to improve this.

PR #1471, opened 2025-05-09
Fix reranking cohere API

lordofthejars:
Updates the API of ranking when using Cohere: https://docs.cohere.com/reference/rerank


geoand:
Thanks for this!

I assume you've tried it and all works as expected, right?

lordofthejars:
I have no idea why this is not working (URL is ok, protocol is ok, api key, and path) Probably there is a small thing that makes still return a 404

lordofthejars:
Ok it owrks, the problem was that the model ID is incorrect. But the exception is not putting the response on the exception too, so it just throws 404, but I think that if a 404 is thrown or in general any error code, and there is a body message, this body message should be consumed and put in the exception.

lordofthejars:
Ok it works it can be merged.Also docs are wrong: https://docs.quarkiverse.io/quarkus-langchain4j/dev/reranking.html I will send another PR on Monday.

geoand:
Thanks for checking!

PR #1475, opened 2025-05-10
Add chatbot example with Easy RAG implementation in Kotlin

kpavlov:
This commit introduces a new chatbot example leveraging Easy Retrieval-Augmented Generation (RAG) using Quarkiverse LangChain4j and Kotlin. It includes a WebSocket-based chatbot interface, dynamic import map generation, UI enhancements, and relevant dependencies in `pom.xml`. The example showcases RAG-based conversational AI for banking product queries.


geoand:
cc @jmartisk

geoand:
Thanks a lot for this!

I'll let @jmatrisk review, but I'll just say that once everything is ready to go, we'll need to have the commits squashed

samples/chatbot-easy-rag-kotlin/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+The chatbot is a financial assistant that:
+1. Answers questions about financial products using information retrieved from documents
+2. Provides current stock prices for selected companies (AAPL, GOOG, MSFT)
+3. Analyzes sentiment in user messages
+4. Content moderation: Detects malicious content in user messages and sends a warning by email, if detected
+
+### Setting up the document catalog
+
+Before using the RAG functionality, you need to create a `catalog` directory with your financial product documents:
```
jmartisk, line 35: This sounds like a prerequisite before running the app, but the app is configured to use the `catalog` directory that is under the root of the project, so this isn't actually needed, no? Rather we could just make a note that the app is configured to look for documents in a `catalog` directory relative to the current working directory (this will be useful in case people will want to run it in prod mode too)

samples/chatbot-easy-rag-kotlin/src/main/resources/META-INF/resources/components/demo-title.js
```diff
@@ -0,0 +1,72 @@
...
+        return html`
+            <div class="title">
+                <h1>Financial Assistant</h1>
+            </div>
+            <div class="explanation">
+                This demo shows how to build a chat bot powered by LLM and retrieval-augmented generation (RAG).
+                The application uses the Easy RAG extension to retrieve information about financial products from provided
+                documents stored in a catalog directory.
+                The description of the different accounts is ingested into a Redis database and relevant information
+              is sent to the LLM before answering the user.It also provides stock price information for selected companies.
```
jmartisk, line 49: Could you add a few example prompts on this page, to recommend what the user should try? We generally do this in other similar sample apps, and I find it extremely helpful. Otherwise, the user who runs the app might not have clear ideas what to say to the bot.

For example
- "You are stupid" to trigger the moderation. Also tell the user to look at http://localhost:8080/q/dev-ui/io.quarkiverse.mailpit.quarkus-mailpit/mailpit-ui afterwards

jmartisk: - "What is the current stock price of GOOG?" (Probably also make an explicit note that it does not really use live data, but random numbers, in case somebody doesn't read the code)

jmartisk: - "What is the minimum opening deposit for the standard savings account?"

samples/chatbot-easy-rag-kotlin/src/main/resources/META-INF/resources/components/demo-title.js
```diff
@@ -0,0 +1,82 @@
...
+                This demo shows how to build a chat bot powered by LLM and retrieval-augmented generation (RAG).
+                The application uses the Easy RAG extension to retrieve information about financial products from provided
+                documents stored in a catalog directory.
+                The description of the different accounts is ingested into a Redis database, and relevant information
+                is sent to the LLM before answering the user. It also provides stock price information for selected companies.
+
+                Here are some examples of questions that you can ask:
+                <ul>
+                  <li>"What is the minimum opening deposit for the standard savings account?" — The Agent is aware of characteristics of financial products</li>
+                  <li>"What is the current stock price of GOOG?" — The Agent uses Tools to retrieve the data, but in this example it does not really use live data, but random numbers, in case somebody doesn't read the code)</li>
```
jmartisk, line 54: Literally putting "in case somebody doesn't read the code" seems a bit excessive and might even offend somebody :)

kpavlov: 🤦🏻 true

geoand:
Can you please squash the commits so we can merge this?

geoand:
Something seems to have gone wrong with rebase, no?

kpavlov:
See #1480 instead

PR #1478, opened 2025-05-12
Make MCP Client Health check actually optional

geoand:
This is needed to avoid warnings at both build time and runtime which stemmed from the fact that the
health check was previously always a bean.

The build time warnings were like:

```
2025-05-12 16:07:09,173 WARN  [io.qua.arc.pro.BeanArchives] (build-40) Failed to index org.eclipse.microprofile.health.HealthCheck: Class does not exist in ClassLoader QuarkusClassLoader:Deployment Class Loader: DEV for quarkus-langchain4j-sample-mcp-client-1.0-SNAPSHOT@5be49b60
2025-05-12 16:07:50,576 INFO  [io.qua.arc.pro.IndexClassLookupUtils] (build-40) Class for name: org.eclipse.microprofile.health.HealthCheck was not found in Jandex index. Please ensure the class is part of the index.
```

and the runtime ones:

```
2025-05-12 16:07:51,180 WARN  [io.qua.arc.ComponentsProvider] (Quarkus Main Thread) Unable to load removed bean type [io.quarkiverse.langchain4j.mcp.runtime.McpClientHealthCheck]: java.lang.NoClassDefFoundError: org/eclipse/microprofile/health/HealthCheck
```

FWIW, in a lot of core extensions in Quarkus this is not an issue because those extensions do **not** include an empty `beans.xml` file.

### DIFF
--- a/mcp/deployment/src/main/java/io/quarkiverse/langchain4j/mcp/deployment/McpProcessor.java
+++ b/mcp/deployment/src/main/java/io/quarkiverse/langchain4j/mcp/deployment/McpProcessor.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 
 import dev.langchain4j.mcp.client.McpClient;
 import dev.langchain4j.service.tool.ToolProvider;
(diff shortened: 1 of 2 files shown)


jmartisk:
Thanks

PR #1479, opened 2025-05-12
Build samples

kpavlov:
Update the build process in the GitHub Actions workflow to include an additional compilation step for the `samples` directory.

### DIFF
--- a/.github/workflows/build-pull-request.yml
+++ b/.github/workflows/build-pull-request.yml
@@ -41,7 +41,10 @@ jobs:
           cache: 'maven'
 
       - name: Build with Maven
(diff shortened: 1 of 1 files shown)


geoand:
We actually used to do this, but we ended up changing it because the samples reflect the latest **released** version.

kpavlov:
@geoand , shouldn't we put the latest released version here: https://github.com/quarkiverse/quarkus-langchain4j/blob/main/samples/pom.xml#L7 ?

geoand:
@kpavlov each sample has the latest release version that is updated automatically by release process

PR #1482, opened 2025-05-12
Refactor SentimentAnalyzer, introduce CustomerCallbackScheduler

kpavlov:
# Refactor SentimentAnalyzer, introduce CustomerCallbackScheduler

- Refactored sentiment SentimentAnalyzer to use ChatModel directly, instead of AiService. This demo is using suspend functions and Kotlin DSL, provided by LangChain4j extensions.
- Fixed pom.xml
- Cleanup QuestionProcessor
- Add QuarkusTest verifying interaction with LLM
- Update language model to `gpt-4.1-mini`
- Load env variables from `.env` file
- Time and CustomerCallbackScheduler tools, improved prompts
- Renamed Stocks to MarketData
- Removed NoopChatMemoryProvider and NoopRetrievalAugmentor
- Remove TestEnvironment and inline MockOpenai initialization for simplicity
- Update README with `.env` file option for API key setup


geoand:
Thanks for this.

> Refactored sentiment SentimentAnalyzer to use ChatModel directly, instead of AiService. This demo uses suspend functions and Kotlin DSL provided by LangChain4j extensions.

Is this really an improvement?

samples/chatbot-easy-rag-kotlin/src/test/kotlin/io/quarkiverse/sample/chatbot/TestEnvironment.kt
```diff
@@ -0,0 +1,7 @@
+package io.quarkiverse.sample.chatbot
```
maxandersen, line 1: this doesn't seem necessary here does it? would it not be sufficient to mock the interface?

kpavlov: This is an integration-level test, but since SentimentAnalizer is not exposed to the web, I had to do a gray-box unit-level test with bean injection. In the production application, we do need black-box tests able to catch also  langchain4j regressions, e.g. response parsing. It is much safer to test end-to end, with the network layer. 

This example is a little bit artificial. It demonstrates continuous testing with mock LLMs.

TestEnvironment abstraction is very helpful in JUnit5 tests when we want to make sure that some services will be shared across all the tests. It can be achieved with static fields in Singletons. Kotlin's `object` is a singleton, under the hood

kpavlov:
> Is this really an improvement?

Yes, it's a simplification. If we want to have another LLM request without using RAG and ChatMemory, then injecting ChatModel directly to the service is better. Also cheaper, as it would not send RAG and Chat history with every request.

It can be achieved also with AiServices with [NoopChatMemoryProvider](https://github.com/quarkiverse/quarkus-langchain4j/blob/7ec8ccbb86786195301f6da09b1108f19d3eaf90/samples/chatbot-easy-rag-kotlin/src/main/kotlin/io/quarkiverse/langchain4j/sample/chatbot/internal/NoopChatMemoryProvider.kt) and [NoopRetrievalAugmentor](https://github.com/quarkiverse/quarkus-langchain4j/blob/7ec8ccbb86786195301f6da09b1108f19d3eaf90/samples/chatbot-easy-rag-kotlin/src/main/kotlin/io/quarkiverse/langchain4j/sample/chatbot/internal/NoopRetrievalAugmentor.kt), but it is weird.

I want to move NoopChatMemoryProvider and NoopRetrievalAugmentor to main Quarkus-LC4j or even to the LC4j repo, because it can be useful for such scenarios. Let me remove NoopChatMemoryProvider and NoopRetrievalAugmentor in the next PR?

PR #1488, opened 2025-05-14
Improve styles in Kotlin sample

kpavlov:
# Improve styles in Kotlin sample

Make it look better on demo screen

- Adjust demo-title explanation width to 90%
- Dropped `height: 100vh` and `overflow: hidden` to fix scrolling
- Increase text size in chat-bubble


geoand (approved):
Thanks!

PR #1495, opened 2025-05-15
Refactor Chatbot Kotlin sample

kpavlov:
# Refactor Chatbot Kotlin sample

- Unhardcode sentiment analizer model name. Add and inject `app.sentiment-analyzer.model-name property`
- Prompt Engineering

- Implement simple session management and error handling for chatbot.
  - Added session ID tracking in the frontend using `localStorage` to maintain continuity between user interactions.
  - Enhanced WebSocket logic, error logging, and proper message handling with session information.
  - Introduced error handling in the backend to return a default response in case of processing issues.

- Updated configurations in `application.properties` and `pom.xml`, including reducing max tokens from 2000 to 1000 and adding macOS-specific dependencies.
- Cleaned up `@V`annotation in Assistant

PR #1496, opened 2025-05-16
Kotlin sample: Refactor chatbot for validation, session handling, and error logging

kpavlov:
# Kotlin sample:  Refactor chatbot for validation, session handling, and error logging

Updated callback scheduling to return detailed error messages. Introduced session ID handling and secure WebSocket connections in the frontend. Enhanced logging, configuration management, and prompt guidance for clarity and improved user experience.

This PR replaces #1495

### DIFF
--- a/samples/chatbot-easy-rag-kotlin/pom.xml
+++ b/samples/chatbot-easy-rag-kotlin/pom.xml
@@ -27,6 +27,13 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
+        <groupId>org.jetbrains.kotlin</groupId>
+        <artifactId>kotlin-bom</artifactId>
+        <version>${kotlin.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
         <groupId>${quarkus.platform.group-id}</groupId>
         <artifactId>${quarkus.platform.artifact-id}</artifactId>
         <version>${quarkus.platform.version}</version>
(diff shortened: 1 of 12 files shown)


samples/chatbot-easy-rag-kotlin/pom.xml
```diff
@@ -26,6 +26,13 @@
 
   <dependencyManagement>
     <dependencies>
+      <dependency>
+        <groupId>org.jetbrains.kotlin</groupId>
+        <artifactId>kotlin-bom</artifactId>
```
kpavlov, line 31: latest kotlin

samples/chatbot-easy-rag-kotlin/pom.xml
```diff
@@ -332,6 +339,25 @@
...
+        <dependency>
+          <groupId>io.netty</groupId>
+          <artifactId>netty-resolver-dns-native-macos</artifactId>
+          <scope>runtime</scope>
+        </dependency><dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-transport-native-kqueue</artifactId>
+        <scope>runtime</scope>
+      </dependency>
+      </dependencies>
```
kpavlov, line 359: to fix slow DNS issue on mac

geoand: We should not that this will cause the sample to not work in native mode with GraalVM. It's a small thing, but we need to be aware

geoand (approved):

samples/chatbot-easy-rag-kotlin/pom.xml
kpavlov: @geoand, true. It might not be specific to this sample, but it might be a much wider issue.

geoand: Right, but this is the only sample we added the native trasport too :)

kpavlov: There are also libraries for linux and windows , e.g. [epool](https://mvnrepository.com/artifact/io.netty/netty-transport-native-epoll)

geoand: I know, but my point is that the mere use of the native transports means that the application will fail to build in native mode.

Anyway, it's a small problem.

If I had any ask, it would be move this transport to a Maven profile that is only activated when the OS is a Mac. Do you mind doing that please?

Issue #1102, opened 2024-11-24
Create a ChatBot codestart


iocanel:
For demo, presentations and workshops I would like to be able to easily create a chatbot from scratch.

Currently, we resort in cloning or copying and pasting.

I would prefer if we had a codestart to generate the chatbot stuff.

PR #1533, opened 2025-05-30
Add description to codestart to avoid inheriting parent description

holly-cummins:
At the moment the codestart looks silly on http://quarkus.io/extensions and http://code.quarkus.io:

image

image

Even with the description, though, I'm a bit confused by this extension. We don't normally create standalone extensions for our codestarts, because the idea is that they'd just be hooked into code.quarkus.io. The other langchain4j codestarts aren't published as extensions (you can see at https://quarkus.io/extensions/?search-regex=codestart this the chatbot codestart is the only codestart published that way). 

I guess maybe we've done it that way because, unlike the EasyRAG codestart, it's a cross-extension codestart? It feels like it's maybe mixing concepts from samples and codestarts, though.

### DIFF
--- a/codestarts/chatbot/pom.xml
+++ b/codestarts/chatbot/pom.xml
@@ -14,6 +14,8 @@
     <packaging>pom</packaging>
     <name>Quarkus Langchain4j - Codestarts - Chatbot - Parent</name>
 
+    <description>A chatbot, similar to the one used in the in quarkus-langchain4j-workshop.</description>
+
     <modules>
         <module>deployment</module>
         <module>runtime</module>


geoand:
> I guess maybe we've done it that way because, unlike the EasyRAG codestart, it's a cross-extension codestart? It feels like it's maybe mixing concepts from samples and codestarts, though.

Not sure, so I'll defer to @ia3andy and @jmartisk

cescoffier (approved):

holly-cummins:
> just noticed it was merged a while ago in #1102
> 
> this isn't a codestart nor a quickstart...we historically did not include full blown apps in codestarts like is done here. This does not scale well if these codestarts popsup on equal footing with extensions. not sure what best action is here imo.

Yeah, I had the same concern and the same uncertainty. :/

PR #1549, opened 2025-06-11
Revamp documentation

cescoffier:


cescoffier:
Ready.

There is a ton of content to add, but at least the existing documentation has been updated and fixed. There is already a lot of new content. 

I'm worried about kroky on the CI, let's see how it goes (as the kroki remote service seems to be down)

cescoffier:
Fixed the build issue (formatting and a few missing properties)

Issue #1575, opened 2025-06-24
Streamable HTTP transport


jmartisk:
Given some peculiarities of the specification, this will be implemented in Quarkus first, before we figure out how to do it in langchain4j itself.

PR #1576, opened 2025-06-24
MCP: Streamable http transport

jmartisk:
Fixes: #1575 

Don't merge yet, we need to make sure that the tests run properly in the CI (they require jbang..)


jmartisk:
Yeah, the CI has passed and the new tests did execute.

integration-tests/mcp/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,15 @@
+<configuration>
```
geoand, line 1: Do we really need logback?

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/http/QuarkusStreamableHttpMcpTransport.java
```diff
@@ -0,0 +1,281 @@
...
+                                            // a response, just mark the operation done when
+                                            // the server accepted it
+                                            future.complete(null);
+                                        }
+                                        response.result().bodyHandler(bodyBuffer -> {
+                                            try {
+                                                String responseString = bodyBuffer.toString();
+                                                JsonNode node = objectMapper.readTree(responseString);
+                                                if (logResponses) {
+                                                    log.info("Response: " + responseString);
```
geoand, line 166: I don't think the logging level should be `INFO` as that will result in spamming the logs

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/http/QuarkusStreamableHttpMcpTransport.java
```diff
@@ -0,0 +1,281 @@
...
+        return uni;
+    }
+
+    private MultivaluedMap<String, Object> toMultivaluedMap(MultiMap multiMap) {
+        MultivaluedTreeMap<String, Object> map = new MultivaluedTreeMap<>();
+        multiMap.forEach((key, value) -> map.add(key, value));
+        return map;
+    }
+
+    // FIXME: this may be brittle, is there a more standard way to parse SSE events?
```
geoand, line 191: We have `org.jboss.resteasy.reactive.client.impl.SseParser`

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/http/QuarkusStreamableHttpMcpTransport.java
jmartisk: I think we talked about this before (maybe it was someone else) but I think it makes sense since `logResponses` is `false` by default, so the user can enable the logs by simply setting the property, and doesn't have to also adjust logging settings... It's in line with how langchain4j does it.

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/http/QuarkusStreamableHttpMcpTransport.java
jmartisk: That one seems to be rather tightly tied to RestEasy clients and depends on having a `SseEventSourceImpl`, which we don't since we're using bare vert.x APIs here,

integration-tests/mcp/src/test/resources/logback-test.xml
jmartisk: Right.. Honestly I'm not sure why this was there (langchain4j contains it). For sure it was meant for controlling the MCP server's logging, but it doesn't seem to change anything after removing it..

integration-tests/mcp/src/test/java/io/quarkiverse/langchain4j/mcp/test/McpPromptsHttpTransportTest.java
```diff
@@ -0,0 +1,49 @@
...
+                    .addAsResource(
+                            new StringAsset(
+                                    """
+                                            quarkus.langchain4j.mcp.client1.transport-type=http
+                                            quarkus.langchain4j.mcp.client1.url=http://localhost:8082/mcp/sse
+                                            quarkus.langchain4j.mcp.client1.log-requests=true
+                                            quarkus.langchain4j.mcp.client1.log-responses=true
+                                            quarkus.log.category."io.quarkiverse".level=DEBUG
+                                            """),
+                            "application.properties"));
```
geoand: Very minor gripe: Using `overrideConfigKey` and `overrideRuntimeConfigKey` make for a slightly more maintainable test. Same for the other tests

jmartisk:
Also @sberyozkin might want to have a look at the use of `McpClientAuthProvider` here. I had to make `AuthInputImpl` package-private to use it.

geoand (approved):

PR #1542, opened 2025-06-06
Add MongoDB support for chat memory

MoheyEl-DinBadr:
Create a MongoDB connection for chat memory

PR #1591, opened 2025-07-04
Add MongoDB support for memorystore

MoheyEl-DinBadr:
- Replaces: #1542


MoheyEl-DinBadr:
I belive the error might be the same as in RedisChatMemoryStoreTest as I can see that it is annotated with  "@Disabled("this only fails in CI and I've spent too much time trying to figure out why...")"

MoheyEl-DinBadr:
@gastaldi Done with the pull request, could you please do the rebase part as I am afraid that the previous error happens again

geoand (approved):
Thanks!

PR #1627, opened 2025-07-21
MCP streamablehttp accept header seprated by comma

nasonawa:
the Accept header separates multiple media types with commas (e.g., application/json, text/event-stream). Semicolons are reserved for parameters within a type (e.g., application/json;q=0.9 for quality factors). Sending a semicolon-separated list is malformed, and strict servers may reject it as a bad request during header parsing—before even attempting to read the body. [Accept Header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept).

- Fixes: #1628


geoand (approved):
Nice catch!

PR #1631, opened 2025-07-21
Check the if id is null or not before starting the SSE channel

nasonawa:
In the [QuarkusStreamableHttpMcpTransport.java#L146](https://github.com/quarkiverse/quarkus-langchain4j/blob/e8e5dc89a8934e6c3f0d2bd81520cda7a8bc5b1d/mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/http/QuarkusStreamableHttpMcpTransport.java#L146)  start the SSE channel only for non null ID responses. otherwise it will end up in the 'java.util.concurrent.TimeoutException'


geoand:
Thanks for this!

Do you know when `id` can be null?

nasonawa:
Thank @geoand , for reviewing this. After successful initialization, the client sends an initialized notification like {"jsonrpc":"2.0","method":"notifications/initialized"} to indicate it is ready to begin normal operations.  This has no id field (equivalent to null), as it's not expecting a reply—it's just signaling to the server that init is complete. If the SSE subscriber or transport code attempts to start/wait on a channel for a response tied to a null/missing id, it will hang indefinitely (leading to the TimeoutException you observed), since no SSE event will arrive for it. 

https://modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle

PR #1643, opened 2025-07-24
Semantic compression guide

cescoffier:


geoand:
This is very nice!

Perhaps you can also add an example interaction that shows the messages before and after the compression? I think it will a long way in showing users the exact value proposition.

cescoffier:
Let me add the "summarized" system message. They are in the log.

cescoffier:
@geoand Done.... even found a bug in my implementation :-)

cescoffier:
Fixed the import order.

geoand (approved):

PR #1660, opened 2025-07-29
Update to LC4j 1.2.0

edeandrea:
Update to upstream LangChain4j 1.2.0


mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/QuarkusMcpToolProvider.java
```diff
@@ -20,9 +21,10 @@
 
 public class QuarkusMcpToolProvider extends McpToolProvider {
 
+    // TODO DefaultMcpResourcesAsToolsPresenter needs to be examined
```
jmartisk, line 24: We may add an option for the app to provide a custom presenter as a bean - this is not super important for now, but... For now, at the very least, we need a way to turn off the presenter altogether. I guess this should be a config property, like `quarkus.langchain4j.mcp.CLIENT_NAME.expose-resources-as-tools`, and if that is false, then you would pass `null` to the `super` constructor

I see you have some failures in your tests that are caused by exactly this - the presenter adds two new tools that the tests don't expect...

jmartisk (changes requested):

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/QuarkusMcpToolProvider.java
jmartisk: And the default value of `expose-resources-as-tools` should be `false` (to align with upstream langchain4j and avoid unexpected surprises for users).

edeandrea:
I'm on PTO this week. I dropped it so I could finish up the guardrails stuff (turns out I didn't need this).

Will probably be a week or so before I get back to this. 

If you'd like to take it from where it is that's fine. If not I'll work on it when I get back.

edeandrea:
Feel free to close this then

Issue #1655, opened 2025-07-28
Tool calling which invokes another vertx runblocking tool might hang forever


tomas1885:
We use Bedrock models, which uses jaxrs client underneath. When using a streaming model with non-streaming model as tool, the non streaming model will causes a deadlock, since the vertx worker used to execute the tool is also used as the underlaying worker for the http client. The main culprit is in the executeOnWorkerThread method in the QuarkusAiServiceStreamingResponseHandler class, which uses an ordered execution. I guess there is or was a motive for using ordered execution but this basically means that a tool that might also use an ordered executeBlocking within the same vertx context will cause this issue.
I don't mind opening a PR, but I think I need to understand the motive for this, I tested removing the ordered flag and it works, but maybe another approach is better (such as a dedicated vertx SharedWorkerExecutor, or Mutiny Infrastructure)?

geoand:
Thanks, I'll try and have a look soon

geoand:
> Do you have a Bedrock credentials

I can generate some keys, it's not an issue. 

> if so I can give some basic code that does this but not as a whole project.

I don't need the whole project, just a sample that I can use to see the problem in action

PR #1661, opened 2025-07-30
Disable the ordered execution of blocking actions

tomas1885:
Disable the ordered execution of blocking actions as it causes a deadlock when the tool itself also requires running on the a blocking action on vertx context.

- Fixes: #1655


geoand:
Oh... The test failures definitely seem related...

tomas1885:
@geoand 
The issue was that the listener calls were also non ordered, I separated it from the tool calling.

Issue #828, opened 2024-08-27
Support Multi<ChatCompletionResponse> AI services 


dastrobu:
When declaring an AI service with signature: 

```java
public interface AiService {
    @SystemMessage("You are a professional poet")
    @UserMessage("""
            Write a poem about {topic}. The poem should be {lines} lines long.
        """)
    Multi<ChatCompletionResponse> writeAPoem(String topic, int lines);
}
```

I get 

```txt
dev.langchain4j.exception.IllegalConfigurationException: Only Multi<String> is supported as a Multi return type. Offending method is 'fooAiService#writeAPoem'
	at dev.langchain4j.exception.IllegalConfigurationException.illegalConfiguration(IllegalConfigurationException.java:12)
	at io.quarkiverse.langchain4j.deployment.AiServicesProcessor.handleDeclarativeServices(AiServicesProcessor.java:442)
```

To be able to access the metadata of the response, such as `usage` or `finishReason` it would be necessary to access the underlying response objects.

geoand:
We can probably add support for that.

geoand:
> To be able to access the metadata of the response, such as usage or finishReason it would be necessary to access the underlying response objects.

How do you plan to use these in a streaming fashion (as `Multi` implies)?

dastrobu:
@geoand you would check on the `finish_reason` of the last event before the `[DONE]` event.

Here is an example of a raw api call (with `max_tokens` set to 1): 

```
data: {"choices":[{"content_filter_results":{"hate":{"filtered":false,"severity":"safe"},"self_harm":{"filtered":false,"severity":"safe"},"sexual":{"filtered":false,"severity":"safe"},"violence":{"filtered":false,"severity":"safe"}},"delta":{"content":"It"},"finish_reason":null,"index":0,"logprobs":null}],"created":1724759298,"id":"chatcmpl-A0oyIz0Qfd9Hq22NwN3VyT7FtgCzb","model":"gpt-4o-2024-05-13","object":"chat.completion.chunk","system_fingerprint":"fp_abc28019ad"}

data: {"choices":[{"content_filter_results":{},"delta":{},"finish_reason":"length","index":0,"logprobs":null}],"created":1724759298,"id":"chatcmpl-A0oyIz0Qfd9Hq22NwN3VyT7FtgCzb","model":"gpt-4o-2024-05-13","object":"chat.completion.chunk","system_fingerprint":"fp_abc28019ad"}

data: [DONE]
```

[...]

PR #1662, opened 2025-07-30
Allow returning a ChatResponse Multi in streaming mode

tomas1885:
Allow returning a ChatResponse Multi in streaming mode

Otherwise wecan't get the metadata and token usage and associate it with a specific request.

Prevent duplicate final message when using Multi<String> return value

Allow using a new ChatEvent producing multi so consumer can get a stream of all chat events.

relates to #828

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
@@ -177,6 +177,10 @@ public class AiServiceMethodImplementationSupport {
         Map<String, Object> templateVariables = getTemplateVariables(methodArgs, methodCreateInfo.getUserMessageInfo());
 
         Type returnType = methodCreateInfo.getReturnType();
+        boolean isMulti = TypeUtil.isMulti(returnType);
+
+        final boolean isStringMulti = (isMulti && returnType instanceof ParameterizedType
+                && TypeUtil.isTypeOf(((ParameterizedType) returnType).getActualTypeArguments()[0], String.class));
         if (TypeUtil.isImage(returnType) || TypeUtil.isResultImage(returnType)) {
             return doImplementGenerateImage(methodCreateInfo, context, systemMessage, userMessage, memoryId, returnType,
                     templateVariables, auditSourceInfo);
@@ -217,7 +221,7 @@ public class AiServiceMethodImplementationSupport {
             Metadata metadata = Metadata.from(userMessage, memoryId, chatMemory);
             AugmentationRequest augmentationRequest = new AugmentationRequest(userMessage, metadata);
 
-            if (!TypeUtil.isMulti(returnType)) {
+            if (!isMulti) {
                 augmentationResult = context.retrievalAugmentor.augment(augmentationRequest);
                 userMessage = (UserMessage) augmentationResult.chatMessage();
             } else {
@@ -244,10 +248,18 @@ public class AiServiceMethodImplementationSupport {
                                 var stream = new TokenStreamMulti(messagesToSend, effectiveToolSpecifications,
                                         finalToolExecutors, ar.contents(), context, memoryId,
                                         methodCreateInfo.isSwitchToWorkerThreadForToolExecution(), isRunningOnWorkerThread);
-                                return stream.plug(m -> ResponseAugmenterSupport.apply(m, methodCreateInfo,
-                                        new ResponseAugmenterParams((UserMessage) augmentedUserMessage,
-                                                memory, ar, methodCreateInfo.getUserMessageTemplate(),
-                                                templateVariables)));
+                                return stream
+                                        .filter(event -> {
+                                            return !isStringMulti || event instanceof ChatEvent.PartialResponseEvent;
+                                        }).map(event -> {
+                                            if (isStringMulti && event instanceof ChatEvent.PartialResponseEvent) {
+                                                return ((ChatEvent.PartialResponseEvent) event).getChunk();
+                                            }
+                                            return event;
+                                        }).plug(m -> ResponseAugmenterSupport.apply(m, methodCreateInfo,
+                                                new ResponseAugmenterParams((UserMessage) augmentedUserMessage,
+                                                        memory, ar, methodCreateInfo.getUserMessageTemplate(),
+                                                        templateVariables)));
                             }
 
                             private List<ChatMessage> messagesToSend(UserMessage augmentedUserMessage,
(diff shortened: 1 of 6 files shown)


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -78,12 +78,7 @@
 import dev.langchain4j.service.tool.ToolProviderRequest;
 import dev.langchain4j.service.tool.ToolProviderResult;
 import dev.langchain4j.spi.ServiceHelper;
-import io.quarkiverse.langchain4j.audit.AuditSourceInfo;
-import io.quarkiverse.langchain4j.audit.InitialMessagesCreatedEvent;
-import io.quarkiverse.langchain4j.audit.LLMInteractionCompleteEvent;
-import io.quarkiverse.langchain4j.audit.LLMInteractionFailureEvent;
-import io.quarkiverse.langchain4j.audit.ResponseFromLLMReceivedEvent;
-import io.quarkiverse.langchain4j.audit.ToolExecutedEvent;
+import io.quarkiverse.langchain4j.audit.*;
```
geoand: Please avoid star imports. The IDE likely did it automatically, but we still try to avoid them nevertheless

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -247,7 +246,15 @@ public Flow.Publisher<?> apply(AugmentationResult ar) {
                                 return stream.plug(m -> ResponseAugmenterSupport.apply(m, methodCreateInfo,
                                         new ResponseAugmenterParams((UserMessage) augmentedUserMessage,
                                                 memory, ar, methodCreateInfo.getUserMessageTemplate(),
-                                                templateVariables)));
+                                                templateVariables)))
+                                        .filter(event -> {
+                                            return !isStringMulti || event instanceof ChatEvent.PartialResponseEvent;
+                                        }).map(event -> {
```
geoand, line 254: How come we need to filter now when we didn't in the past?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -306,7 +313,15 @@ private List<ChatMessage> messagesToSend(UserMessage augmentedUserMessage,
                 return stream.plug(m -> ResponseAugmenterSupport.apply(m, methodCreateInfo,
                         new ResponseAugmenterParams(actualUserMessage,
                                 chatMemory, actualAugmentationResult, methodCreateInfo.getUserMessageTemplate(),
-                                Collections.unmodifiableMap(templateVariables))));
+                                Collections.unmodifiableMap(templateVariables))))
+                        .filter(event -> {
+                            return !isStringMulti || event instanceof ChatEvent.PartialResponseEvent;
+                        }).map(event -> {
```
geoand: Same question here

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/ChatEvent.java
```diff
@@ -0,0 +1,104 @@
+package io.quarkiverse.langchain4j.runtime.aiservice;
+
+import java.util.List;
+
+import dev.langchain4j.model.chat.response.ChatResponse;
+import dev.langchain4j.model.chat.response.ChatResponseMetadata;
+import dev.langchain4j.rag.content.Content;
+import dev.langchain4j.service.tool.ToolExecution;
+
+public class ChatEvent {
```
geoand, line 27: I am not really sure about this... 
Isn't there any domain object in LangChain4j that we could reuse instead of introducing something new?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/GuardrailsSupport.java
```diff
@@ -3,11 +3,7 @@
 import static dev.langchain4j.data.message.UserMessage.userMessage;
 import static io.quarkiverse.langchain4j.guardrails.InputGuardrailParams.rewriteUserMessage;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
geoand: Same comment as above about star imports

geoand (commented):
Thanks for this!

Now I see what you are after.
See my comment about `ChatEvent`

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
tomas1885: Becuase now the Multi also emits the completion event, and when the aiService returns a String Mulit, it doesn't except the whole chat string as the last event, only the chunks.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
tomas1885: Same as the above

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/ChatEvent.java
tomas1885: Haven't seen any object that can wrap all of the tokenStream events. The whole point is to wrap each token stream listener parameter in a container event so it would be propagated through the Multi to the caller.

geoand: Okay, let's go with this

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
geoand: Makes sense, thanks!

geoand:
Thanks a lot!

Two things:

* Please add some documentation about this feature (otherwise very few people will ever know it exists)
* Once done, please squash the commits

tomas1885:
@geoand I added java doc to the class and update the streaming response doc. If you think more should be done please point me in that direction, or feel free to suggest.

geoand:
I would like to see https://github.com/quarkiverse/quarkus-langchain4j/blob/main/docs/modules/ROOT/pages/guide-streamed-responses.adoc updated as well

geoand:
So please squash the commits and force push

geoand (approved):

PR #1664, opened 2025-07-31
Support same method name different tool name

### LINKED ISSUE #1663: Same method name for tool should be allowed if different tool name
This should be allowed, but currently fails:

`    @RegisterAiService
    @ApplicationScoped
    public interface MyChatService {
        @SystemMessage("You are a helpful assistant.")
        @Tool("help chat")
    }

    @RegisterAiService
    @ApplicationScoped
    public interface ChatTools {

        @SystemMessage("You are a helpful assistant.")
        @Tool("help chat")
        String chat(@UserMessage String message, int number);
    }

    @RegisterAiService(tools = { MyChatService.class })
    @ApplicationScoped
    public interface AnotherChatService {

        @SystemMessage("You are a helpful assistant.")
        @Tool("help chat")
        String chat(@UserMessage String message, int number);
    }
`

 ToolProcessor looks at every single @Tool and compares the method names.  This is wrong

* If there are same method names, but different tool names, this should be allowed
* Context matters too.  If there are same method/tool names for @Tool, but they are used in different context this should be allowed

patriot1burke:
Fix #1663


geoand (approved):
Thanks!

geoand (approved):

PR #1691, opened 2025-08-12
Integration tests for exposing resources as tools

jmartisk:


geoand:
@jmartisk can I ask you for a favor? I created the `triage/backport-1.1` label which should be pretty self-explanatory, so if you could add this to fixes you think should be backported, that will be great.

PR #1619, opened 2025-07-17
Refactor upstream guardrails

edeandrea:
Refactor to use upstream guardrails.

For now I'm leaving the existing Quarkus-specific implementation in place, but fully deprecated. We can decide if we'd like to rip it out as part of this PR, or have both implementations for a little bit of time in order to give people time to transition.

Closes #1284

PR #1728, opened 2025-08-28
Refactor upstream guardrails

### LINKED ISSUE #1284: Refactor guardrails once they are moved upstream
I'm planning to move the guardrail implementation [upstream to LangChain4j](langchain4j/langchain4j#2549). Once implemented there I'd like to retrofit the Quarkus implementation to re-use what's done in upstream.

This will take some time of course, so I'm just creating this now as a placeholder for future work.

edeandrea:
Refactor to use upstream guardrails (instead of #1619).

For now I'm leaving the existing Quarkus-specific implementation in place, but fully deprecated. We can decide if we'd like to rip it out as part of this PR, or have both implementations for a little bit of time in order to give people time to transition.

- Closes #1284


geoand:
cc @mariofusco

mariofusco (approved):
@edeandrea thanks a lot, amazing job!

I strongly suggest to remove the old guardrails implementation immediately, or anyway before putting out a new release. For users the migration from one implementation to another will be quite trivial and having both versions around will only confuse them. 

I don't mind if you want to remove the legacy part within this PR or in a subsequent one, that's up to you and both alternatives are ok for me. I just have a small preference in removing the old stuff in this PR because in this way we will have the integration tests running also against those removals and be sure that none of the old parts is still in use.

Also don't forget to update documentation accordingly.

edeandrea:
@mariofusco / @geoand See #1735 for work for removing the Quarkus-specific implementation. I'm at an event this week so I will start working on it next week.

PR #1731, opened 2025-08-29
Bump to LangChain4j 1.4.0

mariofusco:


pom.xml
```diff
@@ -36,10 +36,10 @@
   <properties>
     <quarkus.version>3.20.0</quarkus.version>
     <quarkus.extension-processor.version>${quarkus.version}</quarkus.extension-processor.version>
-    <langchain4j.version>1.3.0</langchain4j.version>
-    <langchain4j-beta.version>1.3.0-beta9</langchain4j-beta.version>
-    <langchain4j-community.version>1.2.0-beta8</langchain4j-community.version>
-    <langchain4j-embeddings.version>1.3.0-beta9</langchain4j-embeddings.version>
+    <langchain4j.version>1.4.0</langchain4j.version>
+    <langchain4j-beta.version>1.4.0-beta10</langchain4j-beta.version>
+    <langchain4j-community.version>1.3.0-beta9</langchain4j-community.version>
```
geoand, line 41: Is this correct?

mariofusco: I did the same also for former releases. That's only because `langchain4j-community` is not out yet, it generally takes 1 or 2 more days and I didn't want to wait. I believe that we could temporarily bring this in as it is and bump also community when it will be available.

geoand: 👍🏽

jmartisk (approved):

PR #1732, opened 2025-08-30
Use deep link for easy rag docs

holly-cummins:
I was a bit disappointed when I followed the guide link for easy rag and it just took me to the top level langchain4j docs. When we have the option and there's an obvious page, should we deep link instead?


geoand (approved):
Thanks!

Issue #1730, opened 2025-08-28
oidc-client-mcp-auth-provider with streamable-http transport type: getAuthorization() tries to block the event-loop thread


btison:
I am trying to use the oidc-client-mcp-auth-provider in combination with streamable-http transport type. 
When the Mcp client is initializing, the `getAuthorization()` method in `OidcClientMcpAuthProvider` tries to block the event-loop thread by calling `Uni.await().indefinitely()`.

This happens when the mcp client is trying to send an 'initialized' request to the mcp server, even if the mcp client is called in an worker pool thread

```
2025-08-28 22:08:55,935 INFO  [io.qua.lan.mcp.run.htt.QuarkusStreamableHttpMcpTransport] (executor-thread-1) Request: {"jsonrpc":"2.0","id":0,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{"roots":{"listChanged":false}},"clientInfo":{"name":"langchai
n4j","version":"1.0"}}}                                                                                                                                                                                                                                                                   
2025-08-28 22:08:56,012 INFO  [io.qua.lan.mcp.run.htt.QuarkusStreamableHttpMcpTransport] (vert.x-eventloop-thread-3) Response: {"jsonrpc":"2.0","id":0,"result":{"capabilities":{"logging":{},"tools":{}},"serverInfo":{"version":"1.0","name":"Mail MCP"},"protocolVersion":"2024-11-05"}
}                                                                                                                                                                                                                                                                                         
2025-08-28 22:08:56,014 INFO  [io.qua.lan.mcp.run.htt.QuarkusStreamableHttpMcpTransport] (vert.x-eventloop-thread-3) Request: {"jsonrpc":"2.0","method":"notifications/initialized"}
```

Stack trace:
```
: jakarta.enterprise.inject.CreationException: Error creating synthetic bean [hsVtFY5C2fd5CgqaJqo8cfZe1YE]: java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.IllegalStateException: The current thread cannot be blocked: vert.x-eventloop-thread-3
        at dev.langchain4j.mcp.client.McpClient_hsVtFY5C2fd5CgqaJqo8cfZe1YE_Synthetic_Bean.doCreate(Unknown Source)
        at dev.langchain4j.mcp.client.McpClient_hsVtFY5C2fd5CgqaJqo8cfZe1YE_Synthetic_Bean.create(Unknown Source)
[...]

PR #1734, opened 2025-09-01
Send the initialized notification on a worker thread

jmartisk:
Fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/1730

### DIFF
--- a/mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/http/QuarkusStreamableHttpMcpTransport.java
+++ b/mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/http/QuarkusStreamableHttpMcpTransport.java
@@ -30,6 +30,7 @@ import dev.langchain4j.mcp.client.transport.McpOperationHandler;
 import dev.langchain4j.mcp.client.transport.McpTransport;
 import io.quarkiverse.langchain4j.mcp.auth.McpClientAuthProvider;
 import io.smallrye.mutiny.Uni;
+import io.smallrye.mutiny.infrastructure.Infrastructure;
 import io.vertx.core.MultiMap;
 import io.vertx.core.http.HttpClient;
 import io.vertx.core.http.HttpMethod;
@@ -70,7 +71,9 @@ public class QuarkusStreamableHttpMcpTransport implements McpTransport {
 
     @Override
     public CompletableFuture<JsonNode> initialize(McpInitializeRequest request) {
-        return execute(request, request.getId()).onItem()
+        return execute(request, request.getId())
+                .emitOn(Infrastructure.getDefaultWorkerPool())
+                .onItem()
                 .transformToUni(
                         response -> execute(new McpInitializationNotification(), null).onItem().transform(ignored -> response))
                 .subscribeAsCompletionStage();


sberyozkin:
@jmartisk Interesting that the SSE transport does not emit on a worker pool.

@btison, can you please try the SSE transport ? It 100% worked for me, but may be I was just a bit lucky...

jmartisk:
IIUC for the legacy transport it uses an event loop but doesn't block it. We use a RestEasy Reactive client which takes care of it internally. In Streamable HTTP, we directly use vert.x apis so we need to move to a worker thread manually, which is what this PR does.

geoand (approved):
Thanks!

I would also be nice to have tests for stuff like this at some point

jmartisk:
There's no API that allows to easily hook into this and see what thread is executing. Is there any more general way for tests how to detect event loop blocking?

jmartisk:
So the way to avoid these surprises would be to always make sure everything is fully rebased before merging.

geoand:
That's the thing, it was theoretically rebased...

jmartisk:
Nope, if you look at the 1.4 upgrade branch from Mario: https://github.com/mariofusco/quarkus-langchain4j/commits/lc4j-140/
Its history doesn't contain the guardrails work. That's the problem, the CI on the 1.4 upgrade PR ran without taking into account the guardrails changes.

jmartisk:
Anyway either @mariofusco or @edeandrea should look into how to fix it :)

sberyozkin:
Thanks for the clarification, @jmartisk , about why it works with the SSE transport

geoand:
@mariofusco can you please take a look?

geoand:
In any case, this is my fault for not being proactive enough about monitoring these two changes closely enough

jmartisk:
Right, I swapped the order of these two PRs in my explanation by error (the 1.4 upgrade was first, guardrails second) but the general explanation stays the same - the latter one (guardrails) wasn't fully rebased on top of the previous one (1.4)

@edeandrea already deleted his guardrails branch so it's hard to check, but I'm looking at the logs of the CI (of the guardrails PR) and it says (in the `actions/checkout` step):

```
  HEAD is now at 40668ab Merge cbb10d77bbfdfffe356f3817f9c172deee61b024 into eebf266c238eac1d732ed2dc7c0194958099fb2d
```

And `eebf266c238eac1d732ed2dc7c0194958099fb2d` is something that came in history BEFORe the 1.4 upgrade, so that confirms my theory.

jmartisk:
The only way to avoid this is to really require PRs to be 'serializable' in the sense that you have to run the CI AND merge the PR only after it is fully rebased, and after **every** merge, **every** open PR has to be rebased and have its CI run again.

geoand:
BTW thanks a lot @jmartisk for digging into the git history and the CI logs!

geoand:
Thanks!

@mariofusco is taking a quick look now, but if it's not trivial, we'll need your input whenever you can check. This isn't too critical, but I do hope to do a `1.2.0.CR1` release sometime tomorrow if possible.

geoand:
All I can say is try it and let's see what happens

edeandrea:
My git history is showing that it contains the 1.4 stuff...

image

geoand:
Please rebase this onto the latest `main`

PR #1737, opened 2025-09-02
Introduce the Agentic module 

geoand:
This is based on the [upstream](https://github.com/langchain4j/langchain4j/pull/3458) `langchain4j-agentic` and is intentionally bare-bones for now


mariofusco (approved):

iocanel (approved):
LGTM

I do have a questio though! Do we really need the `@Agent` annotation ? 
It seems to me that we could just use `Tool` unless of course we want a Service to act as an agent but not really as a Tool.

PR #1742, opened 2025-09-02
Adapt samples to work with platform

fedinskiy:
- Take versions of dependencies from Quarkus Platform BOMs
- Use platform-dependent version of quarkus-maven-plugin (required for work with RHBQ)
- Drop deployment dependencies (it doesn't look like their usage affects the size of the build at all)


geoand:
I actually think that this change is not helpful for people coming to Quarkus LangChain4j for the samples.

geoand:
I don't want the samples to be tied to the latest version on the Quarkus platform - I want them to always reflect the latest Quarkus LangChain4j version.

PR #1743, opened 2025-09-03
Add the cacheToolList property for MCP clients

### LINKED ISSUE #1741: Allow to disable tool cache for MCP clients
Basically, add a Quarkus configuration property for the `cacheToolList` parameter that was recently added via https://github.com/langchain4j/langchain4j/pull/3578

jmartisk:
- Fixes: #1741

### DIFF
--- a/mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/config/McpClientRuntimeConfig.java
+++ b/mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/config/McpClientRuntimeConfig.java
@@ -81,4 +81,12 @@ public interface McpClientRuntimeConfig {
      */
     Optional<String> tlsConfigurationName();
 
+    /**
+     * Whether to cache the tool list obtained from the MCP server.
+     * When set to true (the default), the tool list is cached until the server notifies of changes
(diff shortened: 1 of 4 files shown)


mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/config/McpClientRuntimeConfig.java
```diff
@@ -81,4 +81,12 @@ public interface McpClientRuntimeConfig {
      */
     Optional<String> tlsConfigurationName();
 
+    /**
+     * Whether to cache the tool list obtained from the MCP server.
```
geoand, line 85: This first sentence doesn't seem right

jmartisk: Why not? :open_mouth: I mean, it's not even a sentence - there's no subject, but it looks OK to me.

geoand: It reads awkwardly, but it doesn't matter

Issue #1694, opened 2025-08-13
Missing configuration properties for thinkingConfig for Gemini Chat Models.


hulsbo:
Hello, 

### Issue

Right now the [thinkingConfig](https://ai.google.dev/api/generate-content#ThinkingConfig) API values for [Gemini Chat Models](https://docs.quarkiverse.io/quarkus-langchain4j/dev/gemini-chat-model.html) are not included in the list of configurable application properties.

Regarding the code, the superclass GenerationConfig.java does not refer to a ThinkingConfig.java class at the moment. This class would then hold the gemini API values `thinkingBudget` or `includeThoughts`. 

### Implications with no support

Source: https://ai.google.dev/gemini-api/docs/thinking

- gemini-2.5-flash-lite has thinkingbudget set to 0 as default, disabling thinking. Therefore, this model cannot currently be run in thinking mode.
- Include thoughts in response not possible (default for includeThoughts seems to be `false`).
- Disabling thinking for gemini-2.5-flash not possible (pro model cannot be disabled).
- Setting dynamic/static thinking budgets not possible.

Kind regards,
Oskar

geoand:
Thanks for reporting!

https://github.com/quarkiverse/quarkus-langchain4j/pull/1711 should cover it. Mind giving it a try?

hulsbo:
Hello, sorry for a bit of a delay regarding testing.

I saw you merged already in 1.2.0.CR1 so I tested but unfortunately, neither 

`quarkus.langchain4j.ai.gemini."model-name".chat-model.thinking.thinkingBudget`

or 

`quarkus.langchain4j.ai.gemini.chat-model.thinking.thinkingBudget`

worked for me when I tested requesting the gemini-2.5-flash-lite and gemini-2.5-flash models.

Looking on the Gemini API request logging, the "generationConfig" remains without the "thinkingConfig" in the request.

geoand:
Have you tried using `quarkus.langchain4j.ai.gemini.chat-model.thinking.include-thoughts`?

hulsbo:
Just tried, yes then it works. I can see two issues here:
1. include-thoughts is **not** be an on/off switch for thinking mode, thinking-budget != 0 **is** (as earlier mentioned model flash-lite has this as default). 

Fix: conditional on Line 70 in GeminiChatLanguageModel.java needs to be adjusted to:

```
        if (thinkingBudget != null || includeThoughts) {
[...]

PR #1744, opened 2025-09-04
Patch thought configuration for `quarkus-langchain4j-ai-gemini`

hulsbo:
This patches #1711 which were merged in 1.2.0.CR1. It aims to solve the 2 issues which were noted about it in #1694 by myself.  

Note: due to a persistent build error from another extension of quarkus-langchain (see #1694) I have not been able to test this yet. If someone would like to test (or instruct me how I can fix the build error / test), I'd be grateful. 

## Tests that needs to be done before a merge
- Will a Gemini 2.5 model activate thinking (look for the thinkingConfig object in request logging) when the thinking-budget property is set to a non-null property integer?
- Will the property include-thoughts=true make the returned AIMessage hold the model's thoughts (retrievable via .thinking())? Will the actual LLM response (via .text()) NOT include the thoughts?


geoand (approved):
Thanks!

PR #1758, opened 2025-09-10
Make MCP to work with langchain4j agents

mariofusco:


geoand:
I've fixed most of the problems. No changes are needed to upstream

lordofthejars:
Thanks for this Mario

PR #1794, opened 2025-09-19
Propagate AWS error body in exceptions from AWS Bedrock

yurikotikov:
Include response body in ClientWebApplicationException messages instead of only status info. This makes it easier to debug AWS service errors (e.g. Bedrock) where useful details are returned in the response body.

- add safe readErrorBody(Response) helper
- close response after reading entity
- include body text in exception message


yurikotikov:
@geoand Here is the example how it will look like
```
HTTP 403 Forbidden {"message":"The security token included in the request is expired"}
```

geoand:
Very useful indeed. Is there any chance that message could contain sensitive information?

yurikotikov:
Let me double-check, but what I remember, for Bedrock integration, it's not.

yurikotikov:
@geoand  Is that fall under sensitive info?
```
HTTP 403 Forbidden {"Message":"User: arn:aws:sts::******:assumed-role/spotinst-worker-******-components/******** is not authorized to perform: bedrock:InvokeModel on resource: arn:aws:bedrock:******:*****:inference-profile/eu.amazon.nova-pro-v1:0 because no identity-based policy allows the bedrock:InvokeModel action"}
```

geoand (approved):

PR #1801, opened 2025-09-22
Fix OIDC and OIDC Client McpClientAuthProvider extension names in quarkus-extension.yaml

sberyozkin:


cescoffier (commented):
Why are you using the same name?

sberyozkin:
They are two slightly different extensions, one is `OIDC McpClientAuthProvider` that propagates available tokens with `quarkus-oidc`, the other one is designed to support command line Quarkus LangChain4j applications, called  `OIDC Client McpClientAuthProvider` - it uses `quarkus-oidc-client` to acquire tokens (`client-credentials` grant, etc) itself and propagate them. 
I had hard time figuring out how to name both extensions uniquely :-)
The `description` can be expanded but it also gives a short clarification.

PR #1828, opened 2025-09-29
MCP registry client integration

### LINKED ISSUE #1790: Integration for MCP registry clients
Langchain4j now supports MCP registry clients (see https://github.com/langchain4j/langchain4j/issues/3671). Add some integration (CDI+configuration at least) and tests on the Quarkus side.

jmartisk:
Fixes #1790

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-langchain4j-mcp.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-langchain4j-mcp.adoc
@@ -350,6 +350,137 @@ endif::add-copy-button-to-env-var[]
 |
 
 
+h|[[quarkus-langchain4j-mcp_section_quarkus-langchain4j-mcp-registry-client]] [.section-name.section-level0]##link:#quarkus-langchain4j-mcp_section_quarkus-langchain4j-mcp-registry-client[Configured MCP registry clients]##
+h|Type
+h|Default
+
+a| [[quarkus-langchain4j-mcp_quarkus-langchain4j-mcp-registry-client-registry-client-name-base-url]] [.property-path]##link:#quarkus-langchain4j-mcp_quarkus-langchain4j-mcp-registry-client-registry-client-name-base-url[`quarkus.langchain4j.mcp.registry-client."registry-client-name".base-url`]##
+ifdef::add-copy-button-to-config-props[]
+config_property_copy_button:+++quarkus.langchain4j.mcp.registry-client."registry-client-name".base-url+++[]
+endif::add-copy-button-to-config-props[]
+
+
+[.description]
+--
+The base URL of the MCP registry, without the API version segment. The default value points at the official registry (https://registry.modelcontextprotocol.io).
+
+
+ifdef::add-copy-button-to-env-var[]
+Environment variable: env_var_with_copy_button:+++QUARKUS_LANGCHAIN4J_MCP_REGISTRY_CLIENT__REGISTRY_CLIENT_NAME__BASE_URL+++[]
+endif::add-copy-button-to-env-var[]
+ifndef::add-copy-button-to-env-var[]
+Environment variable: `+++QUARKUS_LANGCHAIN4J_MCP_REGISTRY_CLIENT__REGISTRY_CLIENT_NAME__BASE_URL+++`
+endif::add-copy-button-to-env-var[]
+--
+|string
+|`https://registry.modelcontextprotocol.io`
+
+a| [[quarkus-langchain4j-mcp_quarkus-langchain4j-mcp-registry-client-registry-client-name-log-requests]] [.property-path]##link:#quarkus-langchain4j-mcp_quarkus-langchain4j-mcp-registry-client-registry-client-name-log-requests[`quarkus.langchain4j.mcp.registry-client."registry-client-name".log-requests`]##
+ifdef::add-copy-button-to-config-props[]
+config_property_copy_button:+++quarkus.langchain4j.mcp.registry-client."registry-client-name".log-requests+++[]
+endif::add-copy-button-to-config-props[]
+
+
+[.description]
+--
+Whether to log requests
+
+
+ifdef::add-copy-button-to-env-var[]
+Environment variable: env_var_with_copy_button:+++QUARKUS_LANGCHAIN4J_MCP_REGISTRY_CLIENT__REGISTRY_CLIENT_NAME__LOG_REQUESTS+++[]
+endif::add-copy-button-to-env-var[]
+ifndef::add-copy-button-to-env-var[]
+Environment variable: `+++QUARKUS_LANGCHAIN4J_MCP_REGISTRY_CLIENT__REGISTRY_CLIENT_NAME__LOG_REQUESTS+++`
+endif::add-copy-button-to-env-var[]
+--
+|boolean
+|`false`
+
+a| [[quarkus-langchain4j-mcp_quarkus-langchain4j-mcp-registry-client-registry-client-name-log-responses]] [.property-path]##link:#quarkus-langchain4j-mcp_quarkus-langchain4j-mcp-registry-client-registry-client-name-log-responses[`quarkus.langchain4j.mcp.registry-client."registry-client-name".log-responses`]##
+ifdef::add-copy-button-to-config-props[]
+config_property_copy_button:+++quarkus.langchain4j.mcp.registry-client."registry-client-name".log-responses+++[]
+endif::add-copy-button-to-config-props[]
+
+
+[.description]
+--
+Whether to log responses
+
+
+ifdef::add-copy-button-to-env-var[]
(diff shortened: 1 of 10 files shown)


mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/config/McpRegistryClientBuildTimeConfig.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.Optional;
+
+import io.quarkus.runtime.annotations.ConfigDocIgnore;
+import io.quarkus.runtime.annotations.ConfigGroup;
+
+/**
+ * Empty right now. Necessary to fix the list of registry client configuration keys at build time.
+ */
+@ConfigGroup
+public interface McpRegistryClientBuildTimeConfig {
```
jmartisk: @geoand  maybe you can advise how to get rid of this class. But I had to add "some" build time config to be able to access a map of the client configuration keys during build (to register them in CDI). I don't know if there's a better solution than an empty build-time config class.

geoand: Ah! I don't know TBH, I've not encountered this. @radcortez what is the best way to handle this situation?

jmartisk: Roberto is off today, so depending on how soon you want to do the final release, we may want to merge this with this hack

geoand: IIUC, we are forcing users to set this, correct? If so, I really don't want the hack

radcortez: I'm working today. Let me have a look.

jmartisk: No, the users don't have to set anything. I had to include an actual config property because otherwise the `Map<String, McpRegistryClientBuildTimeConfig>` appears empty in the processor, but I made that property optional (and marked it with `@ConfigDocIgnore`).

radcortez: The issue here is that at runtime, you may have keys not available at build time (if the user sets the registry client configurations in a runtime-only source).

geoand: But @jmartisk only wants the "structure", not the actual values, right @jmartisk ?

radcortez: Yes, but you don't get keys without the values either. If you only want the keys we could declare a simple `Map<String, String>`, but that is not the problem.

If, as a user, I'm only required to set `quarkus.langchain4j.mcp.registry-client.CLIENT1.base`, and that is a runtime property, I don't have any guarantees that this key will be available at build time.

radcortez: Is it required to have `@McpRegistryClientName("CLIENT1")`, where I know that the keys that show up in the configuration must match with the annotation?

jmartisk: I would like the values to be runtime config, but I need the list of keys at build time - to register them in CDI (I do know that this means you can't add new keys at runtime, but I guess that will simply be a limitation)

> Is it required to have `@McpRegistryClientName("CLIENT1")`, where I know that the keys that show up in the configuration must match with the annotation?

that's why I want the keys available at build time, to be able to register the beans with the `@McpRegistryClientName("CLIENT1")` qualifier

radcortez: And what about getting the keys from the annotations instead of Config?

geoand: I do think we do that for Mongo Client for example

radcortez: What we do in the REST Client is to use a `@WithKeys` annotation that provides the registered keys:

https://github.com/quarkusio/quarkus/blob/593d38bb2a4938ceaa5a41a1464cee5f06a4e8ea/extensions/resteasy-classic/rest-client-config/runtime/src/main/java/io/quarkus/restclient/config/RestClientsBuildTimeConfig.java#L24-L28

https://github.com/quarkusio/quarkus/blob/593d38bb2a4938ceaa5a41a1464cee5f06a4e8ea/extensions/resteasy-classic/rest-client-config/runtime/src/main/java/io/quarkus/restclient/config/RestClientKeysProvider.java#L7-L14

https://github.com/quarkusio/quarkus/blob/593d38bb2a4938ceaa5a41a1464cee5f06a4e8ea/extensions/resteasy-classic/rest-client-config/deployment/src/main/java/io/quarkus/restclient/config/deployment/RestClientConfigUtils.java#L25-L60

This generates the provided keys, which are available in the annotation. In this case, this has some benefits:

- Because the keys are known, we don't have to iterate the list of names to search for the possible keys
- It would work for sources that don't provide the list of keys (like Vault)
- You can easily validate if the annotation is missing the config or vice versa.

In the case of the REST Client, it also has significant benefits in the rewrite of the relocations / fallbacks, because we have multiple keys for the same configuration, but that is not the case here.

We can use the same pattern here, but it may be overkill.

jmartisk: That would become a problem if somebody accesses the client only programmatically (and so there's no injection point with an annotation). And if somebody accidentally makes a typo on one side (the name in the annotation and the configuration key won't be the same), then we won't be able to detect it (and they will inject a badly configured client).

jmartisk: For regular MCP clients, we do it the same way that this PR proposes for registry clients, except that there is one actual build-time property present (the `transport-type`). Perhaps we should keep the approaches unified..

radcortez: > That would become a problem if somebody accesses the client only programmatically (and so there's no injection point with an annotation).

You mean with `CDI.` or `Arc.`? Or some other way?

radcortez: Anyway, we are discussing two different problems here:

- The original one is how to get rid of that subgroup. If we know all the keys are available, you can use a simple `Map<String, Map<String, String>>`. Why a double Map? Because Config, will take the path and adjust the rest of the segment to the type being mapped, so something like `quarkus.langchain4j.mcp.registry-client.CLIENT1.base`, will store `CLIENT1.base` as a key in a single `Map<String, String>`. With `Map<String, Map<String, String>>`, it will use the next segment of the path for the first `Map` key and adjust the remaining segments to the second `Map`. Since you only want the keys, this would be a way to do it.

- As for the second issue, I'll let you guys decide what is acceptable in this situation.

jmartisk: Ok I've pushed an update that changes the config to `Map<String, Map<String, String>>` and removes the extra interface

jmartisk: Now there's a totally weird error when generating documentation... ugh

geoand: :(

jmartisk: Ah, that was caused by a missing javadoc comment (though the error message was completely useless)

geoand (approved):

PR #1829, opened 2025-09-29
Trace tool calls

sgdesmet:
This PR adds the following attributes to the span when executing tool calls (as defined [here](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md#execute-tool-span)): 

- `gen_ai.operation.name`
- `gen_ai.tool.call.id`
- `gen_ai.tool.name`
- `gen_ai.tool.type`
- `gen_ai.tool.call.arguments`
- `gen_ai.tool.call.result`

This improves tracing tool calls when using for example Langfuse.


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolSpanWrapper.java
```diff
@@ -2,31 +2,55 @@
...
 
     @Override
     public String wrap(ToolExecutionRequest toolExecutionRequest, Object memoryId,
-            BiFunction<ToolExecutionRequest, Object, String> fun) {
-        Span span = tracer.spanBuilder("langchain4j.tools." + toolExecutionRequest.name()).startSpan();
+                       BiFunction<ToolExecutionRequest, Object, String> fun) {
+
+        // from https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-spans.md#execute-tool-span
+        Span span = tracer.spanBuilder( "langchain4j.tools." + toolExecutionRequest.name() )
+                .setSpanKind( SpanKind.INTERNAL )
```
geoand: Can you explain this, please?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolSpanWrapper.java
```diff
@@ -2,31 +2,55 @@
...
+                .setAttribute( "gen_ai.operation.name", "execute_tool" )
+                .setAttribute( "gen_ai.tool.call.id", toolExecutionRequest.id() )
+                .setAttribute( "gen_ai.tool.name", toolExecutionRequest.name() )
+                .setAttribute( "gen_ai.tool.type", "function" )
+                .startSpan();
+        if (includeArguments) {
+            span.setAttribute( "gen_ai.tool.call.arguments", toolExecutionRequest.arguments() );
```
geoand: How does this serialize the arguments?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolSpanWrapper.java
```diff
@@ -2,31 +2,55 @@
...
+        }
         try (Scope scope = span.makeCurrent()) {
             // TODO Handle async method here.
-            return fun.apply(toolExecutionRequest, memoryId);
+            var result = fun.apply( toolExecutionRequest, memoryId );
+            if (includeResult) {
+                span.setAttribute( "gen_ai.tool.call.result", result );
```
geoand: Same question here

geoand:
Thanks a lot for this! I added a couple of questions

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolSpanWrapper.java
sgdesmet: The linked documentation states that tool execution spans should have the `INTERNAL` span kind, so I'm following this convention even though `CLIENT` (or `SERVER`) might seem to make more sense. Looking at the documentation for `io.opentelemetry.api.trace.SpanKind`, `INTERNAL` is the default so the call could be omitted. I currently left it in to make it more explicit.

geoand: Thanks. @brunobat can you comment on this please?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolSpanWrapper.java
sgdesmet: `arguments()` returns a single String, so there is no need to do so here. Skimming through the various implementations in langchain4j this always appears to contain the arguments serialized to JSON.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolSpanWrapper.java
sgdesmet: This also always returns a `String` containing JSON

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolSpanWrapper.java
brunobat: It's ok. The spec says those spans must be `INTERNAL`, which does make sense because they happen while being processed internally by the server.

geoand: 🙏🏽

geoand:
Mind rebasing the PR onto `main`?

sgdesmet:
> Mind rebasing the PR onto `main`?

No problem, done

geoand:
Ah sorry, can you also squash the commits?

geoand:
I rebased and forced pushed so we get the CI fix

PR #1834, opened 2025-10-01
fix: disable topK for anthropic streaming model

Wal8800:
We disabled topK when thinking is enabled for anthropic model

https://github.com/quarkiverse/quarkus-langchain4j/pull/1813

We need to apply the same logic for streaming model as well.


geoand:
The failures are unrelated to the change, @jmartisk is on it

PR #1836, opened 2025-10-01
Use more precise guide link for Ollama

holly-cummins:
The current link just links to the top level and makes users search.


geoand (approved):
Thanks!

PR #1846, opened 2025-10-08
Implement upstream LC4j observability capabilities

### LINKED ISSUE #1844: Implement upstream LC4j AI Service observability
Implement the upstream [LangChain4j AI Service observability](https://docs.langchain4j.dev/tutorials/observability#ai-service-observability) here.

edeandrea:
This implements the new upstream LangChain4j observability capabilities here.

Fixes #1844


edeandrea:
The (re-) implementation here does not require the users to create listener classes. Instead they continue to use CDI observers, but they can use a qualifier to narrow the event to a particular ai service.

Issue #1849, opened 2025-10-09
Tavily Bad Request


Joshua-Hell:
Tavily returns an `Error 400 Bad Request` using when not specifying `include_domains` or `exclude_domains`. An empty list wouldn't be a problem normally, but the config mapping sets both lists as `[]` strings, so the payload that gets sent is:
```
body: {"api_key":"tvly-de**********************************","query":"Naddl bio","search_depth":"basic","include_answer":false,"include_raw_content":false,"max_results":5,"include_domains":["[]"],"exclude_domains":["[]"]}
```

The response to that is:
```
{"detail":{"error":"All domains in include_domains are invalid: ['[]']. Domains must have a valid suffix (e.g., 'tavily.com' instead of 'tavily')."}}
```

Seems like `[]` is an invalid domain, xD.

[I have an repository to reproduce it](https://github.com/Joshua-Hell/minimal-tavily-example), it uses a local mistral model using Ollama, but you can use every other model/provider.
You also have to register at Tavily to get a free API key.

The problem in the code seems that the client is encoding the default value of `[]` wrong, not as an empty list, but as a string, which is then the only element of the list.

If wanted I can directly work on this issue.

PR #1851, opened 2025-10-10
Default empty list for Tavily include and exclude domains

Joshua-Hell:
Config changes for the Tavily web search tool to have a empty list as default and not one filled with `[]` as the only string.

- Fixes: #1849

### DIFF
--- a/tools/tavily/runtime/src/main/java/io/quarkiverse/langchain4j/tavily/runtime/TavilyConfig.java
+++ b/tools/tavily/runtime/src/main/java/io/quarkiverse/langchain4j/tavily/runtime/TavilyConfig.java
@@ -75,13 +75,13 @@ public interface TavilyConfig {
     boolean includeRawContent();
 
     /**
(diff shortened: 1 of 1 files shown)


tools/tavily/runtime/src/main/java/io/quarkiverse/langchain4j/tavily/runtime/TavilyConfig.java
```diff
@@ -77,13 +77,13 @@ public interface TavilyConfig {
     /**
      * A list of domains to specifically include in the search results. Default is [], which includes all domains.
```
jmartisk: Could we update the javadoc to also avoid saying `[]`? It may be a bit confusing. It could say that by default, all domains are included. Similarly for excluded domains - by default, no domains are excluded.

joshua-hell: Thx for the feedback, overlooked that.

PR #1858, opened 2025-10-15
Introduce compile time safety to the agentic module

geoand:
We will now check to make sure that declarative API conforms to all the requirements

### DIFF
--- a/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
+++ b/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
@@ -1,5 +1,10 @@
 package io.quarkiverse.langchain4j.agentic.deployment;
 
+import static io.quarkiverse.langchain4j.agentic.deployment.ValidationUtil.validateAllowedReturnTypes;
+import static io.quarkiverse.langchain4j.agentic.deployment.ValidationUtil.validateNoMethodParameters;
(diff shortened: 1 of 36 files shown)


agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
```diff
@@ -62,18 +71,292 @@ void detectAgents(CombinedIndexBuildItem indexBuildItem, BuildProducer<DetectedA
...
+    private static void validateActivationCondition(ClassInfo iface) {
+        DotName annotationToValidate = AgenticLangChain4jDotNames.ACTIVATION_CONDITION;
+        List<AnnotationInstance> instances = iface.annotations(annotationToValidate);
+        for (AnnotationInstance instance : instances) {
+            if (instance.target().kind() != AnnotationTarget.Kind.METHOD) {
+                log.warnf("Unhandled '@%s' annotation: '%s'", annotationToValidate.withoutPackagePrefix(), instance.target());
+                continue;
+            }
+            MethodInfo method = instance.target().asMethod();
+            validateStaticMethod(method, AgenticLangChain4jDotNames.AFTER_AGENT_INVOCATION);
```
mariofusco: I don't think this is correct?

geoand: Indeed! Nice catch!

Issue #1874, opened 2025-10-23
AiMessage `thinking` field not persisted in Redis ChatMemoryStore


ejstuart:
## Summary

When using the Anthropic streaming client with extended thinking enabled and Redis memory store, the `thinking` field in `AiMessage` is not properly persisted. The thinking content is correctly captured during streaming and serialized to Redis, but it is lost during deserialization when retrieving messages from Redis.

## Root Cause

The `ObjectMapper` in `QuarkusJsonCodecFactory.ObjectMapperHolder` is missing the Jackson mixins required to properly deserialize `AiMessage` objects with the `thinking` field.

**Affected File:**
`core/runtime/src/main/java/io/quarkiverse/langchain4j/QuarkusJsonCodecFactory.java`

The upstream `langchain4j-core` library defines these mixins in `JacksonChatMessageJsonCodec.java`:

```java
@JsonInclude(NON_NULL)
@JsonDeserialize(builder = AiMessage.Builder.class)
private abstract static class AiMessageMixin {}
```

## Impact
Thinking not passed back for extended conversations, this breaks the anthropic usecase as it expects to see thinking information.

## Workaround
For custom `RedisChatMemoryStore` implementations, create a custom `ObjectMapper` with the proper mixins:

```java
import dev.langchain4j.data.message.JacksonChatMessageJsonCodec;
import com.fasterxml.jackson.databind.ObjectMapper;
import com.fasterxml.jackson.core.json.JsonReadFeature;

@ApplicationScoped
public class RedisChatMemoryStore implements ChatMemoryStore {
    // Use custom ObjectMapper with proper mixins instead of QuarkusJsonCodecFactory.ObjectMapperHolder.MAPPER
    private static final ObjectMapper CHAT_MESSAGE_MAPPER =
        JacksonChatMessageJsonCodec.chatMessageJsonMapperBuilder()
            .configure(JsonReadFeature.ALLOW_UNESCAPED_CONTROL_CHARS.mappedFeature(), true)
            .build();
     ...
}
```

geoand:
Great catch!

From what I can tell, you already have a fix in the works :). We'd gladly accept a PR

PR #1884, opened 2025-10-24
Use upstream LangChain4j mixins for Jackson codec

ejstuart:
- Closes: https://github.com/quarkiverse/quarkus-langchain4j/issues/1874


ejstuart:
Oh whoops, looks like some tests are failing because the order of the deserialisation is different now, will fix

PR #1888, opened 2025-10-27
Bump to LangChain4j 1.8.0

mariofusco:


geoand (approved):
Thanks!

mcp/deployment/src/test/java/io/quarkiverse/langchain4j/mcp/test/registryclient/McpRegistryClientTest.java
```diff
@@ -81,8 +80,8 @@ public void testListServersLimit() {
 
     @Test
     public void testGetServer() {
-        McpServer server = client.getServerDetails(
-                "86863c74-2ae5-4430-8880-5474e7ae2155"); // this is io.github.bytedance/mcp-server-filesystem
+        McpGetServerResponse server = client.getSpecificServerVersion(
+                "86863c74-2ae5-4430-8880-5474e7ae2155", "latest"); // this is io.github.bytedance/mcp-server-filesystem
```
jmartisk: Does this really work? The name argument here should be the server name (not UUID) - like we have in the upstream https://github.com/langchain4j/langchain4j/blob/1.8.0/langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/registryclient/McpRegistryClientIT.java#L80

Could you move over both methods `testGetSpecificServerVersion` and `testGetAllVersionsOfServer` (and replace `testGetServer` with them) to be aligned with what is upstream? Generally I would like the testcases to be exactly the same (except of course this one will use the Quarkus way to configure and inject the client)

mariofusco: I followed your suggestion and now the test looks identical to what langchain4j does. I'm not sure if I'm not missing something else though, please double check again.

jmartisk: Oooooh I see this test is still disabled by an annotation, please enable it :) maybe that's why it looked like it was passing even with the wrong argument

jmartisk: And you still don't have `testGetAllVersionsOfServer` here. And in `testListServersUpdatedSince`, I think you're missing a `version("latest")` call

mariofusco: Done.

Issue #1869, opened 2025-10-21
Throwing IllegalArgumentException from the tool breaks streaming response handler


Wal8800:
When I have a tool that can throw IllegalArgumentException, for example

```java
    @Tool("")
    public String testTool() {
        throw new IllegalArgumentException("hello");
    }
```

The IllegalArgumentException is not captured as a failed ToolExecutionResult. The exception is re-throw.

https://github.com/quarkiverse/quarkus-langchain4j/blob/main/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/QuarkusToolExecutor.java#L127

Causing the tool execution runnable task to exit because we don't have exception handling on the `execute`

https://github.com/quarkiverse/quarkus-langchain4j/blob/main/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceStreamingResponseHandler.java#L259

Consequently, no more response are sent to the stream.

For the client, the streaming response stops mid-stream and no error message is shown.

geoand:
Thanks a lot for catching this!

As you've already analyzed the code, would you be interested in opening a pull request?

jmartisk:
I haven't checked this too deep so sorry if I'm off-topic, but I guess we should adjust the `QuarkusToolExecutor` to be consistent with the upstream one where, after recent changes, it is configurable whether execution exceptions should be propagated as a Java exception, or turned into an error message in the result - see https://github.com/langchain4j/langchain4j/blob/1.7.1/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java#L114

PR #1904, opened 2025-10-29
Fix exception rethrow in tool executor

ejstuart:
Went with the [example I posted](https://github.com/quarkiverse/quarkus-langchain4j/issues/1869#issuecomment-3460277220) in the issue's thread, this is to maintain the framework's error as properly propagated while not rethrowing exceptions defined by users in their tools.

- Closes: https://github.com/quarkiverse/quarkus-langchain4j/issues/1869

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/QuarkusToolExecutor.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/QuarkusToolExecutor.java
@@ -109,7 +109,7 @@ public class QuarkusToolExecutor implements ToolExecutor {
             String result;
             if (invocationResult instanceof Uni<?>) { // TODO CS
                 if (io.vertx.core.Context.isOnEventLoopThread()) {
-                    throw new IllegalStateException(
+                    throw new ToolExecutionException(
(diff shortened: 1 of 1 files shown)


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/ToolsRecorder.java
```diff
@@ -53,7 +53,7 @@ public static void populateToolMetadata(Collection<Object> objectsWithTools, Lis
                 QuarkusToolExecutor.Context executorContext = new QuarkusToolExecutor.Context(objectWithTool,
                         invokerClassName, methodCreateInfo.methodName(),
                         methodCreateInfo.argumentMapperClassName(), methodCreateInfo.executionModel(),
-                        methodCreateInfo.returnBehavior());
+                        methodCreateInfo.returnBehavior(), false);
```
jmartisk, line 56: I guess this should be configurable, not just hardcoded to `false`. But that's something I can have a look at in a subsequent PR, so no worries.

geoand:
Heads up: I am going to align with upstream LangChain4j on error handling with https://github.com/quarkiverse/quarkus-langchain4j/pull/2014

PR #1923, opened 2025-11-09
Fire intermediate chat event at the correct spot

ejstuart:
When I initially implemented this I accidentally fired the intermediate Response from the wrong place.
While not necessarily breaking anything it does cause quite a delay when there's a significant number of tools to execute.


geoand (approved):
Thanks!

PR #1926, opened 2025-11-10
Scope google credentials to cloud platform by default

ramesh-spendflo:
This PR fixes issue #1928


sberyozkin:
Thanks @ramesh-spendflo, usually the issue is closed once the PR is merged.

Hmm, once thing I did not realize is that it is also shared here with the AiGemini model which is different to the Vertex Ai Gemini.

It looks like this default model authentication provider should be moved out of the common module, kept as is in the ai-gemini but updated as per this PR for Vertex AI Gemini

sberyozkin:
@ramesh-spendflo Sure, thanks. Are you OK with updating this PR ? Please move the current default `ModelAuthProvider` out of the common gemini module to the ai-gemini one, and have another added to the vertex ai gemini model provider code as per this PR

sberyozkin:
@ramesh-spendflo Thanks, did you try both Vertex AI Gemini and Ai Gemini model providers ?

sberyozkin:
Thanks for your help @ramesh-spendflo, we can also make it configurable later if necessary.
LGTM, CC @geoand

PR #1937, opened 2025-11-13
Fix GPU resources utilization

### LINKED ISSUE #1935: Fine-grain resource management for GPU model provider
Currently, the `GPULlama3Processor.generateBeans` method initializes *two* instances of GPULlama3 (one for `ChatModel`, one for `StreamingChatModel`) when GPULlama3 is deployed as the model provider during the initialization of a Quarkus application. Even though GPULlama3 has been integrated following a similar approach with other pre-existing quarkus-langchain4j model providers, in its case this is rather inefficient and a waste of resources as these two instances occupy GPU memory of the local machine which is limited for most workstations/laptops. How should we refactor the GPULlama3 integration in order to avoid this situation? 

In addition to the above, it seems that fine-grained control on resources (GPULlama3 ChatModel or StreamingChatModel instances) deallocation is needed especially for more complex use-cases.

Any suggestions on how to approach those issues and refactor GPULlama3 integration?

orionpapadakis:
This PR fixes #1935 by implementing lazy model initialization to prevent excessive GPU memory allocation during application startup.

## Problem
During Quarkus LangChain4j initialization, both `chatModel` and `streamingChatModel` beans are registered regardless of application requirements. This results in two GPULlama3 instances being instantiated immediately - one for each model type - without knowing which (if any) the application will actually use. Since each instance consumes significant GPU memory, which is typically limited on workstations and laptops, this represents inefficient resource utilization.

## Solution
This PR introduces lazy initialization for GPULlama3 models:

- **Lazy Factory Methods**: Added `createLazy()` factory methods to both `GPULlama3ChatModel` and `GPULlama3StreamingChatModel`
- **Deferred Initialization**: Model loading, TornadoVM setup, and GPU resource allocation are postponed until the first `doChat()` call
- **Preserved Bean Registration**: Both CDI beans are still registered as usual, maintaining compatibility with existing applications
- **On-Demand Instantiation**: Only the model type actually used by the application consumes GPU memory

## Benefits
- **Efficient Resource Usage**: Applications now consume GPU memory only for models they actually use
- **Flexible Deployment**: Enables deployment of multiple model configurations without upfront resource commitment
- **Improved Startup Performance**: Eliminates expensive GPU initialization from application startup path
- **Backward Compatibility**: Existing applications continue to work without modifications

- Fixes: #1935

### DIFF
--- a/model-providers/gpu-llama3/runtime/src/main/java/io/quarkiverse/langchain4j/gpullama3/GPULlama3ChatModel.java
+++ b/model-providers/gpu-llama3/runtime/src/main/java/io/quarkiverse/langchain4j/gpullama3/GPULlama3ChatModel.java
@@ -7,6 +7,8 @@ import java.io.UncheckedIOException;
 import java.nio.file.Path;
 import java.util.Optional;
 
+import org.jboss.logging.Logger;
+
 import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.internal.ChatRequestValidationUtils;
 import dev.langchain4j.model.chat.ChatModel;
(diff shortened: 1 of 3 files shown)


geoand (approved):
Thanks!

PR #1943, opened 2025-11-18
Ignore empty SSE messages

jmartisk:


jmartisk:
Hold on I'll fix this together with something else

Issue #1951, opened 2025-11-21
QuarkusStreamableHttpMcpTransport doesn't renew a lost Mcp-Session-Id


frami:
In the LangChain4j `StreamableHttpMcpTransport` class they fixed the bug of renew a lost MCP-Session-Id. Look at: https://github.com/langchain4j/langchain4j/pull/3833
But in the `QuarkusStreamableHttpMcpTransport` implementation of `McpTransport` is this bug fixed.

```
Caused by: java.util.concurrent.ExecutionException: java.lang.RuntimeException: Unexpected status code: 404
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2096)
	at io.smallrye.context.CompletableFutureWrapper.get(CompletableFutureWrapper.java:161)
	at dev.langchain4j.mcp.client.DefaultMcpClient.readResource(DefaultMcpClient.java:284)
	... 23 more
Caused by: java.lang.RuntimeException: Unexpected status code: 404
	at io.quarkiverse.langchain4j.mcp.runtime.http.QuarkusStreamableHttpMcpTransport.lambda$execute$4(QuarkusStreamableHttpMcpTransport.java:207)
	at io.vertx.core.impl.future.FutureImpl$4.onSuccess(FutureImpl.java:176)
	at io.vertx.core.impl.future.FutureBase.emitSuccess(FutureBase.java:66)
	at io.vertx.core.impl.future.FutureImpl.tryComplete(FutureImpl.java:259)
	at io.vertx.core.Promise.complete(Promise.java:66)
	at io.vertx.core.http.impl.HttpClientRequestImpl.handleResponse(HttpClientRequestImpl.java:389)
	at io.vertx.core.http.impl.HttpClientRequestBase.handleResponse(HttpClientRequestBase.java:204)
	at io.vertx.core.http.impl.HttpClientRequestBase.lambda$new$0(HttpClientRequestBase.java:69)
	at io.vertx.core.impl.ContextImpl.emit(ContextImpl.java:342)
	at io.vertx.core.impl.ContextImpl.lambda$emit$8(ContextImpl.java:349)
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:173)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:166)
[...]

PR #1958, opened 2025-11-24
feat: Added support to auto reinitialize session if not found

frami:
**Issue**
Closes https://github.com/quarkiverse/quarkus-langchain4j/issues/1951

**Change**
This PR adds support to automatically reinitialize the session with the MCP server if it is not found (e.g., after a server restart). This ensures that clients can recover gracefully without manual intervention or a full restart.

Same reinitialize implementation as it was done for class StreamableHttpMcpTransport with PR https://github.com/langchain4j/langchain4j/pull/3833


geoand:
The code in the PR needs to be formatted before being committed - this can be done simply by building the changed modules

Issue #1960, opened 2025-11-25
Quarkus Ecosystem CI failing for 8+ days - McpRegistryClientTest errors


geoand:
The Quarkus Ecosystem CI has been consistently failing for the past 8+ days with errors in `McpRegistryClientTest`. All 6 tests in this class are failing with synthetic bean creation errors.

## Failed Runs

- [Nov 25, 2025](https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/19655300149) - FAILURE
- [Nov 24, 2025](https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/19620787377) - FAILURE  
- [Nov 21, 2025](https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/19557075756) - FAILURE
- [Nov 20, 2025](https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/19522406588) - FAILURE
- [Nov 19, 2025](https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/19486719578) - FAILURE
- [Nov 18, 2025](https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/19450899605) - FAILURE
- [Nov 17, 2025](https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/19415709480) - FAILURE

## Failing Tests

All 6 tests in `io.quarkiverse.langchain4j.mcp.test.registryclient.McpRegistryClientTest`:
- `testPing`
- `testGetAllVersionsOfServer`
- `testGetSpecificServerVersion`  
- `testListServers`
- `testListServersLimit`
- `testHealth`

## Example Stacktrace

```
[ERROR] io.quarkiverse.langchain4j.mcp.test.registryclient.McpRegistryClientTest.testPing -- Time elapsed: 0.077 s <<< ERROR!
jakarta.enterprise.inject.CreationException: Error creating synthetic bean [3yCvcoNgEDWIfJAIpr2ddlhf_Es]: java.lang.IllegalStateException: this should not be called
	at dev.langchain4j.mcp.registryclient.McpRegistryClient_3yCvcoNgEDWIfJAIpr2ddlhf_Es_Synthetic_Bean.create(Unknown Source)
	at dev.langchain4j.mcp.registryclient.McpRegistryClient_3yCvcoNgEDWIfJAIpr2ddlhf_Es_Synthetic_Bean.create(Unknown Source)
[...]

PR #1963, opened 2025-11-25
Bump the Quarkus dependency to 3.27.1

jmartisk:
- Closes: https://github.com/quarkiverse/quarkus-langchain4j/issues/1960


jmartisk:
That test doesn't make a lot of sense to me. It connects to the `/mcp/sse` endpoint of the MCP server and expects 200, but that endpoint opens a SSE channel instead of returning a regular response, hence the timeout.

geoand:
> RestAssured doesn't seem to have proper support for SSE and it always waits until the server closes the connection (which never happens), so I think this test should be removed.

I usually resort to using the JAX-RS client when I need to test anything SSE related

PR #1974, opened 2025-11-27
Bump to LangChain4j 1.9.1

mariofusco:


mariofusco:
> Best not to depend snapshots

@geoand It should be ok now.

PR #1982, opened 2025-11-28
Update Azure docs to link to advanced use cases

hubert123490:


sberyozkin:
@hubert123490 Please also have a look at the secure multiple models demo where user login tokens are passed.
How exactly are you refreshing tokens ? Using client _credentials grant ?
If yes, then having ModelAuthProvider that uses OidcClient, exactly as one of the MCP client authentication providers does, may work well, as OidcClient manages token refresh well

hubert123490:
@sberyozkin
Thanks, that's exactly what I was looking for, but for gemini. Would you mind, if I create an issue and propose my solution for this in enhancement request later this week (I would really like to contribute and it should be equivalent to gemini one anyways). I can also test it for my use case. 

Should I close this issue?

sberyozkin:
@hubert123490 Thanks, it would be really great indeed, if you could find time to do the enhancement for Azure OpenAi as well. So yes, please open another issue and create a PR when you get a chance.

>Should I close this issue?

It will be closed once Georgios merges this PR

Thanks

PR #1996, opened 2025-12-04
Add default auth model provider

### LINKED ISSUE #1995: Support default authentication via Managed Identity / Workload Identity for Azure OpenAI
Continuation of issue https://github.com/quarkiverse/quarkus-langchain4j/issues/1981

As a developer, I would like Quarkus LangChain4j to support out-of-the-box authentication for Azure OpenAI using Managed Identity / Workload Identity, as provided by the Azure Identity SDK:
https://learn.microsoft.com/en-us/java/api/overview/azure/identity-readme?view=azure-java-stable

Currently, this is only achievable by supplying a custom model auth provider manually:
https://docs.quarkiverse.io/quarkus-langchain4j/dev/openai-chat-model.html#_using_authprovider

hubert123490:
My solution for default azure auth model provider. Please note that Azure advises to use different authentication method for production https://learn.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential?view=azure-dotnet.

> Simplifies authentication while developing apps that deploy to Azure by combining credentials used in Azure hosting environments with credentials used in local development. In production, it's better to use something else. See [Usage guidance for DefaultAzureCredential](https://aka.ms/azsdk/net/identity/credential-chains#usage-guidance-for-defaultazurecredential).

However it should be sufficient for development purposes and further collaboration. 

- Fixes: https://github.com/quarkiverse/quarkus-langchain4j/issues/1995

### DIFF
--- a/model-providers/openai/azure-openai/deployment/src/main/java/io/quarkiverse/langchain4j/azure/openai/deployment/AzureOpenAiProcessor.java
+++ b/model-providers/openai/azure-openai/deployment/src/main/java/io/quarkiverse/langchain4j/azure/openai/deployment/AzureOpenAiProcessor.java
@@ -9,6 +9,7 @@ import java.util.List;
 
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.inject.Any;
+import jakarta.enterprise.inject.spi.DeploymentException;
 
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.ClassType;
@@ -17,6 +18,7 @@ import org.jboss.jandex.ParameterizedType;
 import org.jboss.jandex.Type;
 
 import io.quarkiverse.langchain4j.ModelName;
+import io.quarkiverse.langchain4j.auth.ModelAuthProvider;
 import io.quarkiverse.langchain4j.azure.openai.AzureOpenAiChatModel;
 import io.quarkiverse.langchain4j.azure.openai.AzureOpenAiEmbeddingModel;
 import io.quarkiverse.langchain4j.azure.openai.AzureOpenAiImageModel;
@@ -59,6 +61,32 @@ public class AzureOpenAiProcessor {
     @BuildStep
     FeatureBuildItem feature() {
         return new FeatureBuildItem(FEATURE);
+    }
+
+    @BuildStep
+    @Record(ExecutionTime.RUNTIME_INIT)
+    public void registerDefaultModelAuthProvider(
+            AzureOpenAiRecorder recorder,
+            BuildProducer<SyntheticBeanBuildItem> producer,
+            LangChain4jAzureOpenAiBuildConfig config) {
+        if (!config.useAzureCredentialModelAuthProvider())
+            return;
+
+        try {
+            Class.forName("com.azure.identity.DefaultAzureCredentialBuilder");
+        } catch (ClassNotFoundException e) {
+            throw new DeploymentException(
+                    "use-azure-credential-model-auth-provider=true requires io.quarkiverse.azureservices:quarkus-azure-identity on the classpath. Add it to your project dependencies.");
+        }
+
+        producer.produce(
+                SyntheticBeanBuildItem
+                        .configure(ModelAuthProvider.class)
+                        .scope(ApplicationScoped.class)
+                        .setRuntimeInit()
+                        .defaultBean()
+                        .createWith(recorder.defaultAzureCredentialModelAuthProvider())
+                        .done());
     }
 
     @BuildStep
--- a/model-providers/openai/azure-openai/deployment/src/main/java/io/quarkiverse/langchain4j/azure/openai/deployment/LangChain4jAzureOpenAiBuildConfig.java
+++ b/model-providers/openai/azure-openai/deployment/src/main/java/io/quarkiverse/langchain4j/azure/openai/deployment/LangChain4jAzureOpenAiBuildConfig.java
@@ -4,6 +4,7 @@ import static io.quarkus.runtime.annotations.ConfigPhase.BUILD_TIME;
 
 import io.quarkus.runtime.annotations.ConfigRoot;
 import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithDefault;
 
 @ConfigRoot(phase = BUILD_TIME)
 @ConfigMapping(prefix = "quarkus.langchain4j.azure-openai")
@@ -28,4 +29,10 @@ public interface LangChain4jAzureOpenAiBuildConfig {
      * Image model related settings
      */
     ImageModelBuildConfig imageModel();
+
+    /**
+     * If true, provides model auth provider, which uses Default Azure Credential for authentication.
+     */
+    @WithDefault("false")
+    boolean useAzureCredentialModelAuthProvider();
 }
--- a/model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorder.java
+++ b/model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorder.java
@@ -19,6 +19,11 @@ import jakarta.enterprise.inject.Any;
 import jakarta.enterprise.inject.Instance;
 import jakarta.enterprise.util.TypeLiteral;
 
+import com.azure.core.credential.AccessToken;
+import com.azure.core.credential.TokenCredential;
+import com.azure.core.credential.TokenRequestContext;
+import com.azure.identity.DefaultAzureCredentialBuilder;
+
 import dev.langchain4j.model.chat.ChatModel;
 import dev.langchain4j.model.chat.DisabledChatModel;
 import dev.langchain4j.model.chat.DisabledStreamingChatModel;
@@ -307,6 +312,16 @@ public class AzureOpenAiRecorder {
         }
     }
 
+    public Function<SyntheticCreationalContext<ModelAuthProvider>, ModelAuthProvider> defaultAzureCredentialModelAuthProvider() {
+        return new Function<>() {
+            @Override
+            public ModelAuthProvider apply(
+                    SyntheticCreationalContext<ModelAuthProvider> modelAuthProviderSyntheticCreationalContext) {
+                return new DefaultAzureCredentialModelAuthProvider();
+            }
+        };
+    }
+
     static String getEndpoint(LangChain4jAzureOpenAiConfig.AzureAiConfig azureAiConfig, String configName, EndpointType type) {
         var endpoint = azureAiConfig.endPointFor(type);
 
@@ -393,4 +408,24 @@ public class AzureOpenAiRecorder {
         });
     }
 
+    public static class DefaultAzureCredentialModelAuthProvider implements ModelAuthProvider {
+        private static final String SCOPE = "https://cognitiveservices.azure.com/.default";
+        private static final TokenCredential credential = new DefaultAzureCredentialBuilder().build();
+        private static final TokenRequestContext context = new TokenRequestContext().addScopes(SCOPE);
+        private static volatile AccessToken currentToken;
+
+        @Override
+        public String getAuthorization(Input input) {
+            AccessToken token = currentToken;
+            if (token == null || token.isExpired()) {
+                synchronized (this) {
+                    token = currentToken;
+                    if (token == null || token.isExpired()) {
+                        currentToken = credential.getTokenSync(context);
+                    }
+                }
+            }
+            return "Bearer " + currentToken.getToken();
+        }
+    }
 }
--- a/docs/modules/ROOT/pages/azure-openai-chat-model.adoc
+++ b/docs/modules/ROOT/pages/azure-openai-chat-model.adoc
@@ -91,6 +91,32 @@ import jakarta.inject.Inject;
 @Inject @ModelName("some-name") ChatModel chatModel;
 ----
 
+== Default Azure Credential
+
+Quarkus LangChain4j supports Azure OpenAI authentication using Azure AD via https://learn.microsoft.com/en-us/java/api/com.azure.identity.defaultazurecredential?view=azure-java-stable[`DefaultAzureCredential`].
+
+To enable it, add the Azure Identity extension:
+
+[source,xml]
+----
+<dependency>
+    <groupId>io.quarkiverse.azureservices</groupId>
+    <artifactId>quarkus-azure-identity</artifactId>
+    <version>1.2.2</version>
+</dependency>
+----
+
+Then enable DefaultAzureCredential authentication in your configuration:
+
+[source,properties]
+----
+quarkus.langchain4j.azure-openai.use-azure-credential-model-auth-provider=true
+----
+
+When enabled, the SDK automatically authenticates using `DefaultAzureCredential`.
+
+It supports Azure Managed Identity and local developer credentials (Azure CLI, IDE login).
+
 == Advanced usage
 
 Please refer to https://docs.quarkiverse.io/quarkus-langchain4j/dev/openai-chat-model.html#_advanced_usage[Open AI].
--- a/model-providers/openai/azure-openai/deployment/pom.xml
+++ b/model-providers/openai/azure-openai/deployment/pom.xml
@@ -20,6 +20,12 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
+            <groupId>io.quarkiverse.azureservices</groupId>
+            <artifactId>quarkus-azure-identity</artifactId>
+            <version>${io.quarkiverse.azureservices-quarkus-azure-identity.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-junit-internal</artifactId>
             <scope>test</scope>
--- a/model-providers/openai/azure-openai/deployment/src/test/java/io/quarkiverse/langchain4j/azure/openai/test/DefaultAzureCredentialModelAuthProviderDisabledTest.java
+++ b/model-providers/openai/azure-openai/deployment/src/test/java/io/quarkiverse/langchain4j/azure/openai/test/DefaultAzureCredentialModelAuthProviderDisabledTest.java
@@ -1,0 +1,33 @@
+package io.quarkiverse.langchain4j.azure.openai.test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkiverse.langchain4j.auth.ModelAuthProvider;
+import io.quarkiverse.langchain4j.azure.openai.runtime.AzureOpenAiRecorder;
+import io.quarkiverse.langchain4j.openai.testing.internal.OpenAiBaseTest;
+import io.quarkiverse.langchain4j.testing.internal.WiremockAware;
+import io.quarkus.arc.Arc;
+import io.quarkus.test.QuarkusUnitTest;
+
+public class DefaultAzureCredentialModelAuthProviderDisabledTest extends OpenAiBaseTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest unitTest = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class))
+            .overrideRuntimeConfigKey(
+                    "quarkus.langchain4j.azure-openai.endpoint", WiremockAware.wiremockUrlForConfig("/v1"));
+
+    @Test
+    void defaultAzureCredentialProviderBeanIsNotRegistered() {
+        var instance = Arc.container().instance(ModelAuthProvider.class);
+        if (instance.isAvailable()) {
+            assertThat(instance.get())
+                    .isNotInstanceOf(AzureOpenAiRecorder.DefaultAzureCredentialModelAuthProvider.class);
+        }
+    }
+}
--- a/model-providers/openai/azure-openai/deployment/src/test/java/io/quarkiverse/langchain4j/azure/openai/test/DefaultAzureCredentialModelAuthProviderTest.java
+++ b/model-providers/openai/azure-openai/deployment/src/test/java/io/quarkiverse/langchain4j/azure/openai/test/DefaultAzureCredentialModelAuthProviderTest.java
@@ -1,0 +1,62 @@
+package io.quarkiverse.langchain4j.azure.openai.test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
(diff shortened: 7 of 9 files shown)


model-providers/openai/openai-common/runtime/pom.xml
```diff
@@ -26,6 +26,11 @@
             <artifactId>quarkus-langchain4j-core</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
```
sberyozkin: Thanks @hubert123490 

I'm not sure this dependency belongs at the OpenAI common level... It should be at the Azure OpenAI level and the filter should be shipped there. Somehow it needs to replace the one checked at the OpenAI common level, may be by extending the QuarkusOpenAIClient, or being an Alternative CDI bean, it is a bit tricky but should be doable, thanks

hubert123490:
@sberyozkin fair point
Im not quite happy abbout configMap, but it seems to work fine overall. There are some issues tho, like quite long token initialization time for the first request. Also it would be great for developer to be able to specify, which credential to use. Currently it gets first available from top to bottom (I think).
image

model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorder.java
```diff
@@ -329,12 +321,6 @@ private LangChain4jAzureOpenAiConfig.AzureAiConfig correspondingAzureOpenAiConfi
         return azureAiConfig;
     }
 
-    private void throwIfApiKeysNotConfigured(String apiKey, String adToken, boolean authProviderAvailable, String configName) {
-        if ((apiKey != null) == (adToken != null) && !authProviderAvailable) {
```
sberyozkin, line 359: Why does this check have to be removed ?

model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/common/DefaultOpenAiRestApiFilterResolver.java
```diff
@@ -0,0 +1,16 @@
+package io.quarkiverse.langchain4j.openai.common;
+
+import java.util.Map;
+import java.util.Optional;
+
+import io.quarkiverse.langchain4j.auth.ModelAuthProvider;
+import io.quarkus.rest.client.reactive.QuarkusRestClientBuilder;
+
+public class DefaultOpenAiRestApiFilterResolver implements RestApiFilterResolver {
```
sberyozkin: I'm just wondering, AFAIK, so far this filter can only be used with Azure OpenAI, I believe for direct OpenAI calls, OpenAI API Key is not passed as a bearer token, @geoand,  is it correct ?

Perhaps we can just move this filter out of OpenAI common ?

geoand: Sorry, I'm not following. With vanilla OpenAI, the authentication is a bearer token

sberyozkin: Sure, so this filter has to stay

model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/AzureRestApiFilterResolver.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.Map;
+import java.util.Optional;
+
+import io.quarkiverse.langchain4j.auth.ModelAuthProvider;
+import io.quarkiverse.langchain4j.openai.common.OpenAiRestApi;
+import io.quarkiverse.langchain4j.openai.common.RestApiFilterResolver;
+import io.quarkus.rest.client.reactive.QuarkusRestClientBuilder;
+
+public class AzureRestApiFilterResolver implements RestApiFilterResolver {
```
sberyozkin: I believe it has to become a synthetic CDI bean - you can see an example of how such beans are created in AzureOpenAI recorder - and you'd have the config properties available to the function that creates it

model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/common/QuarkusOpenAiClient.java
```diff
@@ -131,10 +132,13 @@ public MultivaluedMap<String, String> update(MultivaluedMap<String, String> inco
                         });
                     }
 
-                    ModelAuthProvider
-                            .resolve(builder.configName)
-                            .ifPresent(modelAuthProvider -> restApiBuilder
-                                    .register(new OpenAiRestApi.OpenAIRestAPIFilter(modelAuthProvider)));
+                    Map<String, String> configMap = new HashMap<>();
```
sberyozkin: I think the original code should stay

model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/AzureModelAuthProviderFilter.java
```diff
@@ -0,0 +1,88 @@
...
+import com.azure.core.credential.TokenRequestContext;
+import com.azure.identity.DefaultAzureCredential;
+import com.azure.identity.DefaultAzureCredentialBuilder;
+
+import io.quarkiverse.langchain4j.auth.ModelAuthProvider;
+import io.quarkus.arc.Arc;
+import io.quarkus.arc.InstanceHandle;
+import io.smallrye.mutiny.infrastructure.Infrastructure;
+
+public class AzureModelAuthProviderFilter implements ResteasyReactiveClientRequestFilter {
```
sberyozkin: It should probably extend the Open AI common one

model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/AzureModelAuthProviderFilter.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final Logger log = LoggerFactory.getLogger(AzureModelAuthProviderFilter.class);
+    private final ModelAuthProvider authorizer;
+
+    public AzureModelAuthProviderFilter() {
+        this.authorizer = new ApplicationDefaultAuthProvider();
```
sberyozkin: We should make sure it is loaded only when no other custom model auth providers are available, for example, I linked earlier to the demo where an access token acquired after an Entra ID authorization code flow completion is propagated

model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/AzureModelAuthProviderFilter.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    private record AuthInputImpl(
+            String method,
+            URI uri,
+            MultivaluedMap<String, Object> headers) implements ModelAuthProvider.Input {
+    }
+
+    private static class ApplicationDefaultAuthProvider implements ModelAuthProvider {
```
sberyozkin, line 67: I wonder if we really need the filter resolver at all, and instead, only have this default model auth provider registered as a synthetic CDI bean in the if neither api or ad tokens or custom mode providers are available (this is the check that this PR removes but it should not)

hubert123490:
I've tried to add synthetic bean for model auth provider
```
 @BuildStep
    @Record(ExecutionTime.RUNTIME_INIT)
    public void registerDefaultModelAuthProvider(AzureOpenAiRecorder recorder,
            BuildProducer<SyntheticBeanBuildItem> producer) {
        producer.produce(
                SyntheticBeanBuildItem
                        .configure(ModelAuthProvider.class)
                        .scope(ApplicationScoped.class)
                        .setRuntimeInit()
                        .createWith(recorder.modelAuthProvider())
                        .done());
    }
```
I added in azure recorder 
```
 public Function<SyntheticCreationalContext<ModelAuthProvider>, ModelAuthProvider> modelAuthProvider() {
        if (runtimeConfig.getValue().defaultConfig().apiKey().isPresent()
                || runtimeConfig.getValue().defaultConfig().adToken().isPresent()) {
            return new Function<>() {
                @Override
                public ModelAuthProvider apply(
                        SyntheticCreationalContext<ModelAuthProvider> modelAuthProviderSyntheticCreationalContext) {
                    return null;
                }
            };
        } else {
            return new Function<>() {
                @Override
                public ModelAuthProvider apply(
                        SyntheticCreationalContext<ModelAuthProvider> modelAuthProviderSyntheticCreationalContext) {
                    return new AzureOpenAiRecorder.ApplicationDefaultAuthProvider();
                }
            };
        }
    }
```

but that resulted in 
Error
"Null contextual instance was produced by a normal scoped synthetic bean: SYNTHETIC bean [class=io.quarkiverse.langchain4j.auth.ModelAuthProvider, id=FIym7E17BGsirk4ooWp8Cr-hKM0]"
when api key or ad token is provided
and no matter what I did, I couldn't create this bean conditionally.

However I was successful with BuildConfig approach, such as I added 

> quarkus.langchain4j.azure-openai.azure-default-credentials-enabled=true

```
@ConfigRoot(phase = BUILD_TIME)
@ConfigMapping(prefix = "quarkus.langchain4j.azure-openai")
public interface LangChain4jAzureOpenAiBuildConfig {

    /**
     * Chat model related settings
     */
    ChatModelBuildConfig chatModel();

    /**
     * Embedding model related settings
     */
    EmbeddingModelBuildConfig embeddingModel();

    /**
     * Moderation model related settings
     */
    ModerationModelBuildConfig moderationModel();

    /**
     * Image model related settings
     */
    ImageModelBuildConfig imageModel();

    /**
     * If true, enables the use of Azure Default Credentials for authentication.
     * If not set, the default behavior is disabled.
     */
    @ConfigDocDefault("true")
    Optional<Boolean> azureDefaultCredentialsEnabled();
}
```

and in processor
```
    @BuildStep
    @Record(ExecutionTime.RUNTIME_INIT)
    public void registerDefaultModelAuthProvider(AzureOpenAiRecorder recorder,
            BuildProducer<SyntheticBeanBuildItem> producer,
            LangChain4jAzureOpenAiBuildConfig buildConfig) {

        if (buildConfig.azureDefaultCredentialsEnabled().isPresent() &&
                buildConfig.azureDefaultCredentialsEnabled().get()) {
            producer.produce(
                    SyntheticBeanBuildItem
                            .configure(ModelAuthProvider.class)
                            .scope(ApplicationScoped.class)
                            .setRuntimeInit()
                            .createWith(recorder.modelAuthProvider())
                            .done());
        }
    }
```

And that approach I've already pushed. Since it's default bean it also can be overwritten. Only downside is that even when api-key is provided with quarkus.langchain4j.azure-openai.azure-default-credentials-enabled, the azure model auth provider is used, but maybe it can be handled by warning or exception. Anyways please let me know if it is acceptable one.

model-providers/openai/azure-openai/deployment/src/main/java/io/quarkiverse/langchain4j/azure/openai/deployment/LangChain4jAzureOpenAiBuildConfig.java
```diff
@@ -28,4 +31,11 @@ public interface LangChain4jAzureOpenAiBuildConfig {
      * Image model related settings
      */
     ImageModelBuildConfig imageModel();
+
+    /**
+     * If true, enables the use of Azure Default Credentials for authentication.
+     * If not set, the default behavior is disabled.
+     */
+    @ConfigDocDefault("true")
+    Optional<Boolean> azureDefaultCredentialsEnabled();
```
sberyozkin: Thanks @hubert123490, IMHO this property is not needed, it is activated as a last fallback if neither api key or ad token or custom provider is registered

model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorder.java
```diff
@@ -281,6 +288,16 @@ public ImageModel apply(SyntheticCreationalContext<ImageModel> context) {
         }
     }
 
+    public Function<SyntheticCreationalContext<ModelAuthProvider>, ModelAuthProvider> modelAuthProvider() {
+        return new Function<>() {
+            @Override
+            public ModelAuthProvider apply(
+                    SyntheticCreationalContext<ModelAuthProvider> modelAuthProviderSyntheticCreationalContext) {
+                return new AzureOpenAiRecorder.ApplicationDefaultAuthProvider();
```
sberyozkin: I believe that here you should register it only if no api key or adtoken or custom provider is available...

then indeed, your earlier idea where you removed `throwIfApiKeysNotConfigured` may actually work...

What do you think, @geoand ?

geoand:
It seems like the native failure is related

model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorder.java
sberyozkin: @hubert123490 Please keep in mind that another provider dealing with Azure OpenAI may be registered, as in the demo that I linked to earlier. So this default one should only be registered if no other model provider is available.

IMHO the check done in `throwIfApiKeysNotConfigured` can be done here instead of its current location, the only difference would be, instead of throwing the exception you'd register the default provider

hubert123490: Yes, sorry you are ofcourse right. However I don't think I can register bean in recorder based on condition. I got some errors when I tried that. I might have done something wrong though. Also I don't have access to parameters that the method throwIfApiKeysNotConfigured requires. Anyways please check my latest update, if that is what you had in mind. If not, can you please give me a hint about what you mean about registering provider here?

model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/common/QuarkusOpenAiClient.java
```diff
@@ -131,6 +131,11 @@ public MultivaluedMap<String, String> update(MultivaluedMap<String, String> inco
                         });
                     }
 
+                    if (builder.defaultModelAuthProvider != null && ModelAuthProvider.resolve(builder.configName).isEmpty()) {
```
sberyozkin: I'm sorry @hubert123490, I think it is getting closer, but I do think the default provider should be optionally registered in the Azure recorder - in fact only azure openai deployment code should be changed.

If you'd like I can create something on my branch to suggest how it should look like, it is just that I can't do right now. But to be honest you have already done most of the work here, just don't change anything in the runtime code, only make sure the default provider is optionally registered as a CDI bean.

How about this, update this PR to only return a default provider as a CDI bean,non-optionally, and then we will focus on the next step, how to make sure it is created optionally...

In any case, thanks for your work, enjoy the holidays

model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorder.java
sberyozkin: @hubert123490 Sure, let's start with your PR `only` registering this default provider as a synthetic bean, nothing else. Then we can look in the next commit at moving the existing check out of another recorder function that can throw an exception into your function instead - there will be no need to throw exception but only register a default provider only when we know no other authentication options are set and there we will know that as a last resort, the default provider will always be there...

sberyozkin: It is really all that should be done 1) get the default provider registered as a bean 2) remove the check from the other function and move to your new function, but instead of throwing an exception, register the default one only if no other auth option exists

sberyozkin:
Thanks @hubert123490, let's take it further from this point, I'll try to  experiment as well with your PR, we just need to move that `throw` check that checks the tokens and a custom provider to your new function and register this provider only if if that check returns false

hubert123490:
After several attempts, I believe registering a synthetic bean for DefaultModelAuthProvider correctly is not possible due to the CDI nature of Quarkus — or at least I was not able to find a way to do it. Here are my observations:

1. The synthetic bean must always be registered. This means that when an AD token, API key, or custom ModelAuthProvider is present, we are forced to return a null instance of DefaultModelAuthProvider. The problem is that later in the OpenAI common module, this null instance is picked up and registered as the ModelAuthProvider.

2. ModelAuthProvider has higher priority than API key and AD token in the resolution chain. So when the default bean is registered — even with a null instance — it takes priority and the API key and AD token are never reached.

The only real fixes would be to either pass the ModelAuthProvider directly into the OpenAI common module, or change the REST API filter logic — both of which are probably best avoided.

Given all of this, I suggest we register this bean based on build-time config instead. This way we keep things simple and still give the user a choice to opt into using the default Azure ModelAuthProvider.

This also makes sense from a practical standpoint. A developer running locally would typically want to use DefaultAzureCredentialBuilder, which conveniently picks up credentials from the environment, CLI, or IDE. However, in production they may want to restrict this to a specific credential type — for example, workload identity only — for security or compliance reasons. A build-time flag gives them exactly that control without any extra complexity.

I would really appreciate it if you could take a moment to consider this approach.

model-providers/openai/azure-openai/deployment/src/main/java/io/quarkiverse/langchain4j/azure/openai/deployment/AzureOpenAiProcessor.java
```diff
@@ -8,13 +8,12 @@
 import java.util.List;
 
 import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.spi.DeploymentException;
 
-import org.jboss.jandex.AnnotationInstance;
-import org.jboss.jandex.ClassType;
-import org.jboss.jandex.ParameterizedType;
-import org.jboss.jandex.Type;
+import org.jboss.jandex.*;
```
geoand, line 13: Please avoid star imports

model-providers/openai/azure-openai/deployment/src/main/java/io/quarkiverse/langchain4j/azure/openai/deployment/LangChain4jAzureOpenAiBuildConfig.java
```diff
@@ -28,4 +29,10 @@ public interface LangChain4jAzureOpenAiBuildConfig {
      * Image model related settings
      */
     ImageModelBuildConfig imageModel();
+
+    /**
+     * If true, enables the use of Azure Default Credentials for authentication.
+     */
+    @WithDefault("false")
+    boolean azureDefaultCredentialsEnabled();
```
geoand, line 37: Is this the proper name?

model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorder.java
```diff
@@ -367,4 +382,24 @@ public void run() {
         });
     }
 
+    public static class ApplicationDefaultAuthProvider implements ModelAuthProvider {
+        private static final String SCOPE = "https://cognitiveservices.azure.com/.default";
+        private static final DefaultAzureCredential credential = new DefaultAzureCredentialBuilder().build();
+        private static final TokenRequestContext context = new TokenRequestContext().addScopes(SCOPE);
+        private static volatile AccessToken currentToken;
```
geoand, line 389: This will probably cause problems in native mode.

hubert123490: Using quarkus extension for azure identity should resolve any native build issues. https://github.com/quarkiverse/quarkus-azure-services/tree/main/common/azure-identity

model-providers/openai/azure-openai/deployment/src/main/java/io/quarkiverse/langchain4j/azure/openai/deployment/LangChain4jAzureOpenAiBuildConfig.java
hubert123490: Changed to more appropriate one in next commit (atleast I think so :D)

model-providers/openai/azure-openai/deployment/src/main/java/io/quarkiverse/langchain4j/azure/openai/deployment/AzureOpenAiProcessor.java
hubert123490: ok

model-providers/openai/azure-openai/runtime/src/main/java/io/quarkiverse/langchain4j/azure/openai/runtime/AzureOpenAiRecorder.java
```diff
@@ -393,4 +407,24 @@ public void run() {
         });
     }
 
+    public static class DefaultAzureCredentialModelAuthProvider implements ModelAuthProvider {
+        private static final String SCOPE = "https://cognitiveservices.azure.com/.default";
+        private static final TokenCredential credential = new DefaultAzureCredentialBuilder().build();
+        private static final TokenRequestContext context = new TokenRequestContext().addScopes(SCOPE);
```
geoand, line 414: This is interesting and makes me wonder whether any of this stuff needs to be configurable. If so, it also affects how we support configuration

geoand:
Can you please squash the commits and rebase onto the latest `main`?

model-providers/openai/azure-openai/deployment/src/main/java/io/quarkiverse/langchain4j/azure/openai/deployment/LangChain4jAzureOpenAiBuildConfig.java
```diff
@@ -28,4 +29,10 @@ public interface LangChain4jAzureOpenAiBuildConfig {
      * Image model related settings
      */
     ImageModelBuildConfig imageModel();
+
+    /**
+     * If true, provides model auth provider, which uses Default Azure Credential for authentication.
+     */
+    @WithDefault("false")
+    boolean defaultAzureCredentialModelAuthProvider();
```
geoand:
```suggestion
    boolean useAzureCredentialModelAuthProvider();
```

geoand (commented):
LGTM, thanks!

@sberyozkin WDYT?

Issue #856, opened 2024-09-03
quarkus.langchain4j.pgvector.use-index and quarkus.langchain4j.pgvector.index-list-size?


gsmet:
I'm experimenting with `pgvector` and the two configuration properties `quarkus.langchain4j.pgvector.use-index` and `quarkus.langchain4j.pgvector.index-list-size` are a bit obscure.

I don't know exactly what's their purpose. I set `use-index` to try and then it complained about me needing to set `index-list-size` (or maybe something else?) and had no idea what value to put there.

Now maybe it's because I'm lacking some background about pgvector but in this case a link would be welcome.

I would contribute something but I don't really know what to do here :).

geoand:
cc @jmartisk

gsmet:
Note that I pushed some improvements here that should probably be merged before doing some work on this: https://github.com/quarkiverse/quarkus-langchain4j/pull/857

gsmet:
BTW, I'm asking because our doc is very informative in general and has a lot of details on the concept and most of the time you don't need to dig a lot further to understand everything.

jmartisk:
also cc @sebastienblanc who wrote this code

PR #2006, opened 2025-12-08
Document embedding index usage and recommendations

myfear:
Added section on embedding index and configuration parameters for PGVector.

- Fixed: [#856](https://github.com/quarkiverse/quarkus-langchain4j/issues/856)

### DIFF
--- a/docs/modules/ROOT/pages/rag-pgvector-store.adoc
+++ b/docs/modules/ROOT/pages/rag-pgvector-store.adoc
@@ -64,7 +64,12 @@ If you switch to a different embedding model, ensure the `dimension` value is up
 Use-index controls whether the engine should create and use an index over the embeddings table. 
 To create an index, you need the index-list-size parameter, which is a fine-tuning parameter.
 This basically creates an IVFFlat index with clusters to extend the effectivness of nearest neighbor search.
-If you set the use-index=true you will have to also se the index-list-size=10.
+If you set the `use-index=true` you will have to also se the `index-list-size`.
+
+[NOTE]
+====
+This is something you rarely would need in development. 
+====
 
 [source,properties]
 ----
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/rag-pgvector-store.adoc
```diff
@@ -59,6 +59,30 @@ If the embedding dimension is missing or mismatched, ingestion and retrieval wil
 If you switch to a different embedding model, ensure the `dimension` value is updated accordingly.
 ====
 
+== Embedding Index
+
+Use-index controls whether the engine should create and use an index over the embeddings table. 
+To create an index, you need the index-list-size parameter, which is a fine-tuning parameter.
+This basically creates an IVFFlat index with clusters to extend the effectivness of nearest neighbor search.
+If you set the use-index=true you will have to also se the index-list-size=10.
```
jmartisk:
```suggestion
If you set the `use-index=true`, you will have to also set the `index-list-size`.
```

docs/modules/ROOT/pages/rag-pgvector-store.adoc
```diff
@@ -59,6 +59,30 @@ If the embedding dimension is missing or mismatched, ingestion and retrieval wil
...
+
+[source,properties]
+----
+quarkus.langchain4j.pgvector.use-index=true
+quarkus.langchain4j.pgvector.index-list-size=10
+----
+
+Higher index size values speed up queries by reducing the search space during query time.
```
jmartisk: I might just be confused but https://deepwiki.com/pgvector/pgvector/5.2-creating-and-using-indexes#configuration-parameters says the opposite

> More lists → better recall but slower build and query

AFAIU by this "index size" you mean the number of lists, right?

jmartisk: Yeah actually I think https://neon.com/blog/understanding-vector-search-and-hnsw-index-with-pgvector (where this was apparently taken from) sounds more correct.

Anyway I would rephrase "Higher index size" to "Higher number of lists".

"index-list-size" itself is a bit confusing because it sounds like a higher number means fewer clusters (because they have a bigger size so there's logically fewer of them), but it actually means more clusters :(

PR #2020, opened 2025-12-11
Update GPULlama3 version

orionpapadakis:
This PR updates the version of gpu-llama3 and adds the new TornadoVM installation steps into documentation pages.

PR #2024, opened 2025-12-14
Add optional preview profile for GPU Llama3

geoand:
Fixes release issues reported [here](https://github.com/quarkiverse/quarkus-langchain4j/pull/2020#issuecomment-3646313224).


geoand:
Pushed another update: the parent POM now inherits  from the running JDK so the runtime/deployment modules compile cleanly with preview on both JDK 21 and 24. Hopefully that resolves the compatibility issue 🤞

Issue #2017, opened 2025-12-11
Bug: ChatMemoryStore.updateMessages() is never called when a tool with ReturnBehavior.IMMEDIATE is executed


stanlayze:
**Context**
In our project we use a custom ChatMemoryStore to persist the LLM-Conversation history as documented [here](https://docs.quarkiverse.io/quarkus-langchain4j/dev/messages-and-memory.html#_memory_backends_with_chatmemorystore).

**Problem**
But when a tool with `ReturnBehavior.IMMEDIATE` is executed and the agent-loop of the `AiService` is ended, the `ChatMemoryStore` is empty for the according `memoryId`. While debugging, I realized the `ChatMemoryStore.updateMessages()`method is never called. As a result, there is NO persisted ChatRequests and ChatResponses.

**Observation**
When removing the `ReturnBehavior.IMMEDIATE` from the tool, the `ChatMemoryStore` is invoked as expected.

**Expected Behavior**
`ChatMemoryStore` should be invoked in all cases - with `ReturnBehavior.TO_LLM` and `ReturnBehavior.IMMEDIATE` tools.

Thank you very much for your support! A quick fix is much appreciated!

codesmith25103:
Hey I am new to open source. But interested in this project and want to contribute.
Also is there any slack channel which i can join, so i can interact with other contributers

geoand:
@stanlayze could you check if https://github.com/quarkiverse/quarkus-langchain4j/pull/2025 fixes your issue?

sabob:
@geoand  For auto-commit, will this work? Or you have something else in mind?

```java
public interface ChatMemoryCommitStrategy {
    boolean isAutoCommit();
}
```
Default implementation:

```java
 ChatMemoryCommitStrategy DEFAULT = () -> false;
```

Can be customized with 
```java
@RegisterAiService(
    chatMemoryCommitStrategySupplier = AutoCommitStrategySupplier.class
)

public class AutoCommitStrategySupplier implements Supplier<ChatMemoryCommitStrategy> {

    @Override
    public ChatMemoryCommitStrategy get() {
        return new ChatMemoryCommitStrategy() {

            @Override
            public boolean isAutoCommit() {
                return true;
            }
        };
    }
}
```

dliubarskyi:
> Do we want to preserve the “system message always first + single instance” behavior for Immediate as well, or is it acceptable that this responsibility shifts entirely to the underlying ChatMemory?

[...]

PR #2025, opened 2025-12-14
Fix: Persist ChatMemory when Tool ReturnBehavior is IMMEDIATE (#2017)

codesmith25103:
- Fixes: #2017


codesmith25103:
However, when I ran the test case after building the project, it still failed.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -468,6 +468,7 @@ private static Object doImplement0(AiServiceMethodCreateInfo methodCreateInfo, I
                 committableChatMemory.add(toolResult);
             }
             if (immediateToolReturn) {
+                committableChatMemory.commit();
```
stanlayze, line 471: Should we commit the messages after the type-check in the next 3 lines?

geoand: Yeah, that probably makes sense

PR #2034, opened 2025-12-18
Fix config mapping registration in gpu-llama3 runtime module

orionpapadakis:
## Fix config mapping registration in gpu-llama3 runtime module

This PR adds `quarkus-extension-processor` to `annotationProcessorPaths` in the gpu-llama3 runtime module's compiler configuration.

Without this processor, `@ConfigMapping` classes were compiled but never registered, causing `SRCFG00027: Could not find a mapping` errors during integration test builds.

The processor was present in the deployment module but missing from runtime, preventing proper config discovery for `BUILD_AND_RUN_TIME_FIXED` phase configs.


orionpapadakis:
Maybe this is also related to the reported CI errors in #2020

Issue #1964, opened 2025-11-25
On some envrionments MCP Server passes wrong values to tools


fedinskiy:
Occasionally, I got the following error when running samples/mcp tools tests:
```
1495-a55f-42bb-86cc-560c6617de2f]: dev.langchain4j.exception.ToolExecutionException: Error: Invalid arguments for read_file: [
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "path"
    ],
    "message": "Invalid input: expected string, received undefined"
  }
]
```
It looks like this error always happens on OSes, which run on VMs. I was able to reproduce it on RHEL on Openstack VM as well as Fedora and Ubuntu on podman and was not able to reproduce it on a bare-metal Fedora.
According to the logs, it is caused on Quarkus/Langchain4j side, which sends an empty request to the tool: `Attempting to execute tool ToolExecutionRequest { id = "call_<edited>", name = "read_file", arguments = "{}" }`. In usual conditions, the command looks differently: ` Attempting to execute tool ToolExecutionRequest { id = "call_<edited>", name = "read_file", arguments = "{"path":"playground/hello.txt"}" }`
How to reproduce:
```
podman run --name fedora --rm -it fedora:latest /bin/bash
# the following is run inside the VM. also works for ubuntu:latest, but with apt and xz-utils
dnf update
dnf install git wget xz curl zip unzip

curl -s "https://get.sdkman.io" | bash
source "/root/.sdkman/bin/sdkman-init.sh"

sdk install java 21.0.8-tem
sdk install maven 3.9.9

wget -q -O node.tar.xz https://nodejs.org/dist/v22.20.0/node-v22.20.0-linux-x64.tar.xz
mkdir node
tar -xf node.tar.xz --directory=node --strip-components=1
PATH=$PATH:$(readlink -f node)/bin
export PATH

git clone  https://github.com/quarkiverse/quarkus-langchain4j
cd quarkus-langchain4j/samples/
API_KEY=__YOUR_API_KEY
mvn clean verify -V -B --no-transfer-progress -fae -Dquarkus.langchain4j.openai.api-key=$API_KEY -Dquarkus.platform.version=3.27.1 -Dquarkus.platform.group-id=io.quarkus.platform -Dplatform-deps -DskipTests=false -pl mcp-tools
```
Version 3.27.0 is affected as well, version 3.26.3 is not. Can be reproduced on both `main` and `1.2` branch

[...]

PR #2035, opened 2025-12-19
Add a minimal implementation of MCP server

fedinskiy:
Add a very basic implementation of MCP server to examples. The server provides a tool to read files.
Why is it needed:
1. This is an implementation, which doesn't depend on external services
2. It can be used as a replacement in mcp-tools example. This frees user from installing NPM and protects us from changes in the related node package (eg https://github.com/quarkiverse/quarkus-langchain4j/issues/1964)

UPD: example for using with mcp-tools:
```
 which npm
/usr/bin/which: no npm in
mvn clean verify --no-transfer-progress -Dquarkus.langchain4j.openai.api-key=$API_KEY -Dquarkus.platform.version=3.27.1 -Dquarkus.platform.group-id=io.quarkus.platform -Dplatform-deps -DskipTests=false -DskipITs=false -Dquarkus.langchain4j.mcp.filesystem.command="java,-Dworking.folder=playground,-jar,$PWD/mcp-fileserver/target/quarkus-langchain4j-sample-mcp-fileserver-1.0-SNAPSHOT-runner.jar" -pl mcp-fileserver,mcp-tools
```


geoand:
Does this really belong here? 

cc @jmartisk @mkouba

jmartisk:
So this just shows how to build a server with the Quarkus MCP server extension? We do have some other samples that show how to build servers like that (although this repo is for clients, not servers...),

In the tools sample, I would prefer to keep using the server from NPM instead of re-inventing some parts of it (I know we had some issues with it, but they are solved now). On top of that, we have a rather full-blown `filesystem` MCP server sample in https://github.com/quarkiverse/quarkus-mcp-servers/tree/main/filesystem

 Showing how to easily plug an MCP server from an external repository is definitely useful, I wouldn't get rid of that.

PR #2049, opened 2026-01-07
Allow Gemini Schema detection to be null

SimonSuewerUHH:
Good afternoon everyone,

Here is a suggestion that allows the schema to be null in Gemini. This makes ResponseFormatType.Text possible.
Currently, a workaround is possible in which you set the format to null.
Like:
`AiGeminiChatLanguageModel.builder()
                .key(...)
                .modelId(...)
                .responseFormat(null)
`

However, to ensure that interfaces such as the following will work again in the future, I propose the change.
`
@SystemMessage(""".....
            """)
    @UserMessage("""
            QUESTION:
            {question}
            """)
    String answer(String question);
`
I have also moved two methods to the base class. In general, the existing base class could be used even more to avoid duplication.


geoand:
Thanks!

Can you please squash the commits?

SimonSuewerUHH:
> Thanks!
> 
> Can you please squash the commits?

Hi, thanks, squashed my commits into onne.
Cheers

geoand (approved):

Issue #2043, opened 2025-12-30
Upgrade langchain4j to 1.10.0


sabob:
Hi all,

In order to implement #1987 we need to upgrade to langchain4j 1.10.0. 
1.10.0 has breaking changes though eg. the annotation `dev.langchain4j.agentic.declarative.BeforeAgentInvocation`  was removed.

Also, the upstream Langchain4j fix [4215](https://github.com/langchain4j/langchain4j/pull/4215), means we need to add the chatRequest when firing `AiServiceResponseReceivedEvent`:

``` java
public class AiServiceMethodImplementationSupport {
    public Object implement(Input input) {
....
    context.eventListenerRegistrar.fireEvent(
                AiServiceResponseReceivedEvent.builder()
                        .invocationContext(invocationContext)
                        .response(response)
                        .request(chatRequest) // <!-- we need to add this
                        .build());
```
I'm not sure how to handle the removal of those annotations though.

geoand:
cc @mariofusco

PR #2057, opened 2026-01-09
Bump to LangChain4j 1.10

mariofusco:


agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/validation/NonStaticReturnBeforeAgentInvocationTest.java
geoand, line 1: Why was this and some other tests deleted?

mariofusco: Because that annotation doesn't exist anymore. It has been replaced by the new `@AgentListenerSupplier` which I'm testing now. This is the effect of the rework I did on agents observability described that now works as described here https://docs.langchain4j.dev/tutorials/agents#observability

mariofusco: Also see https://github.com/quarkiverse/quarkus-langchain4j/issues/2043#issuecomment-3702007512

geoand: Got it, thanks

PR #2067, opened 2026-01-15
Fix a typo in the ToolBox class javadoc

PierreBtz:


geoand (approved):
Thanks!

jmartisk (approved):

cescoffier (approved):

core/runtime/src/main/java/io/quarkiverse/langchain4j/ToolBox.java
```diff
@@ -7,7 +7,7 @@
 import java.lang.annotation.Target;
 
 /**
- * When used on a method of an AiService annotated with {@link RegisterAiService}, the method will the tool classes provided by
+ * When used on a method of an AiService annotated with {@link RegisterAiService}, the method will use the tool classes provided by
  * {@code value}
  * instead of the ones configured for the entire AiService (via {@link RegisterAiService#tools()})
```
omatheusmesmo, line 12:
```suggestion
 * When used on a method of an AiService annotated with {@link RegisterAiService}, the method will use the tool classes
* provided by {@code value}
* instead of the ones configured for the entire AiService (via {@link RegisterAiService#tools()})
```

Hello @PierreBtz , you can just commit this on the web and will work, but you will need to squash the commits after

PR #2070, opened 2026-01-16
Improve validation of ToolInputGuardrail/ToolOutputGuardrail CDI beans

### LINKED ISSUE #2052: Improve javadoc of ToolInputGuardrail/ToolOutputGuardrail for more complex class hierarchies
In the javadoc I can read _"Implementations must be CDI beans to enable dependency injection."_ but they are applied to a tool method using the `@ToolInputGuardrails`/`@ToolOutputGuardrails` which means that we try to lookup the beans using the class used in the annotation. 

This can cause trouble for more complex class hierarchies. For example, if there is `class FooFilter implements ToolInputGuardrail` and `class BarFilter extends FooFilter` then for `@ToolInputGuardrails({FooFilter.clss, BarFilter.class})` you'll get an ambiguous dependency error.

A simple fix is to use the `@jakarta.enterprise.inject.Typed` , e.g. `class @Typed(BarFilter.class) BarFilter extends FooFilter`. It might be worth documenting this use case.

mkouba:
- handle ambiguous beans at build time
- resolves #2052


geoand:
Thanks! 

Do we also have to update the documentation?

PR #2072, opened 2026-01-20
Prevent stream from hanging when exception is thrown from tool calls

Wal8800:
After the changes in the tool executor, we no longer convert the exception into tool result at executor and throw the exception instead https://github.com/quarkiverse/quarkus-langchain4j/pull/2014.

If the response callback was occurring on the event loop, we would switch to executing a worker thread. When an exception occurs, we don't handle the returned `Future`,  as a result we **swallow the exception.**

The downstream impact is that the stream hangs because we haven't trigger the `onError` handle to terminate the stream.

Key changes:

- Wrap the runnable with try-catch and call `onError`
   - Not sure if this is the best approach, I could see handling the returned `Future` object in `executeOnWorkerThread` as another feasible path.
 - Ported over `executeTool` from the non streaming flow to support Tool execution error handler
 - Add unit tests to replicate both error scenario, the new test cases fail (stream hangs and timeout) without the fix

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceStreamingResponseHandler.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceStreamingResponseHandler.java
@@ -19,6 +19,7 @@ import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.ChatMessage;
 import dev.langchain4j.data.message.ToolExecutionResultMessage;
+import dev.langchain4j.exception.ToolArgumentsException;
 import dev.langchain4j.invocation.InvocationContext;
 import dev.langchain4j.model.StreamingResponseHandler;
 import dev.langchain4j.model.chat.StreamingChatModel;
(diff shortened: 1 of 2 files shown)


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceStreamingResponseHandler.java
```diff
@@ -294,6 +308,55 @@ public Object call() throws Exception {
         }
     }
 
+    private ToolExecutionResult executeTool(ToolExecutionRequest toolExecutionRequest, ToolExecutor toolExecutor,
```
geoand, line 332: Can we add some comments to this explaining what it does?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceStreamingResponseHandler.java
```diff
@@ -265,10 +271,18 @@ public StreamingHandle getStreamingHandle() {
     }
 
     private void executeTools(Runnable runnable) {
+        Runnable safeRunnable = () -> {
+            try {
+                runnable.run();
+            } catch (Exception e) {
+                onError(e);
+            }
+        };
```
geoand: Please replace the lambda with an anonymous class

geoand (commented):
Thanks a lot for this!

I only added some small comments

geoand:
Thanks a lot! Mind squashing the commits?

PR #2075, opened 2026-01-21
Add promptCaching / enableReasoning config to Bedrock model

Dohbedoh:
See https://github.com/quarkiverse/quarkus-langchain4j/discussions/2073 and https://github.com/langchain4j/langchain4j/pull/3664.

Adding prompt caching configurability. Configuration would look like this:

```yaml
  quarkus:
    langchain4j:
      bedrock:
        chat-model:
          model-id: "us.anthropic.claude-sonnet-4-5-20250929-v1:0"
          prompt-caching: AFTER_SYSTEM
          enable-reasoning: 5000
```

- Closes: #2073

### DIFF
--- a/model-providers/bedrock/runtime/src/main/java/io/quarkiverse/langchain4j/bedrock/runtime/config/ChatModelConfig.java
+++ b/model-providers/bedrock/runtime/src/main/java/io/quarkiverse/langchain4j/bedrock/runtime/config/ChatModelConfig.java
@@ -5,6 +5,7 @@ import java.util.Optional;
 import java.util.OptionalDouble;
 import java.util.OptionalInt;
 
+import dev.langchain4j.model.bedrock.BedrockCachePointPlacement;
 import io.quarkus.runtime.annotations.ConfigDocDefault;
 import io.quarkus.runtime.annotations.ConfigGroup;
 import io.smallrye.config.WithDefault;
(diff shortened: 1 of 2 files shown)


model-providers/bedrock/runtime/src/main/java/io/quarkiverse/langchain4j/bedrock/runtime/config/ChatModelConfig.java
```diff
@@ -85,4 +86,21 @@ public interface ChatModelConfig extends AwsClientConfig {
...
+     */
+    Optional<BedrockCachePointPlacement> promptCaching();
+
+    /**
+     * Enables reasoning capabilities of the model. It requires to Set the maximum number of tokens to allocate for reasoning.
+     *
+     * @see <a href="https://docs.aws.amazon.com/bedrock/latest/userguide/inference-reasoning.html">AWS Bedrock inference
+     *      reasoning</a>.
+     */
+    OptionalInt enableReasoning();
```
geoand: Maybe we should rename this because normally, `enableReasoning` would be a boolean, not an int.

Dohbedoh: I preserved the name used in langchain4j. We can rename it but we need to make the reasoning budget configurable with an int.

We could do:

```
    OptionalInt reasoning();
```

Or create a `Reasoning` config:

```
    ReasoningConfig reasoning();

    interface ReasoningConfig {

        /** Whether reasoning is enabled. */
        @WithDefault("false")
        Boolean enabled();

        /** The token budget to allocate for reasoning. */
        Optional<Integer> tokenBudget();

    }
```

Though this might be too much ?

geoand: Let's go with `reasoning` in that case

geoand: Can you please squash the commits?

Dohbedoh: sure. Done!

PR #2077, opened 2026-01-21
Add instructions to websocket-based samples about how to connect via wscat

jmartisk:
I use this quite often myself during development, I thought some people may find it useful


geoand (approved):
Thanks!

PR #2083, opened 2026-01-22
Several test-related fixes

fedinskiy:
This PR contains the following changes:
1. Adds jbang to nightly and ecosystem builds, so MCP tests can be run regularly
2. Tune prompts for ai assistance tests, so these tests can be run with real LLMs and receive predictable answers from them
3. Drop cloud-llm profile from openai module. Currently this profile needs to be overloaded thrice (`quarkus.profile`, `test-profile` and `integration-tests.profile`) to work and there are still occasional issues, when the correct property is not picked up; it is easier to just overload `quarkus.langchain4j.openai.api-key` and `quarkus.langchain4j.openai.base-url` manually. I took the liberty to switch from `gpt-4o` to `gpt-4o-mini` in this commit, but I can move the change to Quarkus QE jobs if this is not acceptable. I am planning to backport this patch to 1.2 branch as well.

### DIFF
--- a/integration-tests/openai/src/test/java/org/acme/example/openai/chat/ChatLanguageModelResourceTest.java
+++ b/integration-tests/openai/src/test/java/org/acme/example/openai/chat/ChatLanguageModelResourceTest.java
@@ -18,7 +18,7 @@ import jakarta.ws.rs.sse.SseEventSource;
 import org.acme.example.openai.TestUtils;
 import org.hamcrest.MatcherAssert;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.EnabledIfSystemProperty;
+import org.junit.jupiter.api.condition.EnabledIf;
 
 import io.quarkus.test.common.http.TestHTTPEndpoint;
 import io.quarkus.test.common.http.TestHTTPResource;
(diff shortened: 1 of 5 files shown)


integration-tests/openai/src/test/java/org/acme/example/openai/chat/ChatLanguageModelResourceTest.java
```diff
@@ -42,7 +42,7 @@ public void blocking() {
     }
 
     @Test
-    @EnabledIfSystemProperty(named = "quarkus.test.profile", matches = "cloud-llm", disabledReason = "The Mock API does not handle streaming properly")
+    @EnabledIf(value = "org.acme.example.openai.TestUtils#usesLLM", disabledReason = "The Mock API does not handle streaming properly")
```
geoand, line 45: This value is weird.

integration-tests/openai/src/test/java/org/acme/example/openai/chat/ChatLanguageModelResourceTest.java
```diff
@@ -84,7 +84,7 @@ public void structuredPrompt() {
...
```
geoand, line 87: Same here

integration-tests/openai/src/test/java/org/acme/example/openai/chat/ChatLanguageModelResourceTest.java
fedinskiy: Unfortunately, this is how `@EnabledIf` annotation works (https://docs.junit.org/5.7.2/api/org.junit.jupiter.api/org/junit/jupiter/api/condition/EnabledIf.html#value()). And, however weird, this works as expected.

If you do not like it for aesthetic reasons, I could create a custom annotation for this module.

geoand: No no, I'm fine with it. What is weird is the use of `#usesLLM`, but I guess we can live with it

PR #2099, opened 2026-02-02
LangChain4j 1.11 + MCP client Micrometer metrics

jmartisk:


mcp/runtime/pom.xml
```diff
@@ -41,6 +41,11 @@
             <artifactId>quarkus-smallrye-health</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-core</artifactId>
+            <optional>true</optional>
```
mariofusco, line 47: I believe that declaring this `optional` is breaking the native build? Or maybe it would be enough to add this dependency (with scope test?) also in the `integraton-test/mcp` module.

jmartisk: Well I don't want to have this as a required dependency. I need to figure out how to get the native build working when it's not present...

Issue #2091, opened 2026-01-27
Provide a demo that shows how MCP Client can be bound dynamically as a tool to AI service


sberyozkin:
It would be great to have a very basic AI assistant demo that can offer a user an option to load an MCP server definition from a disk or simply add an MCP server URL, without having to pre-configure it on the startup, and have this MCP server bound as the AI service tool that the AI service can use to satisfy some LLM function call requests. For ex, a weather assistant that allows user to add a Weather Service MCP server.

Once this is done, I will work on a demo that shows how it can work when the imported MCP server requires a dedicated user login, for us to figure out what can be synthesized out of it into a dedicated utility or extension support or some doc instructions only.

@jmartisk Would you like to have a look at the basic demo ? 

CC @cescoffier @maxandersen

PR #2112, opened 2026-02-06
Assistant with dynamically created MCP clients

jmartisk:
(looking for perhaps a different name for this sample...)

- Closes: https://github.com/quarkiverse/quarkus-langchain4j/issues/2091


geoand:
I think that's much better

sberyozkin:
Awesome, this is a start of something very cool in Quarkus LangChain4j, @jmartisk

jmartisk:
@sberyozkin For STDIO transport, it suggests the `uvx mcp-server-time`, I'm not sure what to choose for remote servers though. I don't want to add a dependency on an external thing that we don't control (that always complicates things a lot). And if we also avoid things that require setting up an API key, it doesn't leave a lot of options I guess. I'll try to look at what options we may have in servers that can be spawned easily and support HTTP and don't require API keys.

I'm not going to merge this into quarkus-langchain4j 1.7.0 (which will be in Quarkus 3.33 LTS), let's polish it up a bit more first (and decide the extent of security support)

Issue #2062, opened 2026-01-13
`AiServiceMethodImplementationSupport.doImplement0` redundantly fetches the messages from the ChatMemoryStore


chirino:
It happens when it 
* calls `prepareSystemMessage
* calls context.chatMemoryService.getChatMemory(memoryId).messages()
* calls new DefaultCommittableChatMemory(chatMemory)

Perhaps we should create the DefaultCommittableChatMemory before the call to prepareSystemMessage so that from that point on we are working with a cached copy of the memory.

geoand:
https://github.com/quarkiverse/quarkus-langchain4j/blob/main/model-providers/openai/openai-vanilla/deployment/src/test/java/org/acme/examples/aiservices/CustomChatMemoryStoreTest.java#L53 may be what you are looking for

PR #2117, opened 2026-02-09
fix: AiServiceMethodImplementationSupport.doImplement0 redundantly fetches the messages from the ChatMemoryStore

chirino:
This change reduces the number of times `ChatMemoryStore.getMessages` gets called to just 1 time. This change depends the new ChatMemory.set method introduced in langchain4j 1.11.0

fixes: #2062


geoand (approved):
Thanks!

PR #2130, opened 2026-02-16
Use upstream watsonx.ai integration from LangChain4j

andreadimaio:
Replaced the custom `watsonx.ai` integration previously implemented in Quarkus with the upstream implementation now available in LangChain4j.

The local implementation has been removed and all references have been updated to rely on the maintained upstream module.

Apologies for the size of this PR 😅


andreadimaio:
I've rebased the branch.

The conflict was caused by the `logCurl` feature that was added. Since `logCurl` is a quarkus specific feature, I couldn't pass it through the builder chain in the usual way.

I opted for a `ThreadLocal` in a `QuarkusRestClientConfig` class.
The `WatsonxRecorder` sets the value before calling `.build()` and clears it in a finally block, while each `Quarkus*RestClient` reads it from the `ThreadLocal` in its constructor. 

This should work safely.

geoand:
> I opted for a ThreadLocal in a QuarkusRestClientConfig class.

Yeah, unfortunately there is no better way to handle this

Issue #2098, opened 2026-02-01
Exceptions during AiServiceEvents are eaten


edeandrea:
This is fixed upstream (see Exceptions during AiServiceEvents are eaten)

Once that is released it'll need to be integrated here with an easy way (through config maybe?) to expose it. Maybe in the `AiServiceSelector`? I'm not sure, but I'll figure something out.

PR #2161, opened 2026-02-24
Exceptions during AiServiceEvents are eaten

edeandrea:
- Fixes: #2098

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
@@ -177,6 +177,21 @@ public @interface RegisterAiService {
     Class<? extends SystemMessageProvider> systemMessageProviderSupplier() default NoSystemMessageProviderSupplier.class;
 
     /**
+     * Indicates whether exceptions thrown during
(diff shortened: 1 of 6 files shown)


core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
```diff
@@ -176,6 +176,13 @@
      */
     Class<? extends SystemMessageProvider> systemMessageProviderSupplier() default NoSystemMessageProviderSupplier.class;
 
+    /**
+     * Indicates whether to throw exceptions on event errors.
+     * If {@code true}, exceptions will be thrown; otherwise, errors will be handled silently.
+     * Default is {@code false} to preserve backwards compatibility.
```
geoand: Reading this, I have no idea what events this is referring to

edeandrea: I struggled with that, because on one hand I could be very specific, but on the other hand, I wanted to leave it open-ended for other things that fire events could tap into.

I'll make it super specific for now and we can change it later if things change.

edeandrea: How's that?

Issue #2174, opened 2026-02-26
Streaming cancellation does not stop the agent tool loop


Wal8800:
## Problem

When a subscriber of `Multi<ChatEvent>` cancels or errors mid-stream, the current HTTP SSE connection is closed via StreamingHandle.cancel(), but the agent tool execution loop inside QuarkusAiServiceStreamingResponseHandler.onCompleteResponse() continues running unchecked.

After cancellation, the handler keeps:
  - Executing tools
  - Saving messages to chat memory
  - Making new LLM streaming requests

All of this runs silently because the dead UnicastProcessor drops events. The loop only stops when the LLM decides to stop calling tools.

Additionally, the existing `stream.getStreamingHandle().cancel()` in TokenStreamMulti only references the first round's StreamingHandle. New handlers created for subsequent tool rounds are local variables in onCompleteResponse() and are unreachable from the cancellation callback.

## Proposed Solution

Introduce a shared AtomicBoolean cancelled flag:
  - Created in TokenStreamMulti.subscribe() and set to true in the onCancellation callback

```java
  Multi<ChatEvent> cancellableMulti = processor.onCancellation().invoke(() -> {
      cancelled.set(true);
  });
```

  - Passed through QuarkusAiServiceTokenStream to QuarkusAiServiceStreamingResponseHandler
  - The handler checks the flag before executing tools and before making the next chat() call
  - The handler also checks in onPartialResponse() and self-cancels its own StreamingHandle, working uniformly across all rounds
  - The same AtomicBoolean reference is passed to new handlers created for subsequent rounds

Happy to work on the solution for this issue.

geoand:
The proposed solution sounds good!

Having a test as well would be great

PR #2177, opened 2026-02-27
Add shared cancel flag for agent tool loop

Wal8800:
Issue: https://github.com/quarkiverse/quarkus-langchain4j/issues/2174

**Key changes**
- Introduced a shared cancel flag to stop the agent tool loop when it's true.
- Return HTTP exception with response body for Anthropic non 200 response.
    - This allows the downstream to figure out what was the error.
- Add a unit to assert for the behaviour

**Manual test**

_built the library and uses it my local API service_

- When cancelling the stream, the agent tool loop stopped
- I can make another request with `AiSerivice.chat` without errors.

### DIFF
--- a/model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/AnthropicRestApi.java
+++ b/model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/AnthropicRestApi.java
@@ -19,6 +19,7 @@ import jakarta.ws.rs.WebApplicationException;
 import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.MultivaluedMap;
+import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.ext.MessageBodyWriter;
 import jakarta.ws.rs.ext.WriterInterceptor;
 import jakarta.ws.rs.ext.WriterInterceptorContext;
@@ -30,10 +31,12 @@ import org.jboss.resteasy.reactive.common.providers.serialisers.AbstractJsonMess
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 
+import dev.langchain4j.exception.HttpException;
 import dev.langchain4j.model.anthropic.internal.api.AnthropicCreateMessageRequest;
 import dev.langchain4j.model.anthropic.internal.api.AnthropicCreateMessageResponse;
 import dev.langchain4j.model.anthropic.internal.api.AnthropicStreamingData;
 import io.quarkiverse.langchain4j.QuarkusJsonCodecFactory;
+import io.quarkus.rest.client.reactive.ClientExceptionMapper;
 import io.smallrye.mutiny.Multi;
 
 @Path("")
@@ -53,6 +56,14 @@ public interface AnthropicRestApi {
     @POST
     @RestStreamElementType(MediaType.APPLICATION_JSON)
     Multi<AnthropicStreamingData> streamMessage(AnthropicCreateMessageRequest request, @BeanParam ApiMetadata apiMetadata);
+
+    @ClientExceptionMapper
+    static RuntimeException toException(Response response) {
+        if (response.getStatusInfo().getFamily() != Response.Status.Family.SUCCESSFUL) {
+            return new HttpException(response.getStatus(), response.readEntity(String.class));
+        }
+        return null;
+    }
 
     class ApiMetadata {
         @HeaderParam(API_KEY_HEADER)


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceStreamingResponseHandler.java
```diff
@@ -403,8 +421,20 @@ public void onCompleteResponse(ChatResponse completeResponse) {
             executeTools(new Runnable() {
                 @Override
                 public void run() {
+                    if (isCancelled()) {
+                        shutdown();
+                        return;
+                    }
                     addToMemory(aiMessage);
                     for (ToolExecutionRequest toolExecutionRequest : aiMessage.toolExecutionRequests()) {
+                        if (isCancelled()) {
```
Wal8800, line 430: If we cancel during the tool execution loop, I opted to provide an error ToolExecutionResultMessage for the rest of the tools.

This is because the memory could go into an inconsistent state. X number of tool uses but fewer than X tool results in memory.

For Anthropic, it errors out. For example:

```json
dev.langchain4j.exception.HttpException: {"type":"error","error":{"type":"invalid_request_error","message":"messages.4: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_012VHAB3QGrjTDAXCRn8oCjV, toolu_016geEpauyGwYuQDoNCtXsGa, toolu_019VNWsAtthETFjbfNq8y2iX, toolu_01Abg9TWUZeRNmij4FNJ4WDY, toolu_01Ezg89FQmj25CHjwoV5vKqg, toolu_01G8igKKv3mBgApJYnSnc61m, toolu_01HSxKqpxqW9gvcWSwKAWNc8, toolu_01M437y8xVbz8CxSuJUKLfmZ. Each `tool_use` block must have a corresponding `tool_result` block in the next message."},"request_id":"req_011CYXTjqDZE3okd1zCDie3X"}
```

I assumed a similar case could happen with other providers, so I added a matching tool result message to avoid the possible error. 

Overall, I aim to cancel the loop in a state where the user can continue without needing to clean up the memory.

model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/AnthropicRestApi.java
```diff
@@ -54,6 +57,14 @@ public interface AnthropicRestApi {
     @RestStreamElementType(MediaType.APPLICATION_JSON)
     Multi<AnthropicStreamingData> streamMessage(AnthropicCreateMessageRequest request, @BeanParam ApiMetadata apiMetadata);
 
+    @ClientExceptionMapper
+    static RuntimeException toException(Response response) {
```
Wal8800, line 61: Sneaking this change in.

This was helpful to see the 400 response error payload from Anthropic.

We have it for the OpenAI client already

https://github.com/quarkiverse/quarkus-langchain4j/blob/b8e82425e7d8d71e009a47145f92d4c497679798/model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/common/OpenAiRestApi.java#L173-L179

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceStreamingResponseHandler.java
```diff
@@ -533,6 +569,10 @@ private void shutdown() {
         }
     }
 
+    private boolean isCancelled() {
+        return cancelled != null && cancelled.get();
```
geoand: When can `cancelled` be `null`?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceTokenStream.java
```diff
@@ -208,16 +212,12 @@ public void start() {
...
-     * @apiNote This uses langchain4j's experimental StreamingHandle API (since 1.8.0)
-     */
-    public StreamingHandle getStreamingHandle() {
-        return handler != null ? handler.getStreamingHandle() : null;
+    StreamingHandle getStreamingHandle() {
+        QuarkusAiServiceStreamingResponseHandler currentHandler = handler;
+        if (currentHandler == null) {
+            return NoopStreamingHandle.INSTANCE;
+        }
+        return currentHandler.getStreamingHandle();
```
geoand: Why this change?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceStreamingResponseHandler.java
Wal8800: Good callout, reviewing the code, there isn't a scenario where we never pass down a `null` or initialise to be `null`.

I will remove it the defensive check.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceTokenStream.java
Wal8800: Sorry, this was me experimenting with the changes.

Initially thought we don't need this method but opted to keep it so we can still cancel the http request directly for the first turn.

Claude code didn't revert the changes as it is for me 😓.

I will revert.

Issue #2227, opened 2026-03-11
OAuth Authentication Succeeds but MCP Reconnection Fails - Requires Qaurkus app Restart


mescaja:
Created an AI Service 

```
@ApplicationScoped
@RegisterAiService
public interface TestMcpAiService {
   @McpToolBox("test-toolbox")
   String callTestTool(@UserMessage String userMessage);
}
```

the AI Service is setup to connect to an MCP server endpoint https://test-backend.poc/mcp

```
quarkus.langchain4j.mcp.wanaku-toolbox.transport-type=streamable-http
quarkus.langchain4j.mcp.wanaku-toolbox.url=https://test-backend.poc/mcp
quarkus.langchain4j.chat-model.provider=mistralai
```

Created a Rest resource /test and protected it with OIDC scheme using Quarkus oidc extension. I reproduce the issue like this

go to http:localhost:8080/test
Authorization code flow is triggered
enter user credentials
Rest resource acquire the access token and run the AI service
AI service runs successfully and return the expected result
soon after quarkus app keeps throwing this exception around every min

```
 WARN  [dev.langchain4j.mcp.client.DefaultMcpClient] (mcp-server-health-checker) mcp server reconnection failed: java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.RuntimeException: Unexpected status code: 401
	at dev.langchain4j.mcp.client.DefaultMcpClient.initialize(DefaultMcpClient.java:169)
	at dev.langchain4j.mcp.client.DefaultMcpClient.triggerReconnection(DefaultMcpClient.java:537)
	at dev.langchain4j.mcp.client.DefaultMcpClient.lambda$startAutoHealthCheck$0(DefaultMcpClient.java:524)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
	at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:358)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
[...]

PR #2240, opened 2026-03-17
Add properties to configure automatic MCP health checks

jmartisk:
See https://github.com/quarkiverse/quarkus-langchain4j/issues/2227

### DIFF
--- a/mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/config/McpClientRuntimeConfig.java
+++ b/mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/config/McpClientRuntimeConfig.java
@@ -121,4 +121,19 @@ public interface McpClientRuntimeConfig {
      */
     Optional<Boolean> cacheToolList();
 
+    /**
+     * Whether the MCP client should perform health checks regularly and try to reconnect immediately if a connection issue is
+     * detected.
+     * The default is true.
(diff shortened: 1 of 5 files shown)


mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/config/McpClientRuntimeConfig.java
```diff
@@ -121,4 +121,19 @@ public interface McpClientRuntimeConfig {
...
+     */
+    @WithDefault("true")
+    Boolean autoHealthCheck();
+
+    /**
+     * The interval for automatic health checks (if enabled using the "autoHealthCheck" property).
+     * The default is 60 seconds.
+     */
+    @WithDefault("60s")
+    Duration autoHealthCheckInterval();
```
geoand, line 137: Would it make sense to group these into a new config group?

jmartisk: Frankly I don't see why, these just correspond to two properties in the MCP client builder, and all other properties of it are in this config group too.

geoand: 👌🏽

PR #2253, opened 2026-03-19
Add a new Vertex AI module using as REST client - Quarkus instead of gRPC and endpoint: rawPredict

cmoulliard:
- New module able to handle on Google Cloud Vertex AI platform partner models like Anthropic Claude using as endpoint: `rawPredict` - https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/projects.locations.endpoints/rawPredict
- Use QuarkusRestBuilder as HTTPClient handling HTTP requests instead of gRPC

### DIFF
--- a/model-providers/google/vertex-ai-models/runtime/pom.xml
+++ b/model-providers/google/vertex-ai-models/runtime/pom.xml
@@ -25,10 +25,6 @@
         </dependency>
         <dependency>
             <groupId>dev.langchain4j</groupId>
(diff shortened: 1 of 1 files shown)


geoand:
You're introducing `quarkus-langchain4j-vertex-ai-anthropic` which is very specific to what you are trying to do. What I am saying is that it should live in your project instead of being a new module here.

geoand:
If we can have a common layer for both API, that would be great

PR #2254, opened 2026-03-20
Add a sample SkillsSystemMessageProvider

jmartisk:
The goal is to allow injecting information about available skills into the system message. This PR adds a sample `SystemMessageProvider` that can do it. This way, the user no longer has to provide the exact skill name because the LLM will be aware of the skills that are available.

### DIFF
--- a/samples/skills/pom.xml
+++ b/samples/skills/pom.xml
@@ -18,8 +18,7 @@
         <quarkus.platform.version>3.27.2</quarkus.platform.version>
         <skipITs>true</skipITs>
         <surefire-plugin.version>3.2.5</surefire-plugin.version>
-        <!-- TODO: update when we have a released version that supports skills -->
-        <quarkus-langchain4j.version>1.8.2</quarkus-langchain4j.version>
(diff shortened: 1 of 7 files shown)


samples/skills/pom.xml
```diff
@@ -19,7 +19,7 @@
         <skipITs>true</skipITs>
         <surefire-plugin.version>3.2.5</surefire-plugin.version>
         <!-- TODO: update when we have a released version that supports skills -->
-        <quarkus-langchain4j.version>1.8.2</quarkus-langchain4j.version>
+        <quarkus-langchain4j.version>999-SNAPSHOT</quarkus-langchain4j.version>
```
geoand, line 21: Let's revert this :)

jmartisk: I can't because the sample now depends on a class that is present only in the snapshot

jmartisk: After releasing, this will be automatically updated to 1.8.3, so I think it's not a problem

geoand: Ok, let's see how it goes

Issue #2258, opened 2026-03-23
Support Anthropic Tool Search Tool (defer_loading)


cekocvetkov:
In November 2025, anthropic introduced [the Tool Search Tool feature](https://github.com/quarkiverse/quarkus-langchain4j/issues/2258#issue-4119051007), which allows tools to be deferred and discovered on demand instead of loaded upfront in the context window. This can save a lot of tokens. Plain langchain4j already supports this via:
```
...
        .beta("advanced-tool-use-2025-11-20")
        .serverTools(toolSearchTool)
        .toolMetadataKeysToSend("defer_loading") // need to specify it explicitly
...
```

Would be nice if the quarkus-langchain4j anthropic extension exposed these somehow as `application.properties` keys.

PR #2259, opened 2026-03-23
Add Support for Anthropic Advanced Tool Use Features (#2258)

cekocvetkov:
## Support for Anthropic Advanced Tool Use Features #2258

In November 2025, Anthropic introduced [Advanced Tool Use](https://www.anthropic.com/engineering/advanced-tool-use), a set of capabilities that allow tools to be used more efficiently without bloating the context window. This PR implements support for all three features, plus the underlying metadata plumbing needed to make them work.

### Added configuration for the following application properties:
```
# Tool Search — on-demand tool discovery
quarkus.langchain4j.anthropic.tool-search.enabled=true
quarkus.langchain4j.anthropic.tool-search.type=bm25  # Optional: bm25, regex (default)

# Programmatic Tool Calling — tools invoked from Claude-generated code
quarkus.langchain4j.anthropic.programmatic-tool-calling.enabled=true

# Tool Use Examples — concrete input examples alongside tool schemas
quarkus.langchain4j.anthropic.tool-use-examples.enabled=true
```

---

### Tool Search Tool

Allows Claude to discover tools on demand instead of loading all of them upfront into the context window — essential when working with large tool sets.

When enabled, this automatically:
- Adds the tool search server tool (`bm25` or `regex`, defaulting to `regex`)
- Sets the required `anthropic-beta` header
- Enables the `defer_loading` tool metadata key

Tools that should be deferred must be annotated with:
```java
@Tool(metadata = "{\"defer_loading\": true}")
```

---

### Programmatic Tool Calling

Allows Claude to orchestrate multiple tool calls from within generated Python code, keeping intermediate results out of the context window rather than accumulating them in the conversation.

When enabled, this automatically:
- Adds the `code_execution_20250825` server tool
- Sets the required `anthropic-beta` header
- Enables the `allowed_callers` tool metadata key

Tools that should be callable from code must be annotated with:
```java
@Tool(metadata = "{\"allowed_callers\": [\"code_execution_20250825\"]}")
```

---

### Tool Use Examples

Provides a standard way to attach concrete input examples to tool definitions, helping Claude learn correct parameter usage, formats, and conventions that JSON Schema alone can't express.

When enabled, this automatically:
- Sets the required `anthropic-beta` header
- Enables the `input_examples` tool metadata key

Tools with examples must be annotated with:
```java
@Tool(metadata = "{\"input_examples\": [{...}, ...]}")
```

---

### Tool Metadata Handling (prerequisite fix)

While implementing the above, I noticed that the `metadata` field on `@Tool` was never actually being passed through. Fixed that as a prerequisite:

- Updated `ToolProcessor` to extract the `defer_loading` (and other) metadata properties from the `@Tool` annotation
- Adjusted `ToolSpecificationObjectSubstitution` to properly handle tool metadata

All three features share the same `anthropic-beta: advanced-tool-use` header and the metadata key forwarding mechanism introduced by this fix.

- fixes https://github.com/quarkiverse/quarkus-langchain4j/issues/2258

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/ToolProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/ToolProcessor.java
@@ -41,6 +41,9 @@ import org.objectweb.asm.ClassVisitor;
 import org.objectweb.asm.Opcodes;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
 
 import dev.langchain4j.agent.tool.ReturnBehavior;
 import dev.langchain4j.agent.tool.Tool;
@@ -102,6 +105,7 @@ import io.quarkus.gizmo.ResultHandle;
 public class ToolProcessor {
 
     private static final DotName TOOL = DotName.createSimple(Tool.class);
+
     private static final DotName TOOL_MEMORY_ID = DotName.createSimple(ToolMemoryId.class);
     private static final DotName JSON_IGNORE = DotName.createSimple(JsonIgnore.class);
     private static final DotName INVOCATION_PARAMETERS = DotName.createSimple(InvocationParameters.class);
@@ -340,6 +344,8 @@ public class ToolProcessor {
 
                     String toolName = getToolName(nameValue, toolMethod);
                     String toolDescription = getToolDescription(descriptionValue);
+
+                    AnnotationValue metadataValue = instance.value("metadata");
                     AnnotationValue returnBehavior = instance.value("returnBehavior");
                     ReturnBehavior returnBehaviorEnum = ReturnBehavior.TO_LLM;
                     if (returnBehavior != null) {
@@ -380,6 +386,20 @@ public class ToolProcessor {
                                     .addProperties(properties)
                                     .required(required)
                                     .build());
+
+                    if (metadataValue != null) {
+                        String metadataJson = metadataValue.asString();
+                        if (!metadataJson.isEmpty()) {
+                            try {
+                                Map<String, Object> toolMetadata = ObjectMapperHolder.OBJECT_MAPPER.readValue(metadataJson,
+                                        ObjectMapperHolder.MAP_TYPE_REF);
+                                builder.metadata(toolMetadata);
+                            } catch (JsonProcessingException e) {
+                                throw new ValidationException("Invalid metadata JSON for tool " + toolName + " in " + className,
+                                        e);
+                            }
+                        }
+                    }
 
                     Map<String, Integer> nameToParamPosition = toolMethod.parameters().stream().collect(
                             Collectors.toMap(MethodParameterInfo::name, i -> Integer.valueOf(i.position())));
@@ -1094,5 +1114,11 @@ public class ToolProcessor {
                 .map(type -> type.name().toString())
                 .distinct()
                 .toList();
+    }
+
+    private static class ObjectMapperHolder {
+        private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+        private static final TypeReference<Map<String, Object>> MAP_TYPE_REF = new TypeReference<>() {
+        };
     }
 }
--- a/model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
+++ b/model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
@@ -3,6 +3,10 @@ package io.quarkiverse.langchain4j.anthropic.runtime;
 import static io.quarkiverse.langchain4j.runtime.OptionalUtil.firstOrDefault;
 
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -12,6 +16,7 @@ import jakarta.enterprise.util.TypeLiteral;
 import org.jboss.logging.Logger;
 
 import dev.langchain4j.model.anthropic.AnthropicChatModel;
+import dev.langchain4j.model.anthropic.AnthropicServerTool;
 import dev.langchain4j.model.anthropic.AnthropicStreamingChatModel;
 import dev.langchain4j.model.chat.ChatModel;
 import dev.langchain4j.model.chat.DisabledChatModel;
@@ -21,6 +26,7 @@ import dev.langchain4j.model.chat.listener.ChatModelListener;
 import io.quarkiverse.langchain4j.anthropic.QuarkusAnthropicClient;
 import io.quarkiverse.langchain4j.anthropic.runtime.config.ChatModelConfig;
 import io.quarkiverse.langchain4j.anthropic.runtime.config.LangChain4jAnthropicConfig;
+import io.quarkiverse.langchain4j.anthropic.runtime.config.ToolSearchType;
 import io.quarkiverse.langchain4j.runtime.NamedConfigUtil;
 import io.quarkus.arc.SyntheticCreationalContext;
 import io.quarkus.runtime.RuntimeValue;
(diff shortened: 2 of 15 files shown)


core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/ToolProcessor.java
```diff
@@ -102,6 +105,8 @@
 public class ToolProcessor {
 
     private static final DotName TOOL = DotName.createSimple(Tool.class);
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
geoand: Let's please use the holder class pattern to avoid initializing this unless it's absolutely necessary

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/ToolProcessor.java
```diff
@@ -381,6 +388,21 @@ public void handleTools(
                                     .required(required)
                                     .build());
 
+                    if (metadataValue != null) {
+                        String metadataJson = metadataValue.asString();
+                        if (!metadataJson.isEmpty()) {
+                            try {
+                                Map<String, Object> toolMetadata = OBJECT_MAPPER.readValue(metadataJson,
+                                        new TypeReference<Map<String, Object>>() {
+                                        });
```
geoand: This should also be part of the holder mentioned above

model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
```diff
@@ -103,9 +107,27 @@ public Function<SyntheticCreationalContext<ChatModel>, ChatModel> chatModel(Stri
...
+            ChatModelConfig.ToolSearchConfig toolSearchConfig = chatModelConfig.toolSearch();
+            if (Boolean.TRUE.equals(toolSearchConfig.enabled())) {
+                ToolSearchType type = ToolSearchType.from(toolSearchConfig.type());
+                AnthropicServerTool toolSearchTool = AnthropicServerTool.builder()
+                        .type(type.getToolType())
+                        .name(type.getToolName())
+                        .build();
+                builder.serverTools(toolSearchTool);
+                builder.toolMetadataKeysToSend("defer_loading");
+                betas.add("advanced-tool-use-2025-11-20");
```
geoand: Is this really necessary?

model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
```diff
@@ -191,9 +213,27 @@ public Function<SyntheticCreationalContext<StreamingChatModel>, StreamingChatMod
...
+            }
+
+            if (!betas.isEmpty()) {
+                builder.beta(String.join(",", betas));
```
geoand, line 288: This code seems to be duplicated. Let's please get rid of the duplications

model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
cekocvetkov: Yes, see:

[Introducing advanced tool use on the Claude Developer Platform # Getting started
](https://www.anthropic.com/engineering/advanced-tool-use)
```
client.beta.messages.create(
    betas=["advanced-tool-use-2025-11-20"],
    model="claude-sonnet-4-5-20250929",
    max_tokens=4096,
    tools=[
        {"type": "tool_search_tool_regex_20251119", "name": "tool_search_tool_regex"},
        {"type": "code_execution_20250825", "name": "code_execution"},
        # Your tools with defer_loading, allowed_callers, and input_examples
    ]
)
```

[Langchain4j Tool Search Tool](https://github.com/langchain4j/langchain4j/blob/main/docs/docs/integrations/language-models/anthropic.md#tool-search-tool)
```
ChatModel chatModel = AnthropicChatModel.builder()
        .apiKey(System.getenv("ANTHROPIC_API_KEY"))
        .modelName(CLAUDE_SONNET_4_5_20250929)
        .beta("advanced-tool-use-2025-11-20")
        .serverTools(toolSearchTool)
        .toolMetadataKeysToSend("defer_loading") // need to specify it explicitly
        .logRequests(true)
        .logResponses(true)
        .build();
```

[AnthropicChatModel.java#L321](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicChatModel.java#L321)

---
Or did you mean something else, here?

geoand: Got it, thanks

model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
cekocvetkov: Yeah, my changes here were the latest additions to this already duplicated pattern, that's why I decided to leave it like this. 

Should I try to refactor the whole method and extract _all_ the duplicated logic (also from before my changes)? Or maybe just the parts for`ThinkingConfig` + the new `ToolSearchConfig`?

geoand: Yes please

cekocvetkov: Sorry, to ping on this again.

I had a closer look here and think we should keep the duplication pattern. A refactoring **_is_** possible but would be very messy and not readable at all. At least the ones I could think of. In reality, the code will become more annoying to maintain.
The main problem is that the `AnthropicChatModel.Builder` and `AnthropicStreamingChatModel.Builder` don't have a common supertype. 

I also checked the upstream `langchain4j` implementation for both the [AnthropicChatModel](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicChatModel.java) and [AnthropicStreamingChatModel](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicStreamingChatModel.java) builders. They follow this same duplication pattern internally. 

If we really want to clean up the duplication, I would suggest opening a separate issue for this and handle it separately from this PR.

I might be off, though, what do you think?

geoand: Okay, let's keep it then

cekocvetkov:
Thanks for the feedback! Really appreciate it! 

Btw, my local build fails on the documentation generation:
`[INFO] Quarkus LangChain4j - Documentation ................ FAILURE [  6.345 s]` not really sure what the problem is there. Maybe the docker set up on my mac? I guess we'll know more when running the github CI workflow.

**Anyways, the implementation seems to work well:**

[Here](https://github.com/cekocvetkov/quarkus-langchain4j-tool-search-demo) is a minimal demo project if you guys want to play around with this. Just build the `quarkus-langchai4j` project locally, and the local snapshot will be used in the demo as noted here: 
[`implementation 'io.quarkiverse.langchain4j:quarkus-langchain4j-anthropic:999-SNAPSHOT'`
](https://github.com/cekocvetkov/quarkus-langchain4j-tool-search-demo/blob/main/build.gradle#L15)

---
Please note the Anthropic ["When to use.." section](https://www.anthropic.com/engineering/advanced-tool-use):
> Use it when:
> 
> Tool definitions consuming >10K tokens
> Experiencing tool selection accuracy issues
> Building MCP-powered systems with multiple servers
> 10+ tools available
> Less beneficial when:
> 
> Small tool library (<10 tools)
> All tools used frequently in every session
> Tool definitions are compact  

In my tests, having 30-40 tools already shows a significant token reduction (~35%), even with compact LangChain4j-generated schemas.


Let me know if there’s anything else you’d like changed 👍 

P.S.: I guess we can squash the commits when merging to main, right?

geoand:
🙏🏽 

@jmartisk is there anything you would like to add ?

PR #2288, opened 2026-03-30
Add MongoDB embedding store support

MoheyEl-DinBadr:


embedding-stores/mongodb/deployment/src/main/java/io/quarkiverse/langchain4j/mongodb/deployment/MongoDBEmbeddingStoreProcessor.java
```diff
@@ -0,0 +1,136 @@
...
+                .map(FeatureBuildItem::getName)
+                .map(MONGODB_MEMORY_STORE_FEATURE::equals)
+                .filter(isMemoryStoreFeature -> isMemoryStoreFeature)
+                .findFirst()
+                .map(aBoolean -> ConfigProvider.getConfig()
+                        .getOptionalValue("quarkus.langchain4j.memory-store.mongodb.client-name", String.class))
+                .map(memoryStoreClientConfigName -> memoryStoreClientConfigName.orElse("<default>"))
+                .map(memoryStoreClientConfigName -> memoryStoreClientConfigName.equals(clientName))
+                .orElse(false);
+    }
```
geoand, line 87: IMHO, this is not readable at all

embedding-stores/mongodb/deployment/src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+quarkus.mongodb.devservices.image-name=moheyeldinbadr/mongodb-community-search:8.2.6-rc0-ubuntu2204-slim-20260318T063436Z
```
geoand: We can't rely on such images

embedding-stores/mongodb/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,12 @@
...
+description: "Provides the MongoDB embedding store for LangChain4j"
+metadata:
+  keywords:
+    - "ai"
+    - "langchain4j"
+    - "mongodb"
+  guide: "https://docs.quarkiverse.io/quarkus-langchain4j/dev/index.html"
+  categories:
+    - "ai"
+  status: "preview"
```
geoand:
```suggestion
  status: "experimental"
```

geoand (commented):
Thanks for this!

I've added some comments, but @jmartisk may have more

embedding-stores/mongodb/deployment/src/main/resources/application.properties
MoheyEl-DinBadr: For the image I had to create this one as there is no image that support the mongo-community and mongot (for vector search) so I had to create it myself, or I need to use docker compose but not sure how to use it here at an extention.

The other solution is that I disable the devservices for mongo, and create my own which is going to use mongo-atlas-local testContainers

geoand: I understand, but we can't rely on images that are not used widely

embedding-stores/mongodb/deployment/src/main/java/io/quarkiverse/langchain4j/mongodb/deployment/MongoDBEmbeddingStoreProcessor.java
MoheyEl-DinBadr: this condition is there so there would be no conflict of mongo beans between this extention and the one from memorystore, so basicallyu what it does if the mongo-memory-store feature is there we check if they refer tp same connection so it won't create same connection as mongodb memory store

MoheyEl-DinBadr: but will try to make it more readable and adding some documentation to it

embedding-stores/mongodb/deployment/src/main/resources/application.properties
MoheyEl-DinBadr: I will try first docker compose approach, if it is not successful I will create a specific devservices for this extension and set the mongo one to disabled

MoheyEl-DinBadr: but I belive either of the 2 options I will have to use the Mongo-atlas-local image from  mongodb as it has the mongot there out of the box, and I can't use this one with the normal mongo as test-containers for mongodb is trying to start a replicaset which is already starting at the starting of mongo-atlas-local image so it fails to start the app as it return an error that it is already initiated

geoand: Sorry for my ignorance, but why is Mongo Atlas needed? I mean, doesn't stock MongoDB support embeddings out of the box?

MoheyEl-DinBadr: no, the stock version dosen't support embeddings (vector search) for eaxamle this is an error when trying to use the stock Mongo with version 8.3 

`Caused by: com.mongodb.MongoCommandException: Command failed with error 6047401 (Location6047401): '$listSearchIndexes stage is only allowed on MongoDB Atlas' on server localhost:32769. The full response is {"ok": 0.0, "errmsg": "$listSearchIndexes stage is only allowed on MongoDB Atlas", "code": 6047401, "codeName": "Location6047401", "$clusterTime": {"clusterTime": {"$timestamp": {"t": 1775467790, "i": 1}}, "signature": {"hash": {"$binary": {"base64": "AAAAAAAAAAAAAAAAAAAAAAAAAAA=", "subType": "00"}}, "keyId": 0}}, "operationTime": {"$timestamp": {"t": 1775467790, "i": 1}}}`

so embeddings are only supported either in MongoAtlas or Mongo-community image 
but for MongoCommunity version you need to have mongot in place so you either use a docker compose or create a custom image that include both mongo and mongot

geoand: And there are no widely used Docker images with mongot?

MoheyEl-DinBadr: Sadly no, it is either docker compose or the atlas version

https://medium.com/@bhat.darshan9/my-first-local-vector-search-mongodb-community-edition-8626f134c1d2

geoand: I see, thanks. @jmartisk what's your take on this?

jmartisk: I haven't been following this (sorry, bit busy) so I might be out of touch, but why are we reinventing from scratch something that is in langchain4j already? https://github.com/langchain4j/langchain4j/tree/main/langchain4j-mongodb-atlas - there it is assumed that MongoDB Atlas will be used

jmartisk: For PgVector, we use the devservices provided by the normal postgres extension, but override the used image to be pgvector. Can't we do the same here and override mongodb's image to `mongodb/mongodb-atlas-local`?

geoand: > but why are we reinventing from scratch something that is in langchain4j already?

I had no idea it was there, thanks!

jmartisk: Yeah I think we could just make a thin wrapper over the implementation from langchain4j. The exact same way that we do it for Milvus, Neo4j and PgVector. Sorry I should have noticed this sooner...

geoand: That's definitely the way to go!

jmartisk: When I don't respond to review requests, just poke me, I sometimes tend to overlook stuff when my inbox is full :)

MoheyEl-DinBadr: for the wrapper part it is doable, I can start working on it.

The part related to devservices is related to how the normal mongodb container is initialized from test containers as it initializes a single-node replica and then check for the status for it.

While mongodb-atlas-local initializes automatically the replicaSet as it is needed for the functionality of vector search to have replicaStarted, so when TestContainers tries to start the replica set (Which already started by default) the response is that the replica is already inizialized so it fails as the response code is not code 0, so you wither create your own devservices to use TestConatiners atlas



`    @Override
    protected void containerIsStarted(InspectContainerResponse containerInfo, boolean reused) {
        if (!this.shardingEnabled) {
            initReplicaSet(reused);
        }
    }`
    
 The above is from the class MongoDBContainer
  
 and it is not Overriden in the class MongoDBAtlasLocalContainer, so it does not initialzie the replicaSet.
  
  
 or The other solution is to create your own custom image that handles this as I did before

geoand:
Can you please rebase onto `main` because currently it's not possible to review the PR?

MoheyEl-DinBadr:
> Can you please rebase onto `main` because currently it's not possible to review the PR?

The last change was actually a rebase, so I will mainly create a new PR to remove those commits, but the main thing for me before starting a new PR is Ibam going to recreate it as a wrapper of mongo-atlas from langchain4j.

But the main thing now, is what is going to happen regarding the devservices, becuase as I mentioned before we can't use mongo-atlas-local image directly it would fail with testContainers Mongo, we need to use the TestContainer atlas as the thing related to replicaSet intiation command sent during container startup in Mongo testConatiners 

So do I
- Create a new devservices for mongo-atlas and disable thr one for mongo if we use this extension
- or create a custom image and use it with the mongo testContainers to accommodate for the issue

geoand:
> The last change was actually a rebase

Something must not have gone right, as a correct rebase would only result in the the commits from your PR showing up in the UI.

> So do I

My suggestion would be to just document what needs to be done for a user to use the Dev Service. I don't want us to be shipping add-hoc images

geoand:
Can you please rebase onto main? Hopefully CI will pass as a result of the rebase

geoand:
Please rebase your branch onto the latest `main` instead of merging `main` into your branch.

PR #2327, opened 2026-04-13
Propagate managed parameters when creating the llm request invocation context

mariofusco:


jmartisk (approved):

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -172,6 +173,7 @@ public Object implement(Input input) {
                 .methodArguments((methodArgs != null) ? Arrays.asList(methodArgs) : List.of())
                 .chatMemoryId(memoryId(createInfo, methodArgs, context.hasChatMemory()))
                 .invocationParameters(findInvocationParams(methodArgs))
+                .managedParameters(LangChain4jManaged.current())
```
geoand, line 176: Are you guys sure this is going to work? 

I don't know what this does, but seeing `ThreadLocal` doesn't make me too confident...

mariofusco: At the moment this is used to pass the `AgenticScope` from the agent to the underlying AIService. Passing through a ThreadLocal (admittedly ugly) was unfortunately the only option, because the agent calls the AIService through the user defined interface and there's no way to change the signature and pass an additional argument. The invocation from the agent to the AiService is always on the same thread (there's no reason to fork a different thread for this, since the eventual parallel execution of agents is managed at the agentic framework level), so this shouldn't be a problem.

More in general, the idea behind this `ThreadLocal`-based `LangChain4jManaged` object, is providing a mechanism to mark components internally managed by the framework that have to be passed to the AiService to do its work, but cannot be passed explicitly for the same reason why I couldn't pass the `AgenticScope` in all the cases the Aiservice invocation should happen on the same thread, so this should be safe as well.

geoand: Got it, thanks

PR #2328, opened 2026-04-13
Rework tracing into a McpClientListener

### LINKED ISSUE #2221: Rework MCP client tracing
Move the tracing logic from a custom tool executor into an implementation of `dev.langchain4j.mcp.client.McpClientListener` and make sure that the context is propagated to server too: https://modelcontextprotocol.io/community/seps/414-request-meta - this will need some changes in upstream langchain4j too (passing `_meta` fields to the request JSON)

jmartisk:
- Fixes: https://github.com/quarkiverse/quarkus-langchain4j/issues/2221


jmartisk:
Hi @brunobat , we talked about this at the f2f. Reviews welcome :)

integration-tests/mcp/src/test/java/io/quarkiverse/langchain4j/mcp/test/McpTracingTest.java
```diff
@@ -0,0 +1,98 @@
...
+    @Test
+    public void executeToolCreatesSpanWithProperAttributes() {
+        ToolExecutionRequest request = ToolExecutionRequest.builder()
+                .name("echoMeta")
+                .arguments("{\"key\": \"traceparent\"}")
+                .build();
+        String result = mcpClient.executeTool(request).resultText();
+
+        // Verify that TracingMcpClientListener created a span with the correct name and attributes
+        await().untilAsserted(() -> assertThat(spanExporter.getFinishedSpanItems())
```
brunobat: Should have a timeout or you might stuck the build

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
```diff
@@ -0,0 +1,141 @@
...
+        SpanAndScope sas = activeSpans.remove(context);
+        if (sas == null) {
+            log.warn("Unknown call context: " + context.message().getId());
+            return;
+        }
+        if (result != null && result.isError()) {
+            sas.span.setAttribute(ERROR_TYPE, "tool_error");
+            sas.span.setStatus(StatusCode.ERROR);
+        }
+        sas.scope.close();
```
cescoffier: Are we sure it will work is beforeExecuteTool and afterExecuteTool are invoked from different threads?

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
```diff
@@ -0,0 +1,141 @@
...
+        sas.span.recordException(error);
+        sas.scope.close();
+        sas.span.end();
+    }
+
+    @Override
+    public void beforeResourceGet(McpCallContext context) {
+        McpReadResourceParams params = (McpReadResourceParams) ((McpReadResourceRequest) context.message()).getParams();
+        String uri = params.getUri();
+        Span span = tracer.spanBuilder("langchain4j.mcp-resources.read")
```
cescoffier: The name is quite different from the `execute_tool`, did you discuss having something closer, like:
`quarkus.langchain4j.resource_read`?

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
```diff
@@ -0,0 +1,141 @@
...
+    @Override
+    public void onResourceGetError(McpCallContext context, Throwable error) {
+        endSpan(context, error);
+    }
+
+    @Override
+    public void beforePromptGet(McpCallContext context) {
+        McpGetPromptParams params = (McpGetPromptParams) ((McpGetPromptRequest) context.message()).getParams();
+        String promptName = params.getName();
+        Span span = tracer.spanBuilder("langchain4j.mcp-prompts." + promptName)
```
cescoffier: Same here, the name is quite different from the execute_tool, did you discuss having something closer, like:
quarkus.langchain4j.prompt_get?

BTW, the prompName should probably be a tag, not in the name of the span.

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Override
+    public void afterPromptGet(McpCallContext context, McpGetPromptResult result, Map<String, Object> rawResult) {
+        endSpan(context, null);
+    }
+
+    @Override
+    public void onPromptGetError(McpCallContext context, Throwable error) {
+        endSpan(context, error);
```
cescoffier, line 129: I would add:

```
sas.span.setAttribute(ERROR_TYPE, error.getClass().getName());
sas.span.setStatus(StatusCode.ERROR, error.getMessage());
```

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Override
+    public void afterResourceGet(McpCallContext context, McpReadResourceResult result, Map<String, Object> rawResult) {
+        endSpan(context, null);
+    }
+
+    @Override
+    public void onResourceGetError(McpCallContext context, Throwable error) {
+        endSpan(context, error);
```
cescoffier, line 104: I would add:

```
sas.span.setAttribute(ERROR_TYPE, error.getClass().getName());
sas.span.setStatus(StatusCode.ERROR, error.getMessage());
```

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private void endSpan(McpCallContext context, Throwable error) {
+        SpanAndScope sas = activeSpans.remove(context);
+        if (sas == null) {
+            log.warn("Unknown call context: " + context.message().getId());
+            return;
+        }
+        if (error != null) {
+            sas.span.recordException(error);
```
cescoffier: Add:
```
sas.span.setAttribute(ERROR_TYPE, error.getClass().getName());
          sas.span.setStatus(StatusCode.ERROR, error.getMessage());
          sas.span.recordException(error);
      }
```

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
```diff
@@ -0,0 +1,141 @@
...
+ */
+public class TracingMcpClientListener implements McpClientListener {
+
+    // GenAI semantic convention attribute keys
+    private static final AttributeKey<String> GEN_AI_OPERATION_NAME = AttributeKey.stringKey("gen_ai.operation.name");
+    private static final AttributeKey<String> GEN_AI_TOOL_NAME = AttributeKey.stringKey("gen_ai.tool.name");
+    private static final AttributeKey<String> ERROR_TYPE = AttributeKey.stringKey("error.type");
+
+    private final Tracer tracer;
+    private final Map<McpCallContext, SpanAndScope> activeSpans = Collections.synchronizedMap(new IdentityHashMap<>());
```
cescoffier, line 46: I think we need to be careful not to have a leak here.
If `before` is called, but neither `after` nor `error` (because of some unhandled event), we may end up with a leak. 

Did you think about using a weak HashMap?

cescoffier (changes requested):

integration-tests/mcp/pom.xml
```diff
@@ -22,6 +22,15 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-micrometer-registry-prometheus</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-opentelemetry</artifactId>
```
brunobat: You should duplicate the test module to make sure things still work without OTel.

mcp/deployment/src/main/java/io/quarkiverse/langchain4j/mcp/deployment/McpProcessor.java
```diff
@@ -158,6 +154,17 @@ public void registerMcpClients(McpBuildTimeConfiguration mcpBuildTimeConfigurati
             runtimeInitializedClasses.produce(
                     new RuntimeInitializedClassBuildItem("io.quarkiverse.langchain4j.mcp.runtime.MetricsMcpListener"));
         }
+        boolean openTelemetryPresent = capabilities.isPresent(Capability.OPENTELEMETRY_TRACER);
```
brunobat: This should use the `OpenTelemetrySdkBuildItem` to verify if the tracing is enabled.

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/McpRecorder.java
```diff
@@ -180,13 +173,64 @@ public McpClient apply(SyntheticCreationalContext<McpClient> context) {
         };
     }
 
-    private void addMetrics(DefaultMcpClient.Builder builder, String key) {
+    private void addObservability(DefaultMcpClient.Builder builder, String key, boolean addMetrics, boolean addTracing) {
```
brunobat, line 179: Only add the listeners if OTel SDK is enabled at runtime

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
```diff
@@ -0,0 +1,141 @@
...
+
+    public TracingMcpClientListener(Tracer tracer) {
+        this.tracer = tracer;
+    }
+
+    @Override
+    public void beforeExecuteTool(McpCallContext context) {
+        McpCallToolParams params = (McpCallToolParams) ((McpCallToolRequest) context.message()).getParams();
+        String toolName = params.getName();
+        Span span = tracer.spanBuilder("execute_tool " + toolName)
```
brunobat: Is the transport layer sending out the current span?

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
```diff
@@ -0,0 +1,141 @@
...
+        endSpan(context, null);
+    }
+
+    @Override
+    public void onPromptGetError(McpCallContext context, Throwable error) {
+        endSpan(context, error);
+    }
+
+    private void endSpan(McpCallContext context, Throwable error) {
+        SpanAndScope sas = activeSpans.remove(context);
```
brunobat, line 137: SpanAndScope sas = activeSpans.remove(context);
        if (sas == null) {
            log.warn("Unknown call context: " + context.message().getId());
            return;
        }
        try(sas.scope) {
            if (error != null) {
                sas.span.recordException(error);
            }
        } finally { 
        sas.span.end();
        }

brunobat: scope and span closing must be always surrounded by try/finally

brunobat (changes requested):
Thanks Jan. 
We need an integration test that will show a server side child span from the client call, to make sure propagation is working.

mcp/deployment/src/main/java/io/quarkiverse/langchain4j/mcp/deployment/McpProcessor.java
jmartisk: So I also need to move `McpBuildTimeConfiguration.tracingEnabled` to runtime?

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
jmartisk: Ah, claude didn't properly catch what semantic conventions I'm referring to, this will be reworked more heavily in a bit

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
jmartisk: I'm reworking this a bit right now. Btw, the prompt name is in the span name:

https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#client

> Span name SHOULD follow the format {mcp.method.name} {target} where target SHOULD match {gen_ai.tool.name} or {gen_ai.prompt.name} when applicable.

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
jmartisk: A leak in the sense that we don't close the span and scope, or that we don't remove it from the map? If we used a weak map, there would be a danger that the `SpanAndScope` disappear before we close them, no?

This is something that will need to be tested for sure, I'm working on tests for various kinds of error conditions, including timeout, that will make sure they are closed.

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
jmartisk: Should be done now

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
jmartisk: What do you mean here? `TracingMcpMetaSupplier` provides the `traceparent` field  that is added to the JSON-RPC messages

jmartisk:
> Thanks Jan. We need an integration test that will show a server side child span from the client call, to make sure propagation is working.

How should that work? In that case we can't use a `InMemorySpanExporter` but have to spin up a regular collector that the MCP server will also be connected to, correct?

UPDATE: Given that we run integration tests against `quarkus-mcp-server`, I assume this will also depend on https://github.com/quarkiverse/quarkus-mcp-server/pull/703 being merged and released there

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
jmartisk: I've unified this into the `endSpan` method, please check

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
jmartisk: I suppose it should, as long as we close the correct scope, right @brunobat ?

integration-tests/mcp-opentelemetry/src/test/resources/logback-test.xml
geoand, line 1: Not important, but do we really need this?

jmartisk: Without it, the logs from the MCP server wouldn't show

geoand: kk, thanks

jmartisk: Hold on, they're not showing anyway. Let me check :D

mcp/deployment/src/main/java/io/quarkiverse/langchain4j/mcp/deployment/McpProcessor.java
brunobat: yes, that should be a runtime config

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/TracingMcpClientListener.java
brunobat: Not sure if it's enough. Will give feedback on this later.

integration-tests/mcp-opentelemetry/src/test/resources/logback-test.xml
jmartisk: Ok I've removed it. It really isn't doing anything. Streamable HTTP server logs show up, and stdio show up after adjusting logging levels a bit.

geoand: Thanks!

geoand:
Can we please squash the commits on this?

jmartisk:
Yeah I'm waiting for some more reviews, I split it into commits because I thought people might find it easier to review, of course I'll squash afterwards

geoand:
🙏🏽

mcp/deployment/src/main/java/io/quarkiverse/langchain4j/mcp/deployment/McpProcessor.java
jmartisk: Ok, I did that yesterday already in https://github.com/quarkiverse/quarkus-langchain4j/pull/2328/changes/6cd5bbb90b96c30e3c65c78f75bef1759527ac3d

brunobat:
> UPDATE: Given that we run integration tests against `quarkus-mcp-server`, I assume this will also depend on [quarkiverse/quarkus-mcp-server#703](https://github.com/quarkiverse/quarkus-mcp-server/pull/703) being merged and released there

yes @jmartisk, I'm working on the MCP context propagation in the server side now. That can later be used to validate the client. Will ping you there once ready for you to take a look.

geoand:
Are we ready to merge this?

brunobat:
> Are we ready to merge this?

Let me give it a test drive with the quarkus mcp instrumentation PR first.

brunobat:
We still need to figure out a few things.

1. **Callbacks**. I think the client is not sending the `traceparent` in the `notifications/initialized` metadata:

image
The server should be propagating the `traceparent in the callbaks.

2. **Span names**. 

image

MCP span names must follow `{mcp.method.name} {target}` where target SHOULD match {gen_ai.tool.name} or {gen_ai.prompt.name}. In the above example the request starts with a GET I did in the browser, and it's fine, but later we see:

image

And They don't follow the convention. We know one of them is asking for the `tools/list` because it's what the server is doing. 

3. Ping is not sending the tracing data. I'm not sure resources and prompts send metadata as well.

brunobat (commented):
We need more tests and instrument other target types.

Did the test with this class:
```java
@Path("/mcp")
public class McpResource {

    @Inject
    @McpClientName("client1")
    McpClient mcpClient;

    @GET
    @Path("/tools")
    public List<ToolSpecification> listTools() {
        return mcpClient.listTools();
    }

    @GET
    @Path("/tools/execute")
    public String executeTool(@QueryParam("name") String name, @QueryParam("arguments") String arguments) {
        ToolExecutionRequest request = ToolExecutionRequest.builder()
                .name(name)
                .arguments(arguments != null ? arguments : "{}")
                .build();
        return mcpClient.executeTool(request).resultText();
    }

    @GET
    @Path("/resources")
    public String readResource(@QueryParam("uri") String uri) {
        return mcpClient.readResource(uri).toString();
    }

    @GET
    @Path("/prompts")
    public String getPrompt(@QueryParam("name") String name) {
        return mcpClient.getPrompt(name, Collections.emptyMap()).toString();
    }
}
``` 

and this config:

```properties
quarkus.langchain4j.mcp.client1.transport-type=streamable-http
quarkus.langchain4j.mcp.client1.url=http://localhost:8080/mcp
quarkus.langchain4j.mcp.client1.log-requests=true
quarkus.langchain4j.mcp.client1.log-responses=true
quarkus.langchain4j.mcp.client1.tool-execution-timeout=3s
quarkus.http.port=8090
quarkus.application.name=mcp-client
quarkus.otel.logs.enabled=true
quarkus.otel.metrics.enabled=true
```

jmartisk:
There's a lot of message types that are simply not yet covered in the `McpClientListener` interface (pings, tools/list, the initialization notification etc.). I'll try to make a list and file that in the upstream langchain4j project. I wouldn't block merging the initial implementation because of that.

geoand:
> I wouldn't block merging the initial implementation because of that.

Agreed

jmartisk:
I've reported https://github.com/langchain4j/langchain4j/issues/4953 for this and will try to implement that next week, but anyway it will need a langchain4j release.

brunobat:
Other issues not related with message types should be addresses before that release, like the span names.

jmartisk:
These spans named `POST` are coming from HTTP calls that are made because you're using the Streamable HTTP transport for your test. For example, after switching to the STDIO transport, these don't appear at all. They're not created by the MCP client library, but the Vert.x HTTP client. When we properly integrate tracing for the other operation types, then the new MCP spans will wrap around these `POST` ones (for HTTP-based transports).

I guess we could expand the test coverage here to covert Streamable HTTP in addition to STDIO here, and the set of expected spans will be different..

cescoffier (approved):

cescoffier:
Don't forget to document the span name change. 

`resource/read` does not seem to be complete (no target?)

jmartisk:
> `resource/read` does not seem to be complete (no target?)

Hmm. From the semantic conventions for resource reads:

> Instrumentation MAY allow users to opt into including {mcp.resource.uri} as target in the span name when it is available but SHOULD NOT include it by default to avoid high cardinality span names.

So this is an optional thing that would be activated by configuration. I don't think it's very important  though because the URI is in an attribute anyway...

brunobat:
> > `resource/read` does not seem to be complete (no target?)
> 
> Hmm. From the semantic conventions for resource reads:
> 
> > Instrumentation MAY allow users to opt into including {mcp.resource.uri} as target in the span name when it is available but SHOULD NOT include it by default to avoid high cardinality span names.
> 
> So this is an optional thing that would be activated by configuration. I don't think it's very important though because the URI is in an attribute anyway...

I agree that we should avoid high cardinality .

In relation to the names. This initial implementation lacks the correct span names and many attributes. This is fine for debug purposes only and should be flagged in the docs.
If ppl try to create dashboards the would need to change them later anyway.

jmartisk:
> In relation to the names. This initial implementation lacks the correct span names and many attributes. This is fine for debug purposes only and should be flagged in the docs. 

I checked the semconv table again and I think the only missing required attribute (for the spans that we already support) is `rpc.response.status_code`, I'll add that in a bit.

Otherwise it should be just a matter of adding spans for the missing operation types. 

> If ppl try to create dashboards the would need to change them later anyway.

Why? No. The span names are correct, as per the explanation in my previous [comment](https://github.com/quarkiverse/quarkus-langchain4j/pull/2328#issuecomment-4268355887).

jmartisk:
As for docs, we should somehow mention the breaking change that the spans are now following the proper semantic conventions instead of inventing our own. That sounds like a release-notes thing but we don't currently publish that afaik :/

geoand:
I was planning on doing a `1.9.0` release today. Should we include it?

brunobat (approved):
@geoand Let's include it with a note that this is under development and we will iterate

geoand:
@jmartisk are you okay to have this in 1.9?

jmartisk:
Yeah!

cescoffier:

PR #2330, opened 2026-04-13
Fix callout rendering in tool guardrails guide

rolfedh:
## Summary

- Remove blank line between numbered list and code block in `tool-guardrails.adoc` so the code block stays attached to the list and callouts render correctly

## Test plan

- [ ] Verify callouts `<1>` through `<6>` render correctly in the Quick Start section


rolfedh:
incorrect

PR #2332, opened 2026-04-14
Allow ParallelMapperAgent to transform the type of its output

mariofusco:


jmartisk (approved):

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
```diff
@@ -674,6 +677,20 @@ public void validateAgenticParameterTypes(CombinedIndexBuildItem indexBuildItem,
...
+    }
+
+    private static boolean isParallelMapperOutput(ClassInfo agentClass, String parameterName) {
+        return agentClass.methods().stream().filter(m -> m.annotation(ParallelMapperAgent.class) != null)
+                .findFirst()
+                .flatMap(m -> Optional.ofNullable(m.annotation(ParallelMapperAgent.class).value("outputKey")))
+                .map(AnnotationValue::asString)
+                .map(outputKey -> outputKey.equals(parameterName))
+                .orElse(false);
+    }
```
geoand, line 692: I know that most people disagree with me, but I fail to see how this is more readable than a normal iteration with if statements :)

mariofusco: Sorry, sometimes I'm still (badly?) influenced by my Scala/Haskell background :)

Anyway I'm not strongly opinionated about this, and you're right about keeping the code style consistent. Let me know if I should rewrite it.

geoand: No need to rewrite, just mentioning it :)

jmartisk: +1, I also find this hard to read

PR #2338, opened 2026-04-16
Add Apicurio Registry integration for dynamic MCP server discovery

### LINKED ISSUE #2334: Feature Proposal: MCP Server backed by Apicurio Registry for dynamic tool discovery
## Summary

I'd like to propose an integration between **Quarkus LangChain4j** and **[Apicurio Registry](https://www.apicur.io/registry/)** to enable dynamic tool discovery via the **Model Context Protocol (MCP)**.

Quarkus LangChain4j already has excellent MCP client support — AI services can connect to MCP servers and dynamically discover/invoke tools at runtime. On the other side, Apicurio Registry is an API/Schema registry that already provides versioned artifact storage, schema validation, compatibility enforcement, and [native MCP server support](https://www.apicur.io/blog/2026/02/05/apicurio-registry-ai-natural-evolution).

The idea is to bridge the two: an **MCP server backed by Apicurio Registry** that exposes tool definitions stored as versioned artifacts, so Quarkus LangChain4j AI services can discover tools dynamically with governance guarantees.

## How it would work

1. Tool definitions (name, description, parameter JSON schema, endpoint metadata) are stored as versioned artifacts in Apicurio Registry.
2. The registry's MCP server exposes these tools via the standard MCP protocol (`listTools`, `executeTool`).
3. Quarkus LangChain4j AI services connect to the registry MCP server with standard configuration:

```properties
quarkus.langchain4j.mcp.registry.transport-type=http
quarkus.langchain4j.mcp.registry.url=http://registry:8080/mcp
```

4. Tools can be versioned, grouped, and governed using the registry's existing rule system (compatibility, validity).

## Value

- **Dynamic tool discovery** — no redeployment needed when tools are added or updated.
- **Schema governance** — the registry's compatibility rules (backward, forward, full) prevent breaking changes to tool parameters.
- **Centralized catalog** — all tools across the organization are discoverable in one place.
- **Versioning** — AI services can pin to specific tool versions or track latest.
- **Already partially built** — Apicurio Registry already has MCP server capabilities (see the [AI-native evolution blog post](https://www.apicur.io/blog/2026/02/05/apicurio-registry-ai-natural-evolution)); the integration is primarily about connecting the dots.

## Context

Apicurio Registry is evolving into an AI-native governance platform, applying proven schema management principles to AI agent infrastructure. The blog post ["Apicurio Registry: AI-Native Evolution"](https://www.apicur.io/blog/2026/02/05/apicurio-registry-ai-natural-evolution) describes the vision: the same governance that APIs have today (schema validation, compatibility enforcement, versioning, discovery) applied to AI agents, tools, and prompts. MCP Tool Definition artifacts are part of the Phase 3 roadmap.

Quarkus LangChain4j's existing MCP client support makes it the natural Java-side consumer for this kind of integration.

## Contribution

I'm happy to implement and contribute this integration if people see value in it. I'd appreciate feedback on the approach and any guidance on the preferred integration pattern (extension, standalone module, etc.).

carlesarnal:
New `mcp-apicurio-registry` extension that lets AI services discover and use MCP servers registered in Apicurio Registry at runtime.

**Tools provided:**
- `searchMcpServers(query)` — Searches for `MCP_TOOL` artifacts by description
- `connectMcpServer(artifactId, groupId)` — Reads connection metadata (`mcp-server-url`, `mcp-transport-type`) from artifact labels, creates a Streamable HTTP MCP client, and adds it to the ToolProvider. Now also lists the server's available tools in the response.
- `callMcpTool(serverKey, toolName, arguments)` — Invokes a tool directly on a connected MCP server. This bypasses the tool provider refresh limitation (the AI service resolves tools once at request start, so dynamically added tools aren't visible until the next request).
- `disconnectMcpServer(artifactId, groupId)` — Removes a connected MCP server

**How it works:**
The LLM chains: search → connect → callMcpTool in a single conversation turn. After connecting, the response lists available tools so the LLM knows what to call.

**Configuration:**
- `quarkus.langchain4j.mcp.apicurio-registry.url` — Registry URL

Also patches `McpProcessor` to create a `ToolProvider` bean even without static MCP clients configured, so dynamically discovered servers can be added.

Includes a sample project and MCP guide documentation.

Closes #2334


jmartisk:
Thanks! I believe this should have a separate module instead of living directly in `mcp` to avoid dragging in additional dependencies (I know it's an optional one, but that still comes with potential issues, like native compilation breaking because of it), would it be a big problem to move to a separate module? I mean, it is not a hard requirement, but I think it would better to split if possible. You could make the Apicurio dependency non-optional in that case.

jmartisk:
.. and a sample project demonstrating this would be awesome :)

carlesarnal:
Sure! Will do. I actually doubted between the two approaches.

carlesarnal:
**Note:** The integration test (`ApicurioRegistryMcpToolsTest`) relies on the Apicurio Registry `latest-snapshot` Docker image (`quay.io/apicurio/apicurio-registry:latest-snapshot`) which includes support for the `MCP_TOOL` artifact type and content validation against the official MCP specification.

**Latest changes:**
- MCP_TOOL artifact content now follows the official [MCP specification (2025-11-25)](https://modelcontextprotocol.io/specification/2025-11-25/server/tools) — `name`, `inputSchema` (required), plus optional `description`, `title`, `annotations`, `outputSchema`
- Connection metadata (`mcp-server-url`, `mcp-transport-type`) is stored as **artifact labels**, keeping the artifact content spec-compliant and validated by the registry's content validation rules
- Fixed ToolProvider client proxy issue using `ClientProxy.unwrap()`
- Fixed Testcontainers classloader issue that caused two registry containers to be started

geoand:
Thanks.

Just keep in mind that this now need a rebase

geoand:
Where are we with this?

carlesarnal:
> Where are we with this?

This is essentially done, waiting for a release on our side that includes MCP definition support. I would hold on merging this until the release is done so it can rely on an stable version.

geoand:
@jmartisk PTAL

jmartisk:
Thanks, I will take a look ASAP

samples/apicurio-registry-mcp/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+- A running Apicurio Registry instance (e.g. via `docker run -p 8080:8080 quay.io/apicurio/apicurio-registry:3.3.0`)
+- MCP tool definitions registered as `MCP_TOOL` artifacts in the registry (following the MCP spec), with connection metadata stored as artifact labels (`mcp-server-url`, `mcp-transport-type`)
+- An OpenAI API key (or configure a different model provider)
+
+## Running
+
+1. Start Apicurio Registry:
+   ```bash
+   docker run -p 8080:8080 quay.io/apicurio/apicurio-registry:3.3.0
```
jmartisk: Can we map this to another port? 8080 is used by Quarkus.
Either you need to move this to another port, or add `quarkus.http.port` to move Quarkus to another port, but I think it would be better to move Apicurio.

samples/apicurio-registry-mcp/README.md
```diff
@@ -0,0 +1,40 @@
...
+   ```bash
+   export QUARKUS_LANGCHAIN4J_OPENAI_API_KEY=your-api-key
+   ```
+
+3. Run the sample:
+   ```bash
+   mvn quarkus:dev
+   ```
+
+4. Open http://localhost:8082 in your browser to interact with the chatbot.
```
jmartisk: `8080` if we move apicurio and leave Quarkus at the default port

jmartisk: There doesn't seem to be any UI or anything like that on the `/` path? Is the user supposed to open the dev ui chat? Or could we add a simple UI (copy from other samples)?

mcp-apicurio-registry/deployment/src/test/java/io/quarkiverse/langchain4j/mcp/test/apicurio/MockStreamableHttpMcpServer.java
```diff
@@ -0,0 +1,141 @@
...
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * A mock Streamable HTTP MCP server endpoint for testing.
```
jmartisk: This will be a pain to maintain. Can't we spin up a full-fledged MCP server like we do in tests in the `integration-tests/mcp` module, based on JBang?

mcp-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/apicurio/ApicurioRegistryMcpTools.java
```diff
@@ -0,0 +1,147 @@
...
+        try {
+            client.close();
+        } catch (Exception e) {
+            // best effort
+        }
+        return "Disconnected from MCP server '" + key + "'";
+    }
+
+    private McpClient createMcpClient(String key, McpServerDefinition serverDef) {
+        QuarkusStreamableHttpMcpTransport transport = new QuarkusStreamableHttpMcpTransport.Builder()
```
jmartisk, line 177: How would security work, what if the server requires authentication? Can the registry declare how authentication against the MCP server needs to work?

samples/apicurio-registry-mcp/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## How It Works
+
+The `Bot` AI service is configured with `ApicurioRegistryMcpTools` which provides three tools:
```
jmartisk, line 71: I would appreciate some instructions on how to deploy an actual MCP server definition to the registry, otherwise the sample can't actually do anything useful, no? 
Either we could run a simple MCP server locally as part of the sample (a JBang script), or connect to a remote public server? (the local solution is probably better because depending on external public stuff has proven itself problematic many times)

mcp-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/apicurio/config/McpApicurioRegistryRuntimeConfig.java
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * The base URL of the Apicurio Registry instance
+     * (e.g. http://localhost:8080/apis/registry/v3).
+     */
+    String url();
+
+    /**
+     * Optional authentication token for the registry.
+     */
+    Optional<String> authToken();
```
jmartisk: I don't see this being read anywhere

mcp-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/apicurio/ApicurioRegistryMcpTools.java
jmartisk, line 177: Is it a goal to support STDIO servers too? I don't know if that's in scope, but if yes, you'll need to create a transport of a different type here

samples/apicurio-registry-mcp/README.md
carlesarnal: Done — moved Apicurio Registry to port 8180 (`-p 8180:8080`) and updated `application.properties` accordingly. Quarkus stays on the default 8080.

samples/apicurio-registry-mcp/README.md
carlesarnal: Fixed — README now points to `http://localhost:8080` for the Quarkus app.

carlesarnal: Added a chatbot UI copied from the `samples/chatbot` pattern — `index.html` with `demo-chat.js` and `demo-title.js` web components. Opening `http://localhost:8080` now shows a chat interface connected to the `/chatbot` WebSocket endpoint.

mcp-apicurio-registry/deployment/src/test/java/io/quarkiverse/langchain4j/mcp/test/apicurio/MockStreamableHttpMcpServer.java
carlesarnal: Replaced with a JBang-based MCP server (`simple_mcp_server.java`) following the same pattern used in `integration-tests/mcp`. Added a `McpServerHelper` class (adapted from the integration-tests one) to manage the JBang process lifecycle in `@BeforeAll`/`@AfterAll`.

mcp-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/apicurio/ApicurioRegistryMcpTools.java
carlesarnal: For this initial version, only unauthenticated servers are supported. Authentication could be added as a follow-up by storing auth hints as additional artifact labels (e.g. `mcp-auth-type`, `mcp-auth-header`) and wiring them into the transport via the existing `McpHeadersSupplier` mechanism. Added a comment in `createMcpClient` noting this.

samples/apicurio-registry-mcp/README.md
carlesarnal: Expanded the README with a complete step-by-step guide: start a local JBang MCP server (included as `mcp-server.java` in the sample), register it in the registry via `curl`, run the sample, and interact with the chatbot.

mcp-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/apicurio/config/McpApicurioRegistryRuntimeConfig.java
carlesarnal: Removed. The `ApicurioRegistryMcpToolsBuilder` already has `basicAuth()` and `oauth2()` methods for programmatic use. We can add config-driven registry auth in a follow-up.

mcp-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/apicurio/ApicurioRegistryMcpTools.java
carlesarnal: Not in scope for this version. Registry-discovered servers are expected to be network-accessible, so only streamable-http transport is supported. STDIO would require registering the server binary path, working directory, environment, etc. as labels, which doesn't fit the discovery model well. Added a comment in `createMcpClient` noting this design decision.

samples/apicurio-registry-mcp/README.md
```diff
@@ -0,0 +1,90 @@
...
+```bash
+docker run -p 8180:8080 quay.io/apicurio/apicurio-registry:3.3.0
+```
+
+### 2. Start the sample MCP server
+
+The included `mcp-server.java` JBang script provides simple math tools (add, multiply) and a current-time tool:
+
+```bash
+jbang -Dquarkus.http.port=8085 mcp-server.java
```
jmartisk: You need to rename the file to `mcp_server.java` (not `mcp-server.java`) to match the class name

samples/apicurio-registry-mcp/README.md
```diff
@@ -0,0 +1,90 @@
...
+```bash
+curl -X POST http://localhost:8180/apis/registry/v3/groups/default/artifacts \
+  -H "Content-Type: application/json" \
+  -d '{
+    "artifactId": "calculator-server",
+    "artifactType": "MCP_TOOL",
+    "name": "Calculator Server",
+    "description": "A simple calculator MCP server with add, multiply, and current-time tools",
+    "labels": {
+      "additionalData": {
```
jmartisk: I get an error when executing this:

```
2026-06+00-16 09:01:43 ERROR [org.jboss.resteasy.plugins.providers.jackson.JsonProcessingExceptionMapper] (executor-thread-5) RESTEASY-JACKSON000100: Not able to deserialize data provided: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `java.lang.String` from Object value (token `JsonToken.START_OBJECT`)
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 7, column: 25] (through reference chain: io.apicurio.registry.rest.v3.beans.CreateArtifact["labels"]->java.util.LinkedHashMap["additionalData"])
```

The apicurio docs (https://www.apicur.io/registry/docs/apicurio-registry/3.3.x/assets-attachments/registry-rest-api.htm#tag/Artifacts/operation/createArtifact) also says that the `labels` must be key-value pairs of type string, but here you have `additionalData` that is an object

jmartisk: It's also somehow weird/confusing to me that an artifact of type `MCP_TOOL` is actually a reference to an MCP server, not a specific tool. I wasn't able to find in the docs (https://www.apicur.io/registry/docs/apicurio-registry/3.3.x/getting-started/assembly-mcp-tool-features.html) anything about importing an MCP server though, so I don't know how to fix this

jmartisk: Ah, you fixed this while I was writing the comments, ok, let me have another look.

geoand:
I'll just note that when everything else is done, the commits will need to be squashed

samples/apicurio-registry-mcp/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,70 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+
+    <script src="/_importmap/dynamic-importmap.js"></script>
```
jmartisk, line 7: For this to work, you also need to copy the `ImportmapResource` class from some other sample

mcp-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/apicurio/ApicurioRegistryMcpTools.java
```diff
@@ -0,0 +1,151 @@
...
+
+    public ApicurioRegistryMcpTools(RegistryClient registryClient, McpToolProvider mcpToolProvider, Vertx vertx) {
+        this.registryClient = registryClient;
+        this.mcpToolProvider = mcpToolProvider;
+        this.vertx = vertx;
+    }
+
+    @Tool("Search Apicurio Registry for available MCP servers matching a query. "
+            + "Returns server names, descriptions, and versions.")
+    public String searchMcpServers(String query) {
```
jmartisk: When I tried asking `Search for available MCP servers` and then `Connect to the calculator server`, the second prompt didn't work because this tool tries to find the server using the name only, and the LLM doesn't tend to include the exact name (`Calculator Server`), it tried `calculator`, `Calculator`,... but that doesn't work because the query is for an exact name match. Perhaps it should be possible to query using the groupId and/or artifactId? The way it is now, it would be way too unreliable

mcp-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/apicurio/ApicurioRegistryMcpTools.java
```diff
@@ -0,0 +1,151 @@
...
+    public String connectMcpServer(String groupId, String artifactId) {
+        String key = groupId + "/" + artifactId;
+
+        if (connectedServers.containsKey(key)) {
+            return "Already connected to MCP server '" + key + "'";
+        }
+
+        try {
+            // Read connection metadata from artifact labels
+            ArtifactMetaData metadata = registryClient.groups().byGroupId(groupId)
```
jmartisk: This is a problem because the `curl` command in the instructions omits the `groupId`, so it becomes null, and the LLM tends to pass a `"null"` (as a string!) here and that doesn't find the artifact.

jmartisk: So maybe we could only call `byGroupId` if the `groupId` parameter isn't null, and the parameter description should say that when there is no group ID, the LLM should pass a null (a real null, not a string "null")

samples/apicurio-registry-mcp/README.md
carlesarnal: Done — renamed to `mcp_server.java`.

samples/apicurio-registry-mcp/README.md
carlesarnal: Fixed — the REST API takes labels as a flat `Map<String, String>`, not wrapped in `additionalData`. Updated the curl command in the README.

samples/apicurio-registry-mcp/src/main/resources/META-INF/resources/index.html
carlesarnal: Added `ImportmapResource.java` to the sample, adapted from the chatbot sample.

mcp-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/apicurio/ApicurioRegistryMcpTools.java
carlesarnal: Reworked the search to use the `description` query parameter instead of exact `name` match, which gives much more flexible results. Also added `@P` annotations with clear parameter descriptions to help the LLM produce the right values.

mcp-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/apicurio/ApicurioRegistryMcpTools.java
carlesarnal: Fixed — `groupId` now defaults to `"default"` when null, blank, or the literal string `"null"`. Also swapped parameter order to `(artifactId, groupId)` so the LLM naturally provides the required artifactId first.

geoand:
Please don't forget to squash the commits when this is done

jmartisk:
Sorry for taking a bit longer, I should be able to get back to this very soon :)

mcp/deployment/src/test/java/io/quarkiverse/langchain4j/mcp/test/MultipleMcpClientsTest.java
jmartisk, line 97: What's the purpose of this change? That's not related to this PR, is it?

samples/apicurio-registry-mcp/src/main/java/io/quarkiverse/langchain4j/sample/apicurio/Bot.java
```diff
@@ -0,0 +1,19 @@
+package io.quarkiverse.langchain4j.sample.apicurio;
+
+import dev.langchain4j.service.SystemMessage;
+import io.quarkiverse.langchain4j.RegisterAiService;
+import io.quarkiverse.langchain4j.mcp.runtime.apicurio.ApicurioRegistryMcpTools;
+
+@RegisterAiService(tools = ApicurioRegistryMcpTools.class)
+public interface Bot {
```
jmartisk, line 11: You need to have a `@SessionScoped` annotation since you're accessing it via a websocket. Otherwise, the memory will be scoped to a single message from the user, so it will immediately forget everything.

samples/apicurio-registry-mcp/README.md
```diff
@@ -0,0 +1,88 @@
...
+        "contentType": "application/json"
+      }
+    }
+  }'
+```
+
+### 4. Set your OpenAI API key
+
+```bash
+export QUARKUS_LANGCHAIN4J_OPENAI_API_KEY=your-api-key
```
jmartisk: In the application, you switched to Ollama instead of OpenAI, so the instructions are desynchronized now.

mcp-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/apicurio/ApicurioRegistryMcpTools.java
```diff
@@ -0,0 +1,168 @@
...
+                .build();
+
+        return new DefaultMcpClient.Builder()
+                .key(key)
+                .transport(transport)
+                .build();
+    }
+
+    private static String resolveGroupId(String groupId) {
+        if (groupId == null || groupId.isBlank()) {
```
jmartisk:
```suggestion
        if (groupId == null || groupId.isBlank() || "null".equals(groupId)) {
```

When the LLM passes "null" as a string here, it still doesn't work for me because it searches for a literal "null" groupId.

docs/modules/ROOT/pages/mcp.adoc
```diff
@@ -371,6 +371,134 @@ The registry client supports:
...
+service can search a registry for available servers and connect to them on demand.
+
+The registry stores MCP tool definitions as versioned `MCP_TOOL` artifacts, validated against the official
+https://modelcontextprotocol.io/specification/2025-11-25/server/tools[MCP specification]. Tool definitions follow the spec
+(`name`, `inputSchema`, optional `description`, `title`, `annotations`, `outputSchema`), while connection metadata
+(server URL, transport type) is stored as **artifact labels** -- keeping the artifact content spec-compliant.
+The Quarkus integration provides `@Tool`-annotated methods that allow the LLM itself to search the registry and connect to
+discovered servers -- no redeployment or config changes needed when new MCP servers become available.
+
+IMPORTANT: This feature is experimental and may change in future releases.
```
jmartisk, line 387: Let's add that the MCP clients added through this are currently 'global' - because the tool provider is application-scoped, so the tools will be visible by all users. This is something that will need fixing soon though.

samples/apicurio-registry-mcp/src/main/java/io/quarkiverse/langchain4j/sample/apicurio/Bot.java
```diff
@@ -0,0 +1,19 @@
...
+
+    @SystemMessage("""
+            You are an intelligent assistant with access to Apicurio Registry.
+            You can search for MCP servers in the registry, connect to them,
+            and then use their tools to help the user.
+
+            When a user asks for help, first search the registry for relevant
+            MCP servers, connect to them, and then use their tools.
+            """)
+    String chat(String message);
```
jmartisk, line 22: The way things are, this AI service will not pick up the tool provider, so the tools from the dynamically added MCP clients won't be visible.

Unfortunately, I see several problems.

First off, this method needs to be annotated with `@McpToolBox` to be eligible to use the generated tool provider.

Then, the current behavior of AI services is that when it contains a `@RegisterAiService(tools = ....)` with an explicit tool class, then only that class will be used and dynamic tool providers will be ignored :(  See https://github.com/quarkiverse/quarkus-langchain4j/blob/1.11.2/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/AiServicesRecorder.java#L257 - I tend to think that this is a bug... I think we should remove the `!hasExplicitTools` part of the condition there, and then, the AI service will actually start seeing that tool provider.

mcp-apicurio-registry/deployment/src/main/java/io/quarkiverse/langchain4j/mcp/deployment/apicurio/McpApicurioRegistryProcessor.java
```diff
@@ -0,0 +1,89 @@
...
+            McpApicurioRegistryRecorder recorder,
+            McpRecorder mcpRecorder) {
+        if (!buildTimeConfig.enabled()) {
+            return;
+        }
+
+        // Ensure an McpToolProvider exists so that dynamically discovered MCP clients
+        // can be added to it at runtime. This is a defaultBean, so if the MCP processor
+        // already created one (when static MCP clients are configured), ours is ignored.
+        if (mcpBuildTimeConfiguration.generateToolProvider().orElse(true)) {
```
jmartisk: No, I think we can't do this, it's not ignored. When you configure a static MCP client, then it will generate the same provider twice and you get this at startup

```
Caused by: java.lang.IllegalStateException: A synthetic bean with identifier sL_YcZTZI2rxkspvkYJxQT2XtaU is already registered: SYNTHETIC bean [types=[dev.langchain4j.service.tool.ToolProvider, java.lang.Object], qualifiers=[@Any, @Default], target=n/a]
```

Could you, instead of this, make sure the MCP processor creates the provider? I could imagine that the Apicurio processor could produce a new build item, something like `AlwaysCreateMcpToolProviderBuildItem`, and the MCP processor would, if this build item exists, create the tool provider even if there are no configured MCP servers.

jmartisk (changes requested):

jmartisk:
Thanks for your work on this. Unfortunately it's getting a bit more complicated than I thought :(

mcp/deployment/src/test/java/io/quarkiverse/langchain4j/mcp/test/MultipleMcpClientsTest.java
carlesarnal: This was a leftover from the rebase — the file is now identical to main. It will disappear when we squash the commits.

samples/apicurio-registry-mcp/src/main/java/io/quarkiverse/langchain4j/sample/apicurio/Bot.java
carlesarnal: Done — added `@SessionScoped` so chat memory persists across the WebSocket session.

samples/apicurio-registry-mcp/README.md
carlesarnal: Fixed — removed OpenAI instructions, README now reflects Ollama with `qwen2.5:7b`.

mcp-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/apicurio/ApicurioRegistryMcpTools.java
carlesarnal: Accepted — added the `"null".equals(groupId)` check back.

docs/modules/ROOT/pages/mcp.adoc
carlesarnal: Added a NOTE in the docs explaining that dynamically added MCP clients are currently global (application-scoped) and visible to all users.

samples/apicurio-registry-mcp/src/main/java/io/quarkiverse/langchain4j/sample/apicurio/Bot.java
carlesarnal: Added `@McpToolBox` to the `chat` method so it picks up the generated tool provider. Regarding the `!hasExplicitTools` condition in `AiServicesRecorder` — I agree that looks like a bug, but fixing it would affect all AI services. Should that be a separate PR?

mcp-apicurio-registry/deployment/src/main/java/io/quarkiverse/langchain4j/mcp/deployment/apicurio/McpApicurioRegistryProcessor.java
carlesarnal: Good catch — replaced the duplicate ToolProvider bean with an `AlwaysCreateMcpToolProviderBuildItem`. The Apicurio processor now produces this build item, and the MCP processor creates the tool provider (even when no static MCP clients are configured) if the build item is present. No more duplicate bean conflicts.

samples/apicurio-registry-mcp/src/main/java/io/quarkiverse/langchain4j/sample/apicurio/Bot.java
jmartisk: I think we can do it in this PR for simplicity, it's kinda a breaking change but not much, it fixes something that should have worked properly

jmartisk:
I can finally get this to work by removing the `!hasExplicitTools` condition at https://github.com/quarkiverse/quarkus-langchain4j/blob/b71ac2e4da73bb247491b3341ad9cf5c0a977b37/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/AiServicesRecorder.java#L264 - are you going to add it?

I know I'm probably a bit annoying, but I do think this should also be covered by a test - to verify that an AI service invocation will actually "see" those tools from the added MCP server. You'd probably need to use a mock chat model that returns information about what tools are available to it - similar to the `MyChatModel` class here https://github.com/quarkiverse/quarkus-langchain4j/blob/b71ac2e4da73bb247491b3341ad9cf5c0a977b37/mcp/deployment/src/test/java/io/quarkiverse/langchain4j/mcp/test/MultipleMcpClientsTest.java#L176

carlesarnal:
> I can finally get this to work by removing the `!hasExplicitTools` condition at
> 
> https://github.com/quarkiverse/quarkus-langchain4j/blob/b71ac2e4da73bb247491b3341ad9cf5c0a977b37/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/AiServicesRecorder.java#L264
> 
> - are you going to add it?
> I know I'm probably a bit annoying, but I do think this should also be covered by a test - to verify that an AI service invocation will actually "see" those tools from the added MCP server. You'd probably need to use a mock chat model that returns information about what tools are available to it - similar to the `MyChatModel` class here
> 
> https://github.com/quarkiverse/quarkus-langchain4j/blob/b71ac2e4da73bb247491b3341ad9cf5c0a977b37/mcp/deployment/src/test/java/io/quarkiverse/langchain4j/mcp/test/MultipleMcpClientsTest.java#L176

Not annoying, no worries. I was still working through this when you made this last comment. Let me push the latest changes.

jmartisk:
The documentation check is failing. That said, I don't really know how to  help fix it because for me, running a build creates a huge bunch of other changes in the docs. This whole thing is weird. @geoand are you able to correctly build the docs so that the check passes?

But, I looked at the other contents of the PR and I'm happy with it now. Thanks!

geoand:
> @geoand are you able to correctly build the docs so that the check passes?

I will give it a try later on because I'm currently out

omatheusmesmo:
> The documentation check is failing. That said, I don't really know how to help fix it because for me, running a build creates a huge bunch of other changes in the docs. This whole thing is weird. @geoand are you able to correctly build the docs so that the check passes?
> 
> But, I looked at the other contents of the PR and I'm happy with it now. Thanks!

Hi @carlesarnal , did you rebase your branch using the latest main? I think you need to run a `clean install` too after rebasing , I did it locally and it worked.

@geoand @jmartisk , do you think we should update the message of this workflow?I've been seeing some contributors having a "hard time" with it sometimes

jmartisk:
> @geoand @jmartisk , do you think we should update the message of this workflow?I've been seeing some contributors having a "hard time" with it sometimes

How do you want to update it? It feels to me that the result somehow depends on other things than just the JDK used, so I still get a different result than the CI run, and I can't pinpoint why.

We could theoretically have an automated action that automatically reformats and commits it, but  I can imagine that's a terrible idea for various reasons.

geoand:
Yeah, let's not have CI update the docs.

I think we should probably revert the change that added the CI check until we figure out exactly what the heck the docs build does and changes files each time.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/AiServicesRecorder.java
```diff
@@ -259,9 +259,7 @@ public QuarkusAiServiceContext apply(SyntheticCreationalContext<QuarkusAiService
                             // if-exists provider
                             Instance<ToolProvider> instance = creationalContext
                                     .getInjectedReference(TOOL_PROVIDER_TYPE_LITERAL);
-                            // if the service has explicit tools and a BeanIfExistsToolProviderSupplier,
-                            // just give priority to the explicit tools, don't throw an error
-                            if (instance.isResolvable() && !hasExplicitTools) {
+                            if (instance.isResolvable()) {
```
geoand, line 264: How come this is needed?

Sorry if this has been addressed before, I've lost track :)

jmartisk: We discussed it in previous comments here, this is some remnant from the times when it wasn't possible to have both static tools and a tool provider. Now, there's no reason to limit to just one of these.

geoand: Got it, thanks!

omatheusmesmo:
@jmartisk @geoand , sorry if I was not so clear, but I meant update the error message and not the feature branch. It did worked locally for me after _rebasing_ onto the latest origin main. What's the error you are getting locally?

geoand:
> but I meant update the error message and not the feature branch

What would you update to?

omatheusmesmo:
> > but I meant update the error message and not the feature branch
> 
> What would you update to?

Usually I rebase my branch using the latest main and then I follow the process described currently, so maybe something like this:

```text
Generated documentation is not up to date.
To fix this:
  1. Ensure you have the upstream remote:
     git remote add upstream https://github.com/quarkiverse/quarkus-langchain4j.git
  2. Fetch and rebase on upstream main:
     git fetch upstream && git rebase upstream/main
  3. Run: mvn -B clean install -DskipTests -Dno-format -ntp
  4. Commit the changed documentation files
```

carlesarnal:
@jmartisk @geoand anything needed from my side to move this forward? I don't think there's anything I can do documentation-wise. Thanks!

geoand:
Have you followed the steps outlined above?

In any case, let's wait for @jmartisk to be back from PTO before getting this in

jmartisk:
I'm back from holiday now.... do we have any idea how to fix the documentation build?

jmartisk:
When I run the docs build locally, I still get a hugely different result (+75000/-14000 lines) and I don't know why or what to do with it.

geoand:
I force-pushed a docs fix

carlesarnal:
CI is all green now. Is there anything else needed from my side, or is this ready to merge?

carlesarnal:
The only failing check is the Documentation check — `quarkus-all-config.adoc` needs regeneration with the new mcp-apicurio-registry config entries. I can't reproduce CI's exact output locally (different classpath produces a much larger diff). I see you do 'Docs rebuild' commits on main after merges so maybe that's the plan?

jmartisk:
@carlesarnal I think I just found the root cause of the documentation weirdness! 
You need to make sure you don't have any git worktrees nested under the directory where you run the build because the documentation generator mistakenly picks up stuff from worktrees that it should ignore.
Please try regenerating the documentation again, but remove any extra worktrees nested under your basedir

jmartisk:
Ok I've sent a PR (https://github.com/quarkusio/quarkus/pull/56105) to Quarkus to fix this as a bug in the documentation generator. For now, we need to work around it by not having worktrees under the basedir.

carlesarnal:
> @carlesarnal I think I just found the root cause of the documentation weirdness! You need to make sure you don't have any git worktrees nested under the directory where you run the build because the documentation generator mistakenly picks up stuff from worktrees that it should ignore. Please try regenerating the documentation again, but remove any extra worktrees nested under your basedir

Ah, let me try. I do have a least two of them.

carlesarnal:
@jmartisk it worked! Thanks for the workaround.

jmartisk (approved):

jmartisk:
Thanks!! And apologies that it took so long to merge...

carlesarnal:
> Thanks!! And apologies that it took so long to merge...

No worries! I'm glad it got merged!

PR #2345, opened 2026-04-17
Have `NoopChatMemory` maintain a local copy of messages

### LINKED ISSUE #2344: NoopChatMemory: messages cannot be null or empty when no chat memory provider is registered
During  investigation of camel-quarkus coexistence with quarkus-langchain4j I encountered a `null` message exception. It seems that tool execution fails with NoopChatMemory (stateless agents) due to empty messages list

```
java.lang.IllegalArgumentException: messages cannot be null or empty
	at dev.langchain4j.internal.Exceptions.illegalArgument(Exceptions.java:25)
	at dev.langchain4j.internal.ValidationUtils.ensureNotEmpty(ValidationUtils.java:70)
	at dev.langchain4j.model.chat.request.ChatRequest.<init>(ChatRequest.java:21)
```

When an AI service uses tools with NoopChatMemory (e.g., agents configured without chat memory provider), the tool execution loop fails during the second  chat request with an error "messages cannot be null or empty".

JiriOndrusek:
Tool execution failed when using NoopChatMemory (agents without persistent chat memory) because the
   second chat request in the tool execution loop received an empty messages list, causing the error:
   "messages cannot be null or empty".                                                               
                  
  NoopChatMemory is intentionally designed to discard all messages (no-op implementation), but the   
  code was relying on committableChatMemory.messages() to build subsequent chat requests during tool
  execution loops.                                                                                   
                  
  Implemented fix:                                                                                           
  
  - Add messages to both committableChatMemory (for agents with memory) AND the local list           
  - Use the local list when building chat requests instead of calling                                
  committableChatMemory.messages()                                   
  - This ensures tool execution works correctly with both stateful and stateless agents 
  - Messages list in NoopChatMemory is changed from unmodifiable list to ArrayList to avoid issue during ative compilation with camel-quarkus-langchain4j on classpath

Test `ToolExecutionWithNoopChatMemoryTest` covers the scenario.

- Fixes: https://github.com/quarkiverse/quarkus-langchain4j/issues/2344

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionWithNoopChatMemoryTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionWithNoopChatMemoryTest.java
@@ -1,0 +1,159 @@
+package io.quarkiverse.langchain4j.test.tools;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+import java.util.function.Supplier;
+
+import jakarta.enterprise.context.control.ActivateRequestContext;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import dev.langchain4j.agent.tool.Tool;
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.data.message.ToolExecutionResultMessage;
+import dev.langchain4j.memory.ChatMemory;
+import dev.langchain4j.memory.chat.ChatMemoryProvider;
+import dev.langchain4j.model.chat.ChatModel;
+import dev.langchain4j.model.chat.request.ChatRequest;
+import dev.langchain4j.model.chat.response.ChatResponse;
+import dev.langchain4j.model.output.FinishReason;
+import dev.langchain4j.model.output.TokenUsage;
+import dev.langchain4j.service.MemoryId;
+import dev.langchain4j.service.UserMessage;
+import io.quarkiverse.langchain4j.RegisterAiService;
+import io.quarkiverse.langchain4j.ToolBox;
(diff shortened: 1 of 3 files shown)


core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionWithNoopChatMemoryTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+        @ToolBox(MyTools.class)
+        String chat(@MemoryId String memoryId, @UserMessage String userMessage);
+    }
+
+    public static class NullChatMemoryProviderSupplier implements Supplier<ChatMemoryProvider> {
+
+        @Override
+        public ChatMemoryProvider get() {
+            return null;
```
geoand, line 80: This is very weird. What is the real-world use case for doing this?

JiriOndrusek: The camel- way of using the langchain4j does not register any AiServices. I created the test to mimic this situation,
I'll share a PR to camel-quarkus here during today. where the whole scenario could be see.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/NoopChatMemory.java
```diff
@@ -23,7 +23,7 @@ public void add(ChatMessage message) {
 
     @Override
     public List<ChatMessage> messages() {
-        return Collections.emptyList();
+        return new ArrayList<>();
```
geoand, line 26: This change is very odd. Why would we do want to do this?

JiriOndrusek: TBH I'm starting to feel, that I should probably register a simple ChatMemoryProvider in camel-quarkus when coexisting with quarkus-langchain4j.
I noticed that when I run camel-quarkus-langchain4j in the **native** with quarkus-langchain4j, this list is being modified by the CQ's buildstep. I originally  ignored this as non-intentional use of the class, but now I'm starting to feel, that adding messages here is not correct. 
I'll share the Camel-quarkus Pr today and then will reach you here. My knowledge about ql4j is rather basic, so I might to be fixing the stuff weirdly.

geoand: > TBH I'm starting to feel, that I should probably register a simple ChatMemoryProvider in camel-quarkus when coexisting with quarkus-langchain4j.

Right, that's perfectly valid

PR #2352, opened 2026-04-19
Add support for additional model options in Ollama extension

### LINKED ISSUE #2255: Ollama Parameters in Dev Services
Hello, is there any way to pass parameters so when the DevService starts the ollama model, it starts with the defined arguments:

Example of possible commands:

`ollama run ibm/granite3.3-guardian:8b >>> /set system violence`

`ollama run ibm/granite3.3-guardian:8b --think=true`

Thank you very much.

omatheusmesmo:
Allow users to pass additional model options (like think=true) to Ollama models via configuration.

### Configuration example:
```
quarkus.langchain4j.ollama.devservices.model-options.think=true
quarkus.langchain4j.ollama.devservices.model-options.num-ctx=4096
```
Or directly at runtime:
```
quarkus.langchain4j.ollama.chat-model.options.think=true
quarkus.langchain4j.ollama.chat-model.options.num-ctx=4096
```
---

Fixes #2255

### DIFF
--- a/model-providers/ollama/deployment/src/main/java/io/quarkiverse/langchain4j/ollama/deployment/devservices/OllamaDevServicesBuildConfig.java
+++ b/model-providers/ollama/deployment/src/main/java/io/quarkiverse/langchain4j/ollama/deployment/devservices/OllamaDevServicesBuildConfig.java
@@ -1,7 +1,9 @@
 package io.quarkiverse.langchain4j.ollama.deployment.devservices;
 
+import java.util.Map;
 import java.util.OptionalInt;
 
+import io.quarkus.runtime.annotations.ConfigDocDefault;
 import io.quarkus.runtime.annotations.ConfigGroup;
 import io.smallrye.config.WithDefault;
 
@@ -35,4 +37,12 @@ public interface OllamaDevServicesBuildConfig {
      * </p>
      */
     OptionalInt port();
+
+    /**
+     * Additional model options to apply when pulling/loading the model.
+     * These options will be propagated to the chat-model configuration.
+     * For example: {@code quarkus.langchain4j.ollama.devservices.model-options.think=true}
+     */
+    @ConfigDocDefault("{}")
+    Map<String, String> modelOptions();
 }
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/devservice/DevServicesOllamaProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/devservice/DevServicesOllamaProcessor.java
@@ -3,6 +3,7 @@ package io.quarkiverse.langchain4j.deployment.devservice;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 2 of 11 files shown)


model-providers/ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/runtime/config/EmbeddingModelConfig.java
```diff
@@ -53,4 +54,11 @@ public interface EmbeddingModelConfig {
      */
     @ConfigDocDefault("false")
     Optional<Boolean> logResponses();
+
+    /**
+     * Additional model options to pass to Ollama.
+     * These are passed as-is to the Ollama API.
+     */
+    @ConfigDocDefault("{}")
+    Map<String, String> options();
```
geoand: Given that this doesn't support arbitrary values, I really don't think it should be a map. We should have specific configuration options for each supported value

geoand (commented):
Thanks for this!

I added an important comment

model-providers/ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/runtime/config/EmbeddingModelConfig.java
omatheusmesmo: > Given that this doesn't support arbitrary values, I really don't think it should be a map. We should have specific configuration options for each supported value

Hi @geoand, I removed the options field from the embedding model after realizing that the settings I was exposing — such as think, mirostat, num_ctx, and others — are specific to chat functionality and aren't applicable to `/api/embed`. Additionally, for the chat model, I switched to typed configuration fields instead of using a Map, as you suggested.

model-providers/ollama/deployment/src/main/java/io/quarkiverse/langchain4j/ollama/deployment/devservices/OllamaDevServicesBuildConfig.java
```diff
@@ -35,4 +37,14 @@ public interface OllamaDevServicesBuildConfig {
      * </p>
      */
     OptionalInt port();
+
+    /**
+     * Additional model options to apply when pulling/loading the model.
+     * These options will be propagated to the chat-model configuration.
+     * For example: {@code quarkus.langchain4j.ollama.devservices.model-options.think=true}
+     *
+     * @see io.quarkiverse.langchain4j.ollama.runtime.config.ChatModelConfig.ModelOptionsConfig
```
geoand: I am pretty sure this is going to be confusing to users, because it seems like this is the same thing, when in fact it's not

model-providers/ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/runtime/config/EmbeddingModelConfig.java
geoand: Thanks!

model-providers/ollama/deployment/src/main/java/io/quarkiverse/langchain4j/ollama/deployment/devservices/OllamaDevServicesBuildConfig.java
omatheusmesmo: @geoand , sorry about this, I already fixed.

omatheusmesmo:
I'm checking why the build failed, maybe I just need to rebase, but I'll test locally first.

omatheusmesmo:
@geoand , is this normal given how recently it was released?

```bash
# Check BOM 1.9.0 on Maven Central
$ curl -s -o /dev/null -w "%{http_code}" "https://repo.maven.apache.org/maven2/io/quarkiverse/langchain4j/quarkus-langchain4j-bom/1.9.0/quarkus-langchain4j-bom-1.9.0.pom"
404
```

geoand:
It seems to have failed

geoand (approved):
Thanks!

omatheusmesmo:
> Thanks!

Thank you for your guidance and patience!

geoand:
🙏

PR #2355, opened 2026-04-20
docs: register tooltip inline macro in Antora build

### LINKED ISSUE #2354: docs: `tooltip:` macro leaks raw source into rendered HTML for enum config properties
## What happens

On the live docs site, the `tooltip:` macro emitted by the Quarkus config-doclet for enum-valued configuration properties is not being rendered — the raw macro source appears in the HTML instead.

Example: [rag-easy-rag.html](https://docs.quarkiverse.io/quarkus-langchain4j/dev/rag-easy-rag.html), configuration row for `quarkus.langchain4j.easy-rag.path-type`.

Rendered HTML (from the live page):

```html
<td class="tableblock halign-left valign-middle">
  <div class="content"><div class="paragraph">
    <p>tooltip:filesystem[The <code>path()</code> represents a filesystem reference], tooltip:classpath[The <code>path()</code> represents a classpath reference]</p>
  </div></div>
</td>
<td class="tableblock halign-left valign-middle">
  <p class="tableblock">tooltip:filesystem[The {@link #path()} represents a filesystem reference]</p>
</td>
```

The table cells end up with literal `tooltip:filesystem[...]` text instead of a hover/tooltip on the enum value. `{@link #path()}` inside one of the tooltip bodies also reaches the HTML verbatim.

Source: [`docs/modules/ROOT/pages/includes/quarkus-langchain4j-easy-rag.adoc`](https://github.com/quarkiverse/quarkus-langchain4j/blob/main/docs/modules/ROOT/pages/includes/quarkus-langchain4j-easy-rag.adoc#L49-L50) lines 49–50:

```asciidoc
a|tooltip:filesystem[The `path()` represents a filesystem reference], tooltip:classpath[The `path()` represents a classpath reference]
|tooltip:filesystem[The {@link #path()} represents a filesystem reference]
```

## Root cause (best guess)

`tooltip:` is a Quarkus-core asciidoctor extension that the main Quarkus docs build registers at build time, but the Antora pipeline for `quarkus-langchain4j` doesn't load it. The `EasyRagPathType` enum's per-constant javadoc gets emitted into the generated config docs in tooltip form, and then falls through.

## Downstream impact

I ran into this while syncing these pages into a downstream docs set that uses its own asciidoctor (no `tooltip:` extension either). I'd love to avoid maintaining a local post-processor if you're willing to handle this upstream.

## Would you consider inlining the tooltip text?

Rather than registering the tooltip extension (which still only works on hover and has no equivalent on print/PDF/AI summaries), would you consider having the doclet (or a small post-processor) inline the description next to the value? Something like:

```asciidoc
# Before
a|tooltip:filesystem[The `path()` represents a filesystem reference], tooltip:classpath[The `path()` represents a classpath reference]

# After
a|`filesystem` (The `path()` represents a filesystem reference), `classpath` (The `path()` represents a classpath reference)
```

This renders correctly in any asciidoctor without extensions, keeps all the information visible (no hover required), and avoids the separate `{@link}` issue in the second cell.

For reference, here's the small normalizer I'm running downstream: [scripts/normalize-config-doclet-macros.py](https://gitlab.cee.redhat.com/quarkus-documentation/quarkus/-/blob/main/scripts/normalize-config-doclet-macros.py) (two regex passes, skips listing blocks so code samples aren't touched).

Happy to open a PR with whichever approach you prefer — just let me know if you'd like the inline-text transform, the `tooltip:` extension registered in the Antora build, or something else. Thanks for maintaining this extension!

rolfedh:
Fixes #2354.

## Summary

Registers a `tooltip` inline macro processor in the Antora/asciidoctor.js build so the config-doclet's `tooltip:enumvalue[description]` emissions stop leaking to the rendered HTML.

The implementation is a direct port of Quarkus's own [`TooltipInlineMacroProcessor.java`](https://github.com/quarkusio/quarkus/blob/main/docs/src/main/java/io/quarkus/docs/generation/TooltipInlineMacroProcessor.java) — same behavior: enum value is rendered in monospace, description is discarded.

## Before / after

Source (unchanged, generated by the config-doclet):

```asciidoc
a|tooltip:filesystem[The `path()` represents a filesystem reference], tooltip:classpath[The `path()` represents a classpath reference]
|tooltip:filesystem[The {@link #path()} represents a filesystem reference]
```

Rendered HTML **before** this PR (current `docs.quarkiverse.io/quarkus-langchain4j/dev/rag-easy-rag.html`):

```html
<p>tooltip:filesystem[The <code>path()</code> represents a filesystem reference], tooltip:classpath[...]</p>
```

Rendered HTML **after** this PR (verified locally by running `npx antora antora-playbook.yml`):

```html
<p><code>filesystem</code>, <code>classpath</code></p>
```

Zero `tooltip:` source appears in the built HTML.

## Implementation notes

- New file: `docs/extensions/tooltip-inline-macro.js` — 23 lines, no dependencies, handles both modern `registry.register(...)` and legacy direct registration APIs.
- `docs/antora-playbook.yml`: added the extension path alongside `asciidoctor-kroki`.
- No changes to source `.adoc` files.

## Test plan

- [x] `npx antora antora-playbook.yml` builds without errors
- [x] Output `rag-easy-rag.html` contains `<code>filesystem</code>, <code>classpath</code>` and zero `tooltip:` strings

### DIFF
--- a/docs/antora-playbook.yml
+++ b/docs/antora-playbook.yml
@@ -14,6 +14,7 @@ ui:
 asciidoc:
   extensions:
     - asciidoctor-kroki
+    - ./extensions/tooltip-inline-macro.js
   attributes:
     kroki-fetch-diagram: false
     page-related-doc-categories: ai,java,ml
--- a/docs/extensions/tooltip-inline-macro.js
+++ b/docs/extensions/tooltip-inline-macro.js
@@ -1,0 +1,32 @@
+// @ts-check
+// Asciidoctor.js port of Quarkus's TooltipInlineMacroProcessor.java:
+// https://github.com/quarkusio/quarkus/blob/main/docs/src/main/java/io/quarkus/docs/generation/TooltipInlineMacroProcessor.java
+//
+// The Quarkus config-doclet emits `tooltip:enumvalue[description]` for
+// enum-valued configuration properties. Without a registered processor the
+// macro leaks to rendered HTML as literal text. Quarkus's own docs build
+// registers the Java processor above, which discards the description and
+// renders just the value in monospace. This module mirrors that behavior for
+// the Antora/asciidoctor.js build used by quarkus-langchain4j.
+//
+// Input : tooltip:filesystem[The `path()` represents a filesystem reference]
+// Output: `filesystem`
+
+module.exports.register = function register (registry) {
+  if (typeof registry.register === 'function') {
+    registry.register(function () {
+      this.inlineMacro('tooltip', function () {
+        this.process(function (parent, target) {
+          return this.createInline(parent, 'quoted', target, { type: 'monospaced' })
+        })
+      })
+    })
+  } else if (typeof registry.inlineMacro === 'function') {
+    registry.inlineMacro('tooltip', function () {
+      this.process(function (parent, target) {
+        return this.createInline(parent, 'quoted', target, { type: 'monospaced' })
+      })
+    })
+  }
+  return registry
+}


geoand:
Thanks @rolfedh!

I have no way of verifying this. Maybe @gsmet or @gastaldi can comment?

rolfedh:
WIP

rolfedh:
Update — CI surfaced a blocker I missed in my local testing.

My verification ran `npx antora antora-playbook.yml` directly and loaded the extension successfully. The CI path goes through `quarkus-antora`, which pipes every `asciidoc.extensions` entry through `npm i` — and per its [README](https://github.com/quarkiverse/quarkus-antora#asciidoc-extensions) and [#92](https://github.com/quarkiverse/quarkus-antora/issues/92), local extension scripts aren't currently supported. `npm i ./extensions/tooltip-inline-macro.js` fails because a bare `.js` file isn't npm-installable, and no directory-with-`package.json` restructure resolves both for npm (cwd=repo root) and Antora (playbook-relative).

Real options:

1. **Fix [quarkiverse/quarkus-antora#92](https://github.com/quarkiverse/quarkus-antora/issues/92)** first — adds local-extension support. Then this PR should work with minor tweaks.
2. **Publish the tooltip processor as an npm package** (e.g., under quarkiverse or my personal npm) so it can be referenced by name. Happy to do this if you'd like that dependency — just wanted to check before adding another package to the build.
3. **Close this PR and leave #2354 open** as a "downstream-only" issue. Consumers that can't register extensions handle it via source-level transforms.

Which do you prefer? Happy to follow your lead.

geoand:
@ppalaga ?

rolfedh:
Thinking more about the original ask on #2354 — "Could we get the tooltip extension registered instead?" — I think the most direct answer is **option 2 from my previous comment**: publish the tooltip processor as an npm package and reference it by name. That fits the existing `quarkus-antora` toolchain exactly, doesn't require fixing a second repo, and matches what registration "normally" looks like.

Before I publish anything, one naming question: would you prefer
- `@rolfedh/asciidoctor-tooltip` (my personal scope, easy to transfer later if you want to adopt it under the Quarkiverse org), or
- an unscoped name like `asciidoctor-tooltip`, or
- something under an existing Quarkiverse/Quarkus-adjacent scope you can point me at?

Happy to publish under whichever naming you prefer and update this PR to reference the package by name.

And if you'd rather not go this way at all, happy to close the PR and #2354 — we already have a downstream fix in place, so there's no pressure on your end.

geoand:
I am pretty sure that @gsmet (correctly) does not want us to use such npm packages, given how many security issues there have been with npm

rolfedh:
Understood — thanks for the signal on npm dependencies. That's a fair concern and I won't press it further. Closing this PR. Our downstream (Red Hat Pantheon / IBM WFM) already renders correctly via a source-level transform, so there's no customer-facing urgency on our side.

geoand:
Got it, thanks!

ppalaga:
I see this PR is closed, but anyway: I though nobody is using quarkus-antora with quarkus-langchain4j, because it loads for too long. It sounds like you use it @rolfedh? I am open to accepting any fixes in quarkus-antora if needed.

PR #2356, opened 2026-04-20
When using a ChatModel either directly or through an agent always include all tag keys with sentinel defaults

### LINKED ISSUE #2349: MetricsChatModelListener registers gen_ai.client.token.usage with inconsistent tag sets in agentic/manual ChatModel producer scenarios
## Summary

In a Quarkus application using `quarkus-langchain4j` with agentic interfaces (`@Agent`, `@SequenceAgent`, etc.), `MetricsChatModelListener` can register the same Micrometer metric name with different tag sets, which causes Prometheus to reject the second registration.

The metric involved is:

- `gen_ai.client.token.usage`

Prometheus error:

```text
java.lang.IllegalArgumentException: Prometheus requires that all meters with the same name have the same set of tag keys.
There is already an existing meter named 'gen_ai_client_token_usage_total' containing tag keys
[gen_ai_operation_name, gen_ai_request_model, gen_ai_response_model, gen_ai_token_type].
The meter you are attempting to register has keys
[ai_service_class_name, ai_service_method_name, gen_ai_operation_name, gen_ai_request_model, gen_ai_response_model, gen_ai_token_type].
```
## Environment

- Quarkus `3.34.x`
- quarkus-langchain4j BOM `1.7.x`
- OpenAI-compatible provider through Quarkus LangChain4j OpenAI
- Micrometer Prometheus registry enabled (`/q/metrics`)
- Agentic APIs (`@Agent`) instead of classic `@RegisterAiService`
- Manual `@Produces` `ChatModel` / `StreamingChatModel`

mariofusco:
The problem was that `MetricsChatModelListener` conditionally adds `ai_service.class_name` and `ai_service.method_name` tags to metrics based on whether a Vert.x duplicated context is active. This means that depending on this condition the same metric name could get registered with different tag key sets, while Prometheus requires all meters with the same name to have identical tag keys, and rejects the second registration with `IllegalArgumentException`.

The fix is simply to always include all tag keys with sensible defaults.

- Closes: #2349

### DIFF
--- a/agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/AgentMeterRegistryTest.java
+++ b/agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/AgentMeterRegistryTest.java
@@ -1,0 +1,172 @@
+package io.quarkiverse.langchain4j.agentic.deployment;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+import jakarta.inject.Inject;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import dev.langchain4j.agentic.Agent;
+import dev.langchain4j.agentic.declarative.ChatModelSupplier;
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.model.chat.ChatModel;
+import dev.langchain4j.model.chat.listener.ChatModelListener;
+import dev.langchain4j.model.chat.request.ChatRequest;
+import dev.langchain4j.model.chat.response.ChatResponse;
+import dev.langchain4j.model.output.TokenUsage;
+import dev.langchain4j.service.V;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import io.quarkiverse.langchain4j.RegisterAiService;
+import io.quarkiverse.langchain4j.openai.testing.internal.OpenAiBaseTest;
+import io.quarkiverse.langchain4j.testing.internal.WiremockAware;
+import io.quarkus.arc.Arc;
+import io.quarkus.arc.ManagedContext;
+import io.quarkus.test.QuarkusUnitTest;
+import io.smallrye.common.vertx.VertxContext;
+import io.vertx.core.Context;
+import io.vertx.core.Vertx;
+
+/**
+ * Reproducer for <a href="https://github.com/quarkiverse/quarkus-langchain4j/issues/2349">#2349</a>:
+ * MetricsChatModelListener registers gen_ai.client.token.usage with inconsistent tag sets
+ * when both @Agent and @RegisterAiService are used in the same application.
+ */
+public class AgentMeterRegistryTest extends OpenAiBaseTest {
+
+    @BeforeAll
+    static void addSimpleRegistry() {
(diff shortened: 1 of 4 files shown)


geoand (approved):

saulocn (approved):

geoand:
CI seems to not like the change :)

mariofusco:
> CI seems to not like the change :)

I saw :) I will give a look.

mariofusco:
@geoand it should be ok now

agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/AgentPrometheusTest.java
```diff
@@ -0,0 +1,176 @@
...
+        Metrics.globalRegistry.add(prometheus);
+    }
+
+    @AfterAll
+    static void removePrometheusRegistry() {
+        if (prometheus != null) {
+            Metrics.globalRegistry.remove(prometheus);
+            prometheus.close();
+        }
+    }
```
geoand: Do we really need this? IIRC, we've never used this and usually just use the in-memory registry for tests

mariofusco: Not sure, that's used to check assertions against that `PrometheusMeterRegistry`. Is there a better way to do this? Should I remove that test completely?

geoand: We need the test, but my point is that we don't need the Prometheus registry just for the test. See https://github.com/quarkusio/quarkus/blob/6b2c1ef050bfed40e842447bb2b0c83d379f20b2/extensions/micrometer/deployment/src/test/java/io/quarkus/micrometer/runtime/MicrometerTimedInterceptorTest.java#L43 for example

mariofusco: Done, please give it another look.

agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/AgentMeterRegistryTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Test
+    void agentAndAiServiceTokenUsageMetricsHaveConsistentTagKeys() throws Exception {
+        agent.ask("test question");
+
+        assertThat(registry.find("gen_ai.client.token.usage").counters())
+                .as("Agent call should have produced token usage metrics")
+                .isNotEmpty();
+
+        io.vertx.core.Context dupCtx = VertxContext.createNewDuplicatedContext(vertx.getOrCreateContext());
```
geoand: Minor nitpick: I don't see why the full class name should be used

agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/AgentMetricsTest.java
```diff
@@ -0,0 +1,132 @@
...
+                .collect(Collectors.toSet());
+
+        assertThat(distinctTagKeySets)
+                .as("All meters with the same name must have identical tag keys (Prometheus requirement)")
+                .hasSize(1);
+    }
+
+    private static void assertTagKeysPresentOnAll(Collection<? extends io.micrometer.core.instrument.Meter> meters,
+            String... expectedKeys) {
+        for (io.micrometer.core.instrument.Meter meter : meters) {
```
geoand: Same comment about the class name

agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/AgentPrometheusTest.java
geoand: Thanks!

agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/AgentMeterRegistryTest.java
mariofusco: Fixed.

agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/AgentMetricsTest.java
mariofusco: Fixed.

geoand (approved):
Thanks!

PR #2359, opened 2026-04-20
Rebuild all-config.adoc

omatheusmesmo:
Regenerate `quarkus-all-config.adoc` which is missing the Llama3-Java
configuration section (409 lines).

This was likely caused by commit `47f30dd0` which partially regenerated
the file when adding Anthropic response format configuration, reverting
the Llama3-Java configs that had been added by the previous "Rebuild docs"
commit (`cebb0699e`).

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-all-config.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-all-config.adoc
@@ -13898,6 +13898,415 @@ endif::add-copy-button-to-env-var[]
 
 
 
+h|[.extension-name]##Quarkus LangChain4j - Llama3 - Java##
+h|Type
+h|Default
+
+a|icon:lock[title=Fixed at build time] [[quarkus-langchain4j-llama3-java_quarkus-langchain4j-llama3-include-models-in-artifact]] [.property-path]##link:#quarkus-langchain4j-llama3-java_quarkus-langchain4j-llama3-include-models-in-artifact[`+++quarkus.langchain4j.llama3.include-models-in-artifact+++`]##
+ifdef::add-copy-button-to-config-props[]
+config_property_copy_button:+++quarkus.langchain4j.llama3.include-models-in-artifact+++[]
+endif::add-copy-button-to-config-props[]
+
+
+[.description]
+--
+Determines whether the necessary Jlama models are downloaded and included in the jar at build time. Currently, this option is only valid for `fast-jar` deployments.
+
+
+ifdef::add-copy-button-to-env-var[]
+Environment variable: env_var_with_copy_button:+++QUARKUS_LANGCHAIN4J_LLAMA3_INCLUDE_MODELS_IN_ARTIFACT+++[]
+endif::add-copy-button-to-env-var[]
+ifndef::add-copy-button-to-env-var[]
+Environment variable: `+++QUARKUS_LANGCHAIN4J_LLAMA3_INCLUDE_MODELS_IN_ARTIFACT+++`
+endif::add-copy-button-to-env-var[]
+--
+|boolean
+|`+++true+++`
+
+a|icon:lock[title=Fixed at build time] [[quarkus-langchain4j-llama3-java_quarkus-langchain4j-llama3-chat-model-enabled]] [.property-path]##link:#quarkus-langchain4j-llama3-java_quarkus-langchain4j-llama3-chat-model-enabled[`+++quarkus.langchain4j.llama3.chat-model.enabled+++`]##
+ifdef::add-copy-button-to-config-props[]
+config_property_copy_button:+++quarkus.langchain4j.llama3.chat-model.enabled+++[]
+endif::add-copy-button-to-config-props[]
+
+
+[.description]
+--
+Whether the model should be enabled
+
+
+ifdef::add-copy-button-to-env-var[]
+Environment variable: env_var_with_copy_button:+++QUARKUS_LANGCHAIN4J_LLAMA3_CHAT_MODEL_ENABLED+++[]
+endif::add-copy-button-to-env-var[]
+ifndef::add-copy-button-to-env-var[]
+Environment variable: `+++QUARKUS_LANGCHAIN4J_LLAMA3_CHAT_MODEL_ENABLED+++`
+endif::add-copy-button-to-env-var[]
+--
+|boolean
+|`+++true+++`
+
+a|icon:lock[title=Fixed at build time] [[quarkus-langchain4j-llama3-java_quarkus-langchain4j-llama3-chat-model-model-name]] [.property-path]##link:#quarkus-langchain4j-llama3-java_quarkus-langchain4j-llama3-chat-model-model-name[`+++quarkus.langchain4j.llama3.chat-model.model-name+++`]##
+ifdef::add-copy-button-to-config-props[]
+config_property_copy_button:+++quarkus.langchain4j.llama3.chat-model.model-name+++[]
+endif::add-copy-button-to-config-props[]
+
+
+[.description]
+--
+Model name to use
+
+
+ifdef::add-copy-button-to-env-var[]
+Environment variable: env_var_with_copy_button:+++QUARKUS_LANGCHAIN4J_LLAMA3_CHAT_MODEL_MODEL_NAME+++[]
+endif::add-copy-button-to-env-var[]
+ifndef::add-copy-button-to-env-var[]
+Environment variable: `+++QUARKUS_LANGCHAIN4J_LLAMA3_CHAT_MODEL_MODEL_NAME+++`
+endif::add-copy-button-to-env-var[]
+--
+|string
+|`+++mukel/Llama-3.2-1B-Instruct-GGUF+++`
+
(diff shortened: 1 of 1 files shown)


mcruzdev:
Good catch!

mcruzdev:
I would like to suggest an approach similar to the one used in the Quarkus Flow extension to ensure documentation is kept up to date. You can see an example here:

[It adds a validation step in the CI pipeline to detect when documentation changes are missing, helping prevent inconsistencies between code and docs.](https://github.com/quarkiverse/quarkus-flow/blob/f0ff1306d9fb85636145437a21b38b7e2bbdfee4/.github/workflows/build-website.yaml#L41)

cc: @gastaldi it is a very common issue in some repositories (do you know a better way to achieve it?)

omatheusmesmo:
> I would like to suggest an approach similar to the one used in the Quarkus Flow extension to ensure documentation is kept up to date. You can see an example here:
> 
> [It adds a validation step in the CI pipeline to detect when documentation changes are missing, helping prevent inconsistencies between code and docs.](https://github.com/quarkiverse/quarkus-flow/blob/f0ff1306d9fb85636145437a21b38b7e2bbdfee4/.github/workflows/build-website.yaml#L41)

Totally agree, this could be a good first issue.

gastaldi:
> I would like to suggest an approach similar to the one used in the Quarkus Flow extension to ensure documentation is kept up to date. You can see an example here:
> 
> [It adds a validation step in the CI pipeline to detect when documentation changes are missing, helping prevent inconsistencies between code and docs.](https://github.com/quarkiverse/quarkus-flow/blob/f0ff1306d9fb85636145437a21b38b7e2bbdfee4/.github/workflows/build-website.yaml#L41)
> 
> cc: @gastaldi it is a very common issue in some repositories (do you know a better way to achieve it?)

That's an interesting check. Maybe something we could do it in the pre-release check in https://github.com/quarkiverse/.github/blob/main/.github/workflows/pre-release.yml? 

Or maybe not because that workflow doesn't build the project

geoand:
> I would like to suggest an approach similar to the one used in the Quarkus Flow extension to ensure documentation is kept up to date. You can see an example here:
> 
> [It adds a validation step in the CI pipeline to detect when documentation changes are missing, helping prevent inconsistencies between code and docs.](https://github.com/quarkiverse/quarkus-flow/blob/f0ff1306d9fb85636145437a21b38b7e2bbdfee4/.github/workflows/build-website.yaml#L41)
> 
> cc: @gastaldi it is a very common issue in some repositories (do you know a better way to achieve it?)

I like it!

Would you like to add it here?

geoand (approved):
Thanks!

jmartisk:
Just to point out to make it clear for everyone - the likely cause why this disappeared from the docs is that it was built with JDK <22. Only with 22+ the llama3-java module activates, so when building the full docs, it's necessary to use JDK 22+.

omatheusmesmo:
> Just to point out to make it clear for everyone - the likely cause why this disappeared from the docs is that it was built with JDK <22. Only with 22+ the llama3-java module activates, so when building the full docs, it's necessary to use JDK 22+.

Hi @jmartisk , I built  using JDK 17 because of the sdkman file. So should I change the JDK version to 22 and run `mvn clean install -pl docs -am` ? If so, it's still outdated at moment.

mcruzdev:
> > I would like to suggest an approach similar to the one used in the Quarkus Flow extension to ensure documentation is kept up to date. You can see an example here:
> 
> > 
> 
> > [It adds a validation step in the CI pipeline to detect when documentation changes are missing, helping prevent inconsistencies between code and docs.](https://github.com/quarkiverse/quarkus-flow/blob/f0ff1306d9fb85636145437a21b38b7e2bbdfee4/.github/workflows/build-website.yaml#L41)
> 
> > 
> 
> > cc: @gastaldi it is a very common issue in some repositories (do you know a better way to achieve it?)
> 
> 
> 
> I like it!
> 
> 
> 
> Would you like to add it here?

@omatheusmesmo Do would like to add it? Please, go ahead!

gsmet:
> Maybe something we could do it in the pre-release check in https://github.com/quarkiverse/.github/blob/main/.github/workflows/pre-release.yml?

The config doc should be updated automatically on release. At least that's why the original templates was supposed to do.

omatheusmesmo:
> > > I would like to suggest an approach similar to the one used in the Quarkus Flow extension to ensure documentation is kept up to date. You can see an example here:
> > 
> > > 
> > 
> > > [It adds a validation step in the CI pipeline to detect when documentation changes are missing, helping prevent inconsistencies between code and docs.](https://github.com/quarkiverse/quarkus-flow/blob/f0ff1306d9fb85636145437a21b38b7e2bbdfee4/.github/workflows/build-website.yaml#L41)
> > 
> > > 
> > 
> > > cc: @gastaldi it is a very common issue in some repositories (do you know a better way to achieve it?)
> > 
> > 
> > 
> > I like it!
> > 
> > 
> > 
> > Would you like to add it here?
> 
> @omatheusmesmo Do would like to add it? Please, go ahead!

I really would like to, but I have my hands full now. Please go ahead.

jmartisk:
> > Just to point out to make it clear for everyone - the likely cause why this disappeared from the docs is that it was built with JDK <22. Only with 22+ the llama3-java module activates, so when building the full docs, it's necessary to use JDK 22+.
> 
> Hi @jmartisk , I built using JDK 17 because of the sdkman file. So should I change the JDK version to 22 and run `mvn clean install -pl docs -am` ? If so, it's still outdated at moment.

When I rebuild the docs in the current `main` branch with JDK 25, I get +93000/-4700 line changes :/ the documentation build is really odd, I find it hard to tell what state is correct TBH...

omatheusmesmo:
> > > Just to point out to make it clear for everyone - the likely cause why this disappeared from the docs is that it was built with JDK <22. Only with 22+ the llama3-java module activates, so when building the full docs, it's necessary to use JDK 22+.
> > 
> > 
> > Hi @jmartisk , I built using JDK 17 because of the sdkman file. So should I change the JDK version to 22 and run `mvn clean install -pl docs -am` ? If so, it's still outdated at moment.
> 
> When I rebuild the docs in the current `main` branch with JDK 25, I get +93000/-4700 line changes :/ the documentation build is really odd, I find it hard to tell what state is correct TBH...

Yeah, I faced this problem too, see #2304 , but I think it's something related to JDK 25, but with JDK 22 just change the all-config last time I checked. 

```bash
$ git -C /mnt/fileshare/projects/java/OpenSource/quarkus-langchain4j diff --stat
.../ROOT/pages/includes/quarkus-all-config.adoc    | 11872 +++++++++++++++++++
 1 file changed, 11872 insertions(+)
 ```

gsmet:
@jmartisk if you have extensions that are only enabled for a given version of the JDK, that might explain the issue.

There shouldn't be any difference in what's generated by the plugin when using different JDK versions. If there are, it's a bug.

omatheusmesmo:
> @jmartisk if you have extensions that are only enabled for a given version of the JDK, that might explain the issue.
> 
> There shouldn't be any difference in what's generated by the plugin when using different JDK versions. If there are, it's a bug.

Hi @gsmet , @jmartisk , so is this expected since there is modules using a different JDK version or is it a bug? If is a bug, where should I open an issue about?

jmartisk:
I guess some of the differences may also be coming from the fact that when you change branches using git, and the new branch doesn't contain a module that existed in the previous branch, the `target` directory of that module stays in your worktree and the documentation builder seems to pick it up even though it's not a valid maven module (there is no pom.xml). This is usually fixed by running `git clean -fd` before the build.
I think we need to specify that the build needs to be done with JDK 25 (to make sure it includes all modules) and add a workflow that validates it. That workflow will obviously need to use 25.

jmartisk:
Let  me try and send a PR with this.. Let's see how it goes.

gsmet:
> he target directory of that module stays in your worktree and the documentation builder seems to pick it up even though it's not a valid maven module (there is no pom.xml).

With recent Quarkus versions, it shouldn't, I fixed it a while ago (https://github.com/quarkusio/quarkus/blob/b1f30319e679097e28076841cf63539add486396/devtools/config-doc-maven-plugin/src/main/java/io/quarkus/maven/config/doc/GenerateConfigDocMojo.java#L279-L283 , included in 3.25).

If you still have this behavior, it's a bug.

jmartisk:
> With recent Quarkus versions, it shouldn't, I fixed it a while ago (https://github.com/quarkusio/quarkus/blob/b1f30319e679097e28076841cf63539add486396/devtools/config-doc-maven-plugin/src/main/java/io/quarkus/maven/config/doc/GenerateConfigDocMojo.java#L279-L283 , included in 3.25).
> 
> If you still have this behavior, it's a bug.

Hmm. Checking for existence of `pom.xml` resolves the problem with switching git branches, but it doesn't resolve the problem where the module might be deactivated due to Maven profiles, so it can still pick up stale data from an older build when the module was activated.

jmartisk:
Anyway we need to always build with all modules active, so we need a check that validates that all modules were built,, so that shouldn't probably be a big problem when we set up CI properly.

But, I welcome any tips how to make the docs check pass in https://github.com/quarkiverse/quarkus-langchain4j/pull/2371
It builds with JDK 25, I built it locally with JDK 25, did `git clean -fd` before it, and the check still fails. I'm puzzled.

PR #2364, opened 2026-04-22
Fix max-retries=0 causing unexpected behavior in OpenAI and Anthropic providers

### LINKED ISSUE #482: Setting max-retries=0 causes infinite retries
If you set, say, `quarkus.langchain4j.openai.max-retries=0`, it will attempt to retry indefinitely.
The correct setting for "no retries" is 1, but given the property name, some people might think that `max-retries=0` means "just try once", which isn't the case. Instead, it causes the retry loop to immediately break through the upper bound, and because the end condition is based on `==` it will probably only stop after an integer overflow...

suryateja-g13:
## Summary

- Adds startup-time validation in `OpenAiRecorder` that throws `ConfigValidationException` when `max-retries < 1` for chat, embedding, and moderation models
- Adds the same validation in `AnthropicRecorder` for the chat model
- Other providers (`QuarkusOpenAiImageModel`, `AzureOpenAiChatModel`, `AzureOpenAiEmbeddingModel`, `QuarkusCohereScoringModel`) already validate `< 1` at construction time — this change makes OpenAI vanilla and Anthropic consistent
- Adds a `QuarkusUnitTest` that verifies the error message is surfaced at startup when `quarkus.langchain4j.openai.max-retries=0`

Fixes: #482

## Test plan

- [ ] `OpenAiInvalidMaxRetriesTest` verifies that setting `quarkus.langchain4j.openai.max-retries=0` fails at startup with a descriptive error message
- [ ] Existing OpenAI and Anthropic tests continue to pass (default `max-retries=1`)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
+++ b/model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
@@ -64,6 +64,10 @@ public class AnthropicRecorder {
                 throw new ConfigValidationException(createApiKeyConfigProblem(configName));
             }
 
+            if (chatModelConfig.maxRetries() < 1) {
+                throw new ConfigValidationException(createMaxRetriesConfigProblem(configName));
+            }
+
             var builder = AnthropicChatModel.builder()
                     .baseUrl(anthropicConfig.baseUrl())
                     .apiKey(apiKey)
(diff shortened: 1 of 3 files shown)


geoand (approved):
Thanks!

geoand:
The CI failures on the model-providers jobs are related

suryateja-g13:
The remaining CI failures in **JVM tests - mcp - Java 17/21** are pre-existing and unrelated to this PR. They are caused by network timeouts when `McpRegistryClientTest` tries to reach `registry.modelcontextprotocol.io` — an external live service that is unavailable in the CI environment. Java 25 passes because it hits a different runner where the connection succeeds. All model-providers, anthropic, and other relevant tests pass cleanly.

jmartisk (approved):

jmartisk:
The MCP registry test was probably a temporary outage of the remote registry, it's passing now. Thanks

PR #2370, opened 2026-04-23
fix #2369: Make topP and temperature optional in ChatModelConfig for `openai` and `azure/openai`

### LINKED ISSUE #2369: Make `topP` or `temperature` optional in `ChatModelConfig` to avoid conflicts with providers that reject setting both
I'm using the OpenAI extension with Azure OpenAI through Portkey (an OpenAI-compatible gateway). When I send a chat request, it fails with:

```
temperature and top_p cannot both be specified for this model. Please use only one.
```

I can't work around this because `ChatModelConfig.topP()` returns `Double` with `@WithDefault("1.0")`, so it's always sent in the request. I tried setting it to empty (`top-p=`) and to `NaN` or `-1`, but none works.

Changing the return type to `Optional<Double>` or `OptionalDouble` (while keeping `@WithDefault("1.0")`) would let me suppress it:

```
quarkus.langchain4j.openai.chat-model.top-p=
```

The same applies to `temperature()`.

If it would help, I could create a PR.

t1:
- Fixes: #2369


geoand (approved):
Thanks!

geoand:
Actually we'll need the commits squashed for this

t1:
I've just squashed them

PR #2371, opened 2026-04-23
Rebuild the docs, add a check

jmartisk:
Per the discussion on https://github.com/quarkiverse/quarkus-langchain4j/pull/2359

I built it using JDK 25 and did `git clean -fd` beforehand, so hopefully this state of the docs is correct...?!


jmartisk:
Ok this is weird, I purposely built it with JDK 21 to miss some files, but the newly added check is complaining about much more. I need to figure out where the difference is coming from..

omatheusmesmo:
> Following.

geoand:
@jmartisk what should we do with this?

jmartisk:
Uhm, let me try this again when I get a bit of time, I think it would be very nice to have, but I need to figure out why the CI is failing when it passes on my laptop.

jmartisk:
replaced by https://github.com/quarkiverse/quarkus-langchain4j/pull/2628

PR #2383, opened 2026-04-24
docs(few-shots): inline-code INPUT/OUTPUT pattern; use [source,bash] for curl

rolfedh:
## Summary

- Wrap the ``INPUT: … OUTPUT: …`` pattern as inline code in `guide-few-shots.adoc` instead of quoting it as prose, so the intent (a literal template for few-shot entries) is clear.
- Switch the curl example block from the unrecognized ``[source,curl]`` to ``[source,bash]``, matching the convention used elsewhere in `docs/modules/ROOT/pages` for shell commands.

No content or behavior changes beyond these two formatting tweaks.


geoand (approved):
Thanks!

rolfedh:
@geoand Could we backport this to the 1.7 branch? Quarkus LangChain4j 1.7 is what we sync to RHBQ 3.33, which is the current downstream release. Thanks!

geoand:
Label added :)

PR #2378, opened 2026-04-24
fix: SemanticSimilarityStrategy - calculateCosineSimilarity returns double greater 1.0

KaiSuchomel:
I got the following Exception:
```
2026-04-24 09:21:30,746 ERROR [io.quarkiverse.langchain4j.testing.evaluation.Scorer] (pool-12-thread-2) Failed to evaluate sample `german_to_english_question`: java.lang.IllegalArgumentException: Score must be between 0.0 and 1.0, got: 1.0000000026365623
	at io.quarkiverse.langchain4j.testing.evaluation.EvaluationResult.<init>(EvaluationResult.java:30)
EvaluationResult.java:30
	at io.quarkiverse.langchain4j.testing.evaluation.EvaluationResult.passed(EvaluationResult.java:43)
EvaluationResult.java:43
	at io.quarkiverse.langchain4j.testing.evaluation.similarity.SemanticSimilarityStrategy.evaluate(SemanticSimilarityStrategy.java:94)
SemanticSimilarityStrategy.java:94
	at io.quarkiverse.langchain4j.testing.evaluation.similarity.SemanticSimilarityStrategy.evaluate(SemanticSimilarityStrategy.java:14)
SemanticSimilarityStrategy.java:14
	at io.quarkiverse.langchain4j.testing.evaluation.Scorer.lambda$evaluate$0(Scorer.java:47)
Scorer.java:47
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
Executors.java:545
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
FutureTask.java:328
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
ThreadPoolExecutor.java:1090
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
ThreadPoolExecutor.java:614
	at java.base/java.lang.Thread.run(Thread.java:1474)

```
Changed calculateCosineSimilarity in SemanticSimilarityStrategy

PR #2384, opened 2026-04-24
docs(messages-and-memory): add explicit kebab-case IDs to all section headings

rolfedh:
> **🔁 Backport request: `triage/backport-1.7`.** Maintainers — please apply the `triage/backport-1.7` label so this is picked up by the backport workflow. (I tried to self-label and got `does not have the correct permissions to execute AddLabelsToLabelable`.)
>
> Both issues this PR fixes — the dangling `<<configure-default-memory,...>>` xref and the legacy `[#_memory_backends_with_chatmemorystore]` ID — exist verbatim on upstream `1.7`. Section structure is essentially identical (the only difference is the `== Chat Memory Flush Strategy` section added by PR #2378, which doesn't conflict). Red Hat Build of Quarkus 3.33 ships LangChain4j 1.7.4 and syncs docs from upstream `1.7`; the downstream Pantheon build's `CheckCrossReferences` step only resolves explicit IDs, so backporting these gives RHBQ a reliable cross-reference path.

## Summary

Replace renderer-dependent auto-generated anchors with explicit `[id="..."]` attributes on every section heading in `docs/modules/ROOT/pages/messages-and-memory.adoc`, so the anchors are stable across Asciidoctor, DITA, and any other downstream renderer.

Also fixes two anchor issues surfaced while doing this:

- The existing `[#_memory_backends_with_chatmemorystore]` ID duplicated the auto-generated slug. Replaced with `[id="memory-backends-with-chatmemorystore"]` and updated the one internal `<<..>>` reference.
- The `<<configure-default-memory, ...>>` xref on (former) line 102 was dangling — no matching ID existed anywhere in the file. Pointed it at the newly added `[id="configuring-the-default-memory"]` on the section it describes.

The recently added "Chat Memory Flush Strategy" section (PR #2378) is included and given the same explicit-ID treatment (`[id="chat-memory-flush-strategy"]`) instead of the legacy `[#_chat_memory_flush_strategy]` form, for consistency with the rest of the file after this change.

No content changes — IDs only.

### DIFF
--- a/docs/modules/ROOT/pages/messages-and-memory.adoc
+++ b/docs/modules/ROOT/pages/messages-and-memory.adoc
@@ -9,6 +9,7 @@ All contextual understanding must come from the current input, which is why **me
 
 This page explains how Quarkus LangChain4j handles _messages_, how it builds and manages memory, and how you can configure, extend, and control these behaviors.
 
+[id="statelessness-and-context-rehydration"]
 == Statelessness and Context Rehydration
 
 Large language models do not retain memory between calls.
@@ -22,6 +23,7 @@ Replaying a large memory consumes CPU and bandwidth.
 
 image:memory.png[width=600,align="center",alt="Context Rehydration"]
 
+[id="messages-the-building-blocks-of-conversation"]
 == Messages: The Building Blocks of Conversation
 
 A #message# is a structured unit of interaction in a conversation.
@@ -55,6 +57,7 @@ Here's a simple multi-turn exchange/conversation:
 
 When the application sends the last message, it includes the entire history of messages to provide context for the model.
 
+[id="memory-and-its-purpose"]
 == Memory and Its Purpose
 
 Memory refers to the accumulation of previous messages in an exchange.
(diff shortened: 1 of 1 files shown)


geoand:
Do we really need all those IDs?

jmartisk:
I think the point is to get predictable IDs across upstream and downstream renderers, which you don't get when you don't specify the ID explicitly. But if we're doing this, I think we should either do it across the whole documentation, or not at all - not in just one specific file.

cescoffier:
I agree with @jmartisk - either nowhere or everywhere.

rolfedh:
The single failed job (`JVM tests - tools - Java 17`) failed with `java.net.ConnectException: Connection timed out` on a CI runner — looks like a transient network issue, unrelated to the docs change in this PR. Could a maintainer re-run that job? Thanks!

rolfedh:
Thanks @geoand @jmartisk @cescoffier — the consensus here was "either everywhere or nowhere," and you're right. The "everywhere" path was going to be tooling to roll this out across the docs, but I've withdrawn that and I'm stepping back from the explicit-heading-ID initiative across these repos. That makes this single-file change exactly the "somewhere" you agreed against, so I'm closing it rather than merging a partial.

One thing worth salvaging, independent of the IDs question: this PR also fixed a genuine bug — a dangling `<<configure-default-memory,...>>` xref (plus a legacy `[#_memory_backends_with_chatmemorystore]` ID) in `messages-and-memory.adoc`. That fix stays on the branch, so anyone is welcome to cherry-pick it as a standalone one-liner if it is useful.

Appreciate the review.

PR #2388, opened 2026-04-24
docs(streamed-responses): use OpenAI in the Project Setup example

rolfedh:
The "Using Streamed Responses" guide is provider-agnostic — streaming works with any chat model that supports it. The Project Setup example currently shows Ollama and the comment says "Or any other model provider that supports streaming, such as OpenAI."

Flip the primary example to OpenAI for three reasons:

- OpenAI is the canonical example across most other quarkus-langchain4j guides (openai-chat-model, quickstart-rag, guide-generating-image), so readers arriving here from elsewhere in the docs see a consistent setup pattern.
- OpenAI-only downstream products (such as the Red Hat Build of Quarkus and IBM EBoQ product) currently ship a broken example here, since quarkus-langchain4j-ollama isn't in their supported extensions set. Using OpenAI lets the guide render as a working example in every distribution that packages quarkus-langchain4j-openai.
- The "Or any other model provider..." comment is retained with Ollama as the flipped example, so community users running Ollama locally still have a visible pointer to the alternative.

Content-only change; no behavior or API change.

### DIFF
--- a/docs/modules/ROOT/pages/guide-streamed-responses.adoc
+++ b/docs/modules/ROOT/pages/guide-streamed-responses.adoc
@@ -32,10 +32,10 @@ Add the following dependencies in your `pom.xml`:
 
 [source,xml,subs=attributes+]
 ----
-<!-- Or any other model provider that supports streaming, such as OpenAI -->
+<!-- Or any other model provider that supports streaming, such as Ollama -->
 <dependency>
     <groupId>io.quarkiverse.langchain4j</groupId>
-    <artifactId>quarkus-langchain4j-ollama</artifactId>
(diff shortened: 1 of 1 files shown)


geoand (approved):
Makes sense, thanks!

rolfedh:
@geoand Could we backport this to the 1.7 branch? Quarkus LangChain4j 1.7 is what we sync to RHBQ 3.33, which is the current downstream release. Since this PR was motivated by downstream distributions that don't ship quarkus-langchain4j-ollama, having it on 1.7 would let RHBQ 3.33 pick up the OpenAI example on the next sync rather than carrying a downstream-only patch. Thanks!

geoand:
Sure thing, I added the label

PR #2391, opened 2026-04-25
fix #2369: make temperature() and topP() OptionalDouble

### LINKED ISSUE #2369: Make `topP` or `temperature` optional in `ChatModelConfig` to avoid conflicts with providers that reject setting both

anuragg-saxenaa:
## Summary

Fixes issue #2369: `ChatModelConfig.topP()` always returns `Double` with `@WithDefault("1.0")`, so `top_p` is always sent in API requests. Developers using OpenAI-compatible gateways hit "temperature and top_p cannot both be specified for this model" errors.

## Changes

- **ChatModelConfig.java**: Changed `temperature()` and `topP()` return types from `Double` (with `@WithDefault`) to `OptionalDouble` (with `@ConfigDocDefault` for docs)
- **OpenAiRecorder.java**: Updated both `chatModel()` and `streamingChatModel()` to use `.orElse(null)` so null is passed when not configured
- **ChatModelConfigSerializationTest.java**: New regression test that verifies:
  - When neither temperature nor topP is set → neither appears in JSON
  - When topP is set → top_p appears in JSON
  - When only topP is set → only top_p appears (not temperature)
  - When only temperature is set → only temperature appears (not top_p)

## Why This Works

The langchain4j `OpenAiChatModel.Builder.temperature(Double)` and `.topP(Double)` accept `null` — when `null`, the field is omitted from the JSON request body. By switching to `OptionalDouble`, we can pass `null` to the builder when the config is absent.

## Backwards Compatibility

- **Config users who had explicit values**: Their configs still work
- **Config users without topP**: Previously always sent top_p=1.0, now omitted entirely
- **Breaking**: Any user relying on the implicit top_p=1.0 default will need to set it explicitly if they want that behavior

Closes #2369


geoand:
Thanks, but we already have https://github.com/quarkiverse/quarkus-langchain4j/pull/2370 doing the same

PR #2403, opened 2026-04-28
Pass invocation context to tool provider request

Wal8800:
Pass invocation context so we can access invocation parameter in the tool provider.

This is useful for dynamic tool access based on user or tenant information

For example

```java
 @Singleton
  public class MyToolProvider implements ToolProvider {

      @Override
      public ToolProviderResult provideTools(ToolProviderRequest request) {
          InvocationParameters params = request.invocationParameters();
          String role = params.get("role");

          var builder = ToolProviderResult.builder();

          // Always provide read-only tools
          builder.add(readToolSpec, readToolExecutor);

          // Only provide write tools for admin users
          if ("admin".equals(role)) {
              builder.add(writeToolSpec, writeToolExecutor);
              builder.add(deleteToolSpec, deleteToolExecutor);
          }

          return builder.build();
      }
  }
```


geoand:
This totally makes sense, but I would like to see an updated test that uses the information you mention

geoand (approved):
Thanks!

PR #2412, opened 2026-05-04
Allow to define the model used by an agent through the @ModelName annotation

mariofusco:
At the moment it is not possible to define the model used by an agent directly with the `@ModelName` annotation and you have to do something like the following

```
public interface MyAgent {

    @UserMessage("""
            Analyze the following request: {{request}}.
            """)
    @Agent("An agent with conflicting model configuration")
    String analyze(@V("request") String request);

   @ChatModelSupplier
    static ChatModel chatModel() {
        return RemediationModel.model;
    }

    @ApplicationScoped
    class RemediationModel {
        private static ChatModel model;

        @Inject
        void init(@ModelName("remediation") ChatModel remediationModel) {
            model = remediationModel;
        }
    }
}
```

This pull request overcomes this limitation simply allowing to do something like

```
public interface MyAgent {

    @UserMessage("""
            Analyze the following request: {{request}}.
            """)
    @Agent("An agent with conflicting model configuration")
    @ModelName("remediation") 
    String analyze(@V("request") String request);
}
```

### DIFF
--- a/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
+++ b/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
@@ -93,10 +93,19 @@ public class AgenticProcessor {
                             AgenticLangChain4jDotNames.CHAT_MODEL_SUPPLIER))
                     .findFirst();
 
+            String modelName = extractModelName(methods);
+
+            if (chatModelSupplier.isPresent() && modelName != null) {
+                throw new IllegalConfigurationException(
+                        "Agent interface '" + classInfo.name()
+                                + "' cannot use both @ChatModelSupplier and @ModelName. "
+                                + "Use one or the other to specify the ChatModel.");
+            }
+
             List<MethodInfo> mcpToolBoxMethods = methods.stream()
                     .filter(mi -> mi.hasAnnotation(LangChain4jDotNames.MCP_TOOLBOX)).toList();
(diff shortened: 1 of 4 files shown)


geoand:
I assume `@ModeName` is handled by upstream LangChain4j?

mariofusco:
> I assume `@ModeName` is handled by upstream LangChain4j?

Nope this only works with the quarkus extension. the `@ModeName` annotation exists only there and in vanilla langchain4j models don't have a name.

geoand (approved):

geoand:
Got it thanks. 

In that case, we should probably also update the docs to mention this capability

mariofusco:
> In that case, we should probably also update the docs to mention this capability

I'm noticing that in the docs we don't have yet anything specifically related with the agentic module. Is it ok to mention this in the `ai-services.adoc`, where the @ModelName annotation is introduced for now, or do we want a new `agents.adoc` file? Having a new doc file only for this could be simply too much?

geoand:
As we don't have an agent-related doc, let's just get this in for now, but we really should have docs on how to use the agentic module

mariofusco:
> As we don't have an agent-related doc, let's just get this in for now, but we really should have docs on how to use the agentic module

Ok, but please wait to merge this, I'm improving something and I also want to add one more test. I will undraft the PR when ready.

geoand:
Sounds good to me

mariofusco:
This should be ok to be merged now.

geoand (approved):

geoand:
CI is failing with the change

mariofusco:
This should be ok now.

PR #2418, opened 2026-05-06
Dispatch and parallelise @RunOnVirtualThread tool batches in the streaming handler

unsignedint:
Tool calls are currently executed sequentially, which has performance impacts for complex agents. This PR provides a path to parallel execution, but iff tools run on virtual threads. That keeps the scope limited but still provides significant value; we run on JDK25 with VT.

Disclosure: AI generated with human-in-the-loop.

## Summary

Two-layer change to how the streaming AI-service runs tool batches:

1. **Dispatch.** When every tool in a batch is `@RunOnVirtualThread`, the whole tool loop is dispatched onto a virtual thread via `VirtualThreadsRecorder`, releasing the calling worker / event-loop thread immediately. Mixed batches keep the current worker-thread behaviour. `QuarkusToolExecutor` short-circuits its `submit().get()` branch when already on a virtual thread (no more `.await().indefinitely()` parking a worker).
2. **Parallelism.** When the batch has more than one tool and they're all virtual-thread, fan each call out onto its own virtual thread instead of running them serially on the carrier. Single-tool batches stay on the carrier. The carrier waits for every task, then appends `ToolExecutionResultMessage`s in original request order so the LLM sees a stable conversation.

## New config (`quarkus.langchain4j.tools.*`)

| Key | Default | Purpose |
|---|---|---|
| `dispatch` | `auto` | `auto` \| `virtual-thread` \| `worker` global override |
| `virtual-thread.max-concurrent` | unbounded | Caps simultaneously-running VT tools (per-tool under parallel, per-batch under opt-out) |
| `mixed-batch-log-level` | `warn` | Severity when a mixed batch falls back to worker dispatch |
| `parallel-virtual-thread-batch` | `true` | Set `false` to restore serialized intra-batch execution |

## Behaviour notes

- **Memory ordering preserved.** Results are appended in request order regardless of completion order.
- **Events.** `BeforeToolExecution` fires in request order on the carrier; `ToolExecutedEvent` fires from the per-tool worker on completion (completion-order observability).
- **`max-tool-calls-per-response`.** Parallel rejects the whole batch up-front when `size > limit`; serial keeps the partial-execution-then-throw contract (fan-out would race with submission).
- **Cancellation.** Best-effort: tools already running complete; not-yet-scheduled tasks record a cancellation result.
- **Error handling.** Each parallel task wraps work in a catch-all that synthesizes a `"Tool execution failed: <msg>"` result into its own slot before re-raising, so request-count == result-count even when a tool throws past its error handler.
- **Interrupt safety.** The carrier drains all futures uninterruptibly, capturing the interrupt and re-setting the flag only after every task has settled.
- **Observability.** `invocationContext` is set on parallel-batch `BeforeToolExecution` / `ToolExecution` builders (mandatory after the recent upstream observability change).
- **Docs.** New "Virtual-thread dispatch for streaming tool batches" subsection in `function-calling.adoc`, plus a clarified note on the Execution Models page covering `Uni`/`CompletionStage` build-time validation and worker-parking semantics.

## Tests included in the PR (all passing locally):

- `ToolExecutionModelVirtualThreadDispatchTest` — auto / global override / bounded dispatch / mixed fallback / VT-from-VT / failure propagation
- `ToolExecutionModelVirtualThreadLegacyDispatchTest` — pre-dispatch behaviour preserved under `dispatch=worker`
- `ToolExecutionModelVirtualThreadMaxConcurrentTest` — per-tool semaphore caps peak in-flight under parallel mode
- `ToolExecutionModelVirtualThreadParallelDispatchTest` — fan-out, ordering, events, cancellation
- `ToolExecutionModelVirtualThreadParallelOptOutTest` — `parallel-virtual-thread-batch=false` falls back to serial
- `ToolExecutionModelVirtualThreadParallelMaxToolCallsTest` — over-limit rejected up-front, at-limit allowed
- `ToolExecutionModelWithStreamingCancellationTest` — cancellation semantics intact

### Manual testing (not in this PR)

Two end-to-end ITs against the real Anthropic API exist locally and were used to validate wall-clock behaviour, but are intentionally excluded:

- `AnthropicParallelToolDispatchITTest` — three 10s `httpbin.org/delay/10` tool calls in a single turn finish in ~10s under parallel VT dispatch.
- `AnthropicSequentialToolDispatchITTest` — same shape without `@RunOnVirtualThread`; ~30s under the legacy serial path.

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionModelTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionModelTest.java
@@ -236,9 +236,9 @@ public class ToolExecutionModelTest {
             }
         }).get();
 
-        // At the moment, we create a virtual thread every time.
-        assertThat(r).contains(uuid, "quarkus-virtual-thread-")
-                .doesNotContain(caller.get());
+        // When the caller is already running on a virtual thread, the tool executes on that
+        // same virtual thread instead of being re-dispatched onto a new one.
+        assertThat(r).contains(uuid, "quarkus-virtual-thread-", caller.get());
     }
 
     @Test
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionModelVirtualThreadDispatchTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionModelVirtualThreadDispatchTest.java
@@ -1,0 +1,264 @@
+package io.quarkiverse.langchain4j.test.tools;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
+import jakarta.enterprise.context.control.ActivateRequestContext;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledForJreRange;
+import org.junit.jupiter.api.condition.JRE;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import org.testcontainers.shaded.org.awaitility.Awaitility;
+
+import dev.langchain4j.agent.tool.Tool;
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.data.message.ToolExecutionResultMessage;
+import dev.langchain4j.memory.ChatMemory;
+import dev.langchain4j.memory.chat.ChatMemoryProvider;
+import dev.langchain4j.memory.chat.MessageWindowChatMemory;
+import dev.langchain4j.model.chat.StreamingChatModel;
+import dev.langchain4j.model.chat.request.ChatRequest;
+import dev.langchain4j.model.chat.response.ChatResponse;
+import dev.langchain4j.model.chat.response.StreamingChatResponseHandler;
+import dev.langchain4j.model.output.FinishReason;
+import dev.langchain4j.model.output.TokenUsage;
+import dev.langchain4j.service.MemoryId;
+import dev.langchain4j.service.UserMessage;
+import io.quarkiverse.langchain4j.RegisterAiService;
+import io.quarkiverse.langchain4j.ToolBox;
+import io.quarkus.arc.Arc;
+import io.quarkus.test.QuarkusUnitTest;
+import io.smallrye.common.annotation.RunOnVirtualThread;
+import io.smallrye.mutiny.Multi;
+import io.vertx.core.Vertx;
+
+public class ToolExecutionModelVirtualThreadDispatchTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest unitTest = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class).addClasses(MyAiService.class));
+
+    @Inject
+    MyAiService aiService;
+
+    @Inject
+    Vertx vertx;
+
+    @Test
+    @EnabledForJreRange(min = JRE.JAVA_21)
+    @ActivateRequestContext
+    void virtualThreadBatchRunsOnVirtualThread() {
+        String uuid = UUID.randomUUID().toString();
+        String r = aiService.singleVirtualTool("mem-" + uuid, "hiVirtualThread - " + uuid)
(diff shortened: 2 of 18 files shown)


geoand:
Thanks for this!

> This PR provides a path to parallel execution, but iff tools run on virtual threads

I am not terribly excited about this. WDYT @cescoffier @jmartisk ?

jmartisk:
I'm kinda surprised we're doing this in the Quarkus extension, shouldn't this live in upstream langchain4j and only have a small integration layer in Quarkus? @dliubarskyi what is your opinion on this?

unsignedint:
> > This PR provides a path to parallel execution, but iff tools run on virtual threads
> 
> I am not terribly excited about this. WDYT @cescoffier @jmartisk ?

Which part, the only virtual threads? If so, I was thinking about comprehensive support for any executor but wanted to start small.

geoand:
> Which part, the only virtual threads?

Yeah.

>  If so, I was thinking about comprehensive support for any executor but wanted to start small.

I understand that, but it makes the use of the feature weird for users

unsignedint:
> > Which part, the only virtual threads?
> 
> Yeah.
> 
> > If so, I was thinking about comprehensive support for any executor but wanted to start small.
> 
> I understand that, but it makes the use of the feature weird for users

Understood 👍  Just to be clear, the current behaviour results in zero change for existing apps/users. Parallel tool execution must be opted in, and requires all tools on VT. I think it would be reasonably easy to broaden this to pure vert.x tool batches, but unsure about mixed execution mode.

geoand:
Right, my objection is about requiring virtual threads only

dliubarskyi:
> I'm kinda surprised we're doing this in the Quarkus extension, shouldn't this live in upstream langchain4j and only have a small integration layer in Quarkus? @dliubarskyi what is your opinion on this?

Parallel tool execution is supported in upstream: https://docs.langchain4j.dev/tutorials/tools#executing-tools-concurrently
One can also customize the `Executor` to be used. I agree that duplicated implementation makes little sense

unsignedint:
> > I'm kinda surprised we're doing this in the Quarkus extension, shouldn't this live in upstream langchain4j and only have a small integration layer in Quarkus? @dliubarskyi what is your opinion on this?
> 
> Parallel tool execution is supported in upstream: https://docs.langchain4j.dev/tutorials/tools#executing-tools-concurrently One can also customize the `Executor` to be used. I agree that duplicated implementation makes little sense

Ugh! Sorry, I've completely missed this and was only focused on this repo. That would greatly simplify the implementation and support both VT and vert.x etc! I will rework this.

unsignedint:
@jmartisk @dliubarskyi Thanks for your feedback. For context: I'm primarily motivated by parallel tool batches (for image gen, web search, DB lookups etc) running on a Quarkus virtual thread pool.

As you know, Quarkus currently reimplements the tool-calling loop in its own runtime to handle some specific behaviors. I agree we’d be better off shifting that work to langchain4j and using `executeToolsConcurrently(Executor)` for parallel dispatch. But, based on my initial analysis, I see three small gaps in langchain4j that I think we need. (Maybe there's more I'll discover) If you agree, I'll implement and send a PR there, and then re-work the Quarkus side:

1. `AiServices.maxToolCallsPerResponse(int)`

Caps the number of tool calls per LLM response. Not the same as maxSequentialToolsInvocations, which counts across iterations. Quarkus has had this for a while. I think this would be the cleanest home for it; we can also implement a ChatModel-decorator workaround.

2. Rethrow option for `ToolExecutionErrorHandler`

Some exceptions need to bubble up to the caller instead of getting turned into a tool result and fed back to the LLM. Quarkus uses a marker interface (`PreventsErrorHandlerExecution`) for this, mainly for guardrail and invariant checks. Right now `ToolService.executeWithErrorHandling` catches everything. For compatibility we need a way for the handler to identify what to rethrow. I don’t see a clean way to do this on the Quarkus side without subclassing ToolService internals.

3. Factory hook for `ToolProviderRequest` (maybe?)

The MCP provider relies on a subtype of ToolProviderRequest to filter tools by MCP client name. `createContext` and `refreshDynamicProviders` build the base type so we can’t use a subtype. A factory hook would solve it.

Thoughts?

dliubarskyi:
> 1. AiServices.maxToolCallsPerResponse(int)
> Caps the number of tool calls per LLM response. Not the same as maxSequentialToolsInvocations, which counts across iterations. Quarkus has had this for a while. I think this would be the cleanest home for it; we can also implement a ChatModel-decorator workaround.

OK, makes sense

> 2. Rethrow option for ToolExecutionErrorHandler
> Some exceptions need to bubble up to the caller instead of getting turned into a tool result and fed back to the LLM. Quarkus uses a marker interface (PreventsErrorHandlerExecution) for this, mainly for guardrail and invariant checks. Right now ToolService.executeWithErrorHandling catches everything. For compatibility we need a way for the handler to identify what to rethrow. I don’t see a clean way to do this on the Quarkus side without subclassing ToolService internals.

Not sure I understand the problem. `ToolExecutionErrorHandler` can either return tool result or throw an exception, so it is just a metter of configuring custom `ToolExecutionErrorHandler` on Quarkus side, no?

> 3. Factory hook for ToolProviderRequest (maybe?)
> The MCP provider relies on a subtype of ToolProviderRequest to filter tools by MCP client name. createContext and refreshDynamicProviders build the base type so we can’t use a subtype. A factory hook would solve it.

OK, makes sense

PR #2419, opened 2026-05-06
Drop deprecated `LastRunMetrics` from `GPULlama3BaseModel`

orionpapadakis:
This PR drops the dependency on GPULlama3's `LastRunMetrics` class which is now deprecated.


geoand (approved):
Thanks!

PR #2433, opened 2026-05-08
Add support for named Chroma embedding stores

### LINKED ISSUE #2432: Add support for multiple named Chroma embedding store instances
Parent: #2415

## Description

Add support for configuring multiple named Chroma embedding store instances, each with independent configuration and injectable via `@EmbeddingStoreName`.

Chroma uses a URL-based REST client (no CDI-managed client), so named stores typically share the same Chroma server URL but isolate via distinct `collection-name` values. Named stores without an explicit URL fall back to the default store's URL, which is the common case when all collections live on the same server.

The `collection-name` property is required at build time for named store discovery (it ensures SmallRye's `@WithDefaults` populates the `namedConfig()` map). Chroma also manages its own DevServices, so the DevServices processor must propagate the container URL to each named store's config key.

## Configuration example

```properties
quarkus.langchain4j.chroma.products.collection-name=product_embeddings
quarkus.langchain4j.chroma.products.url=http://chroma.example.com:8000
```

## Injection

```java
@Inject
@EmbeddingStoreName("products")
EmbeddingStore<TextSegment> productsStore;
```

## Implementation pattern

Follow the same approach as pgvector (#2409):

- Refactor config to composition (defaultConfig/namedConfig with `@WithParentName`)
- Add `collection-name` as a build-time property to `ChromaNamedStoreBuildTimeConfig` (required for SmallRye `@WithDefaults` to create named config map entries)
- Change `url` from required `String` to `Optional<String>` — named stores without an explicit URL fall back to the default store's URL
- Produce `SyntheticBeanBuildItem` with `@EmbeddingStoreName` qualifier
- Update the recorder with `correspondingStoreConfig()` and URL fallback logic
- Propagate DevServices URL to named store config keys in `ChromaDevServicesProcessor`

omatheusmesmo:
Enables configuring multiple Chroma embedding stores, each with independent configuration, following the same named-instance pattern used by model providers (OpenAI, Ollama, Anthropic, etc.) and other embedding store providers (pgvector, Redis, Neo4j).

Chroma has no CDI-managed client, so named stores without an explicit `url` fall back to the default store's `url`. Chroma also manages its own DevServices, so the processor propagates the container URL to each named store automatically.

## Configuration

Default store (backward compatible, no changes required):

```properties
quarkus.langchain4j.chroma.collection-name=default
```

Named stores:

```properties
quarkus.langchain4j.chroma.products.collection-name=product_embeddings
quarkus.langchain4j.chroma.products.url=http://chroma.example.com:8000
```

Disable default store when only named stores are needed:

```properties
quarkus.langchain4j.chroma.default-store-enabled=false
```

## Injection

```java
@Inject
EmbeddingStore<TextSegment> defaultStore;

@Inject
@EmbeddingStoreName("products")
EmbeddingStore<TextSegment> productsStore;
```

- Closes: #2432

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-langchain4j-chroma.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-langchain4j-chroma.adoc
@@ -7,6 +7,27 @@ h|[.header-title]##Configuration property##
 h|Type
 h|Default
 
+a|icon:lock[title=Fixed at build time] [[quarkus-langchain4j-chroma_quarkus-langchain4j-chroma-default-store-enabled]] [.property-path]##link:#quarkus-langchain4j-chroma_quarkus-langchain4j-chroma-default-store-enabled[`+++quarkus.langchain4j.chroma.default-store-enabled+++`]##
+ifdef::add-copy-button-to-config-props[]
+config_property_copy_button:+++quarkus.langchain4j.chroma.default-store-enabled+++[]
+endif::add-copy-button-to-config-props[]
+
+
+[.description]
+--
+Whether the default (unnamed) Chroma embedding store should be enabled. Set to `false` when you only want to use named stores.
+
+
+ifdef::add-copy-button-to-env-var[]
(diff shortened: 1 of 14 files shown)


geoand (approved):
Thanks!

PR #2434, opened 2026-05-08
Add named embedding stores support for Milvus

omatheusmesmo:
## Describe your changes

Adds support for multiple named Milvus embedding stores, following the same pattern established by the Redis, Neo4j, and pgvector named store implementations.

Named stores allow applications to manage embeddings for different domains or tenants separately, each pointing to a different Milvus instance or collection. The `host` property in the named store build-time config serves as both a trigger for named store creation and a discriminator for which Milvus instance to connect to.

Key changes:
- `MilvusBuildConfig` now supports `default-store-enabled` and named store configurations via `namedConfig` map
- `MilvusNamedStoreBuildTimeConfig` provides the `host` property (defaults to `<default>` to share the default Milvus instance)
- `MilvusStoreRuntimeConfig` extracted as a shared config group for both default and named stores, with `host`/`port` defaults to avoid SmallRye Config `@WithDefaults` collision
- `MilvusRecorder` updated with `embeddingStoreFunction` and `correspondingStoreConfig` for per-store bean creation
- `MilvusProcessor.types()` now includes `MilvusEmbeddingStore.class` for direct `@Inject` usage
- DevServices config kept as inner interface of `MilvusBuildConfig` (no separate config class)
- Documentation updated with "Named Stores" section in `rag-milvus-store.adoc`

## Check List

- [x] My code follows the code style of this project
- [x] My change requires changes to the documentation
- [x] I have updated the documentation accordingly
- [x] All new and existing tests passed

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-langchain4j-milvus.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-langchain4j-milvus.adoc
@@ -7,6 +7,27 @@ h|[.header-title]##Configuration property##
 h|Type
 h|Default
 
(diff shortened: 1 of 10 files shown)


geoand (approved):
Thanks!

PR #2452, opened 2026-05-13
Allow persistent ChatMemoryStore to survive AI service bean destruction

### LINKED ISSUE #1761: ChatMemoryStore cleared on pod shutdown, and how to avoid for persistent stores?
Hi team,

I noticed that in `QuarkusAiServiceContext.close()` the framework always calls `chatMemoryService.clearAll();`

[QuarkusAiServiceContext](https://github.com/quarkiverse/quarkus-langchain4j/blob/b76fba7c4bc2a34f21f013baef2f660c45f9078c/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceContext.java#L35)
[DefaultComittableChatMemory](https://github.com/quarkiverse/quarkus-langchain4j/blob/b76fba7c4bc2a34f21f013baef2f660c45f9078c/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/DefaultCommittableChatMemory.java#L65
)

This means that when the application (or pod) shuts down, all chat memory is wiped.

That makes sense for InMemoryChatMemoryStore, but if I’m using a persistent store (like Couchbase/Redis/DB), I don’t want to lose all conversations every time a pod restarts.

My question is:

- Is there a reason the framework clears memory unconditionally?
- Is there a recommended way to prevent clearAll() from wiping persistent stores?
- Would it make sense to make this configurable (e.g. quarkus.langchain4j.chat-memory.clear-on-exit=false)?
- Right now, the only workaround seems to be overriding clearAll() in my custom store with a no-op, but that messages are getting duplicated

omatheusmesmo:
## Describe your changes

`QuarkusAiServiceContext.close()` currently calls `chatMemoryService.clearAll()` unconditionally, which propagates `deleteMessages` to the underlying `ChatMemoryStore`. Because `@RegisterAiService` defaults to `@RequestScoped`, this happens at the end of every HTTP request wiping a persistent backend (Redis, Couchbase, JDBC, etc.) between requests, not just on application shutdown.

Introduce `quarkus.langchain4j.chat-memory.clear-on-close` (default: `true`, preserves historical behavior). When set to `false`, `close()` still releases the in-memory bookkeeping in `chatMemoryService` to avoid leaking memory ids for long-lived AI services, but skips the `deleteMessages` propagation to the store.

---

- Closes: #1761

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ChatMemoryClearOnCloseDisabledTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ChatMemoryClearOnCloseDisabledTest.java
@@ -1,0 +1,138 @@
+package io.quarkiverse.langchain4j.test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
+
+import jakarta.enterprise.context.control.ActivateRequestContext;
+import jakarta.inject.Inject;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.memory.chat.ChatMemoryProvider;
+import dev.langchain4j.memory.chat.MessageWindowChatMemory;
+import dev.langchain4j.model.chat.ChatModel;
+import dev.langchain4j.model.chat.request.ChatRequest;
+import dev.langchain4j.model.chat.response.ChatResponse;
+import dev.langchain4j.service.MemoryId;
+import dev.langchain4j.service.UserMessage;
+import dev.langchain4j.service.memory.ChatMemoryAccess;
+import dev.langchain4j.store.memory.chat.ChatMemoryStore;
+import io.quarkiverse.langchain4j.RegisterAiService;
+import io.quarkus.test.QuarkusUnitTest;
+
+/**
+ * Verifies that when {@code quarkus.langchain4j.chat-memory.clear-on-close=false},
+ * the {@link ChatMemoryStore} is not wiped when the AI service bean is closed
+ * (e.g., on application shutdown), so conversations backed by a persistent store
+ * survive application restarts.
+ */
+public class ChatMemoryClearOnCloseDisabledTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest unitTest = new QuarkusUnitTest()
+            .overrideConfigKey("quarkus.langchain4j.chat-memory.clear-on-close", "false")
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(MyAiService.class, CountingChatMemoryStore.class,
+                            ChatMemoryProviderSupplier.class, MirrorChatModelSupplier.class));
+
+    @Inject
+    MyAiService service;
+
+    @Test
+    @ActivateRequestContext
+    void closeShouldNotDeleteMessagesWhenClearOnCloseIsFalse() throws Exception {
+        service.chat("user-1", "hello");
+
+        int deletesBefore = CountingChatMemoryStore.DELETE_COUNT.get();
+
+        ((AutoCloseable) service).close();
+
+        int deletesAfter = CountingChatMemoryStore.DELETE_COUNT.get();
+        assertThat(deletesAfter - deletesBefore)
+                .as("ChatMemoryStore.deleteMessages must not be invoked on close "
+                        + "when clear-on-close=false")
+                .isZero();
+    }
+
+    @Test
+    @ActivateRequestContext
+    void closeShouldStillReleaseInMemoryReferencesWhenClearOnCloseIsFalse() throws Exception {
+        service.chat("leak-test-1", "hi");
+        service.chat("leak-test-2", "hi");
+
+        ChatMemoryAccess access = (ChatMemoryAccess) service;
+        assertThat(access.getChatMemory("leak-test-1")).isNotNull();
+
+        ((AutoCloseable) service).close();
+
+        assertThat(access.getChatMemory("leak-test-1"))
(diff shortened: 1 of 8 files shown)


geoand:
Thanks for this!

@suhas5 would you want to try this and see if it fits what you have in mind?

patriot1burke:
Won't new requests with the same memory id just pull in old messages from the store when a new AiService bean is created and executed on?  There's a reason that memory is tied to the CDI scope. Its so that old, unwanted messages aren't sent with chat requests.  A CDI scope delinates when conversations stop.

The *REAL* problem is that LC4J does not have dual concepts of history and memory.  I think you're better off creating a listener and a history store.

omatheusmesmo:
Thanks @patriot1burke. I wrote a focused test (`ChatMemoryReplayValidationTest`) and confirmed the mechanic you described: with `clear-on-close=false`, the second invocation on the same `memoryId` does reload the store and replay messages to the model.

Two thoughts though:

**On the replay itself**: when `memoryId` is a meaningful per-conversation identifier (e.g. `userId-sessionId`), that reload *is* the desired behavior. It's how a multi-turn conversation continues across HTTP requests. The current default (wipe the store on every `@RequestScoped` close) actually makes persistent backends like Redis/JDBC unusable for multi-turn chat today, since every request starts fresh. So this PR isn't introducing a new trade-off; it's making an already broken use case work. The replay only becomes a footgun with a shared or static `memoryId`, which is an anti-pattern regardless.

**On your "real problem" point**: fully agree LC4J conflating working memory and long-term history is the root cause, and a proper listener + history-store SPI upstream would be cleaner. Happy to open an issue at langchain4j/langchain4j to push for that. But the gap is real today, and the flag here is small, opt-in (default preserves current behavior, ~10 lines, trivially removable once upstream lands a proper split).

@geoand, you mentioned on #1761 that this should be configurable. Does this approach match what you had in mind? Happy to adjust scope or naming, or to hold the PR while we push for the upstream fix first.

patriot1burke:
IMO, your change makes the CDI scope completely irrelevant and you might as well use vanilla LC4J.   

If you use websockets instead of HTTP, you can have a stateful conversation and there's no need to reload chat memory from the store with each requestion.  As it is, HTTP based chats don't support streaming.  Also, The few chat applications I've wrote I've found that I need conversational state *beyond* chat memory that needs to live in between requests.

I'd really like to know why you can't use Websockets.

Finally, I've been doing some work around a new conversational CDI scope called [ChatScopes](https://docs.quarkiverse.io/quarkus-langchain4j/dev/chat-scopes.html).  Removes the need for juggling `@MemoryIds` and gives some structure for managing conversations with the client.  It requires websockets though at the moment.

The original POC of this extensions had an SPI for storing and restoring conversational state (chat memory and app state) and thus supported HTTP, but I didn't bring that into the final implementation as it seemed too much like a bad redo of stateful session replication.  Websockets provided a much better backbone.

omatheusmesmo:
> IMO, your change makes the CDI scope completely irrelevant and you might as well use vanilla LC4J.
> 
> If you use websockets instead of HTTP, you can have a stateful conversation and there's no need to reload chat memory from the store with each requestion. As it is, HTTP based chats don't support streaming. Also, The few chat applications I've wrote I've found that I need conversational state _beyond_ chat memory that needs to live in between requests.
> 
> I'd really like to know why you can't use Websockets.
> 
> Finally, I've been doing some work around a new conversational CDI scope called [ChatScopes](https://docs.quarkiverse.io/quarkus-langchain4j/dev/chat-scopes.html). Removes the need for juggling `@MemoryIds` and gives some structure for managing conversations with the client. It requires websockets though at the moment.
> 
> The original POC of this extensions had an SPI for storing and restoring conversational state (chat memory and app state) and thus supported HTTP, but I didn't bring that into the final implementation as it seemed too much like a bad redo of stateful session replication. Websockets provided a much better backbone.

Thanks for your comment, I'll study and reply later.

omatheusmesmo:
Thanks for the detail @patriot1burke, and sorry for the slow reply.

Stepping back first: I'm not trying to push HTTP over WebSockets as a design. I'm just trying to serve the setup in #1761, where the reporter is on HTTP with a persistent store and loses conversations on pod shutdown.

On "why can't you use WebSockets": for that scenario (multiple replicas, rolling deploys, pod restarts) a connection-pinned conversation is exactly the thing you can't rely on, which is why the state lives in a shared persistent store in the first place. And your point about needing conversational state *beyond* chat memory is fair; this change doesn't try to solve that, and I agree that part belongs to the application.

On "makes the CDI scope irrelevant": I'd gently push back that it moves the boundary rather than removing it. `ChatMemoryRemover.remove(...)` already runs `clear()` -> `deleteMessages()`, so an app can still explicitly mark a conversation as over; it just becomes an explicit call instead of the implicit `@RequestScoped` destroy. Digging into it, the per-request wiping is really a scope effect: `close()` is the `@PreDestroy` path, so with an `@ApplicationScoped` service it doesn't fire between requests and the conversation already continues across requests with today's default. That narrows what this flag uniquely adds to surviving shutdown/restart, which is exactly the title of #1761.

I also went through the ChatScopes code, since it seemed like the natural alternative. It's a great fit for live, in-process conversations, but the `activeScopes` registry is in-memory with no public way to re-begin a scope by id, so it doesn't survive a restart, and `ChatScope.end()` routes through the same `close()` -> `deleteMessages()` path. So it doesn't seem to cover the persistent-store-across-restart case on its own.

I still think your listener + history store is the right long-term fix, and scoped to chat messages only it would sidestep the session-replication concern that made you drop the POC SPI. I'm happy to open that upstream and link it here.

In the meantime I don't want to over-drive this. I'm glad to narrow the PR to "survive shutdown/restart" with the trade-offs documented (the explicit `ChatMemoryRemover` cleanup responsibility, and widening the CDI scope to avoid per-request reloads), or to hold it while we pursue the upstream split. Whatever you all feel is the right call.

patriot1burke:
> I still think your listener + history store is the right long-term fix, and scoped to chat messages only it would sidestep the session-replication concern that made you drop the POC SPI. I'm happy to open that upstream and link it here.
> 
> In the meantime I don't want to over-drive this. I'm glad to narrow the PR to "survive shutdown/restart" with the trade-offs documented (the explicit `ChatMemoryRemover` cleanup responsibility, and widening the CDI scope to avoid per-request reloads), or to hold it while we pursue the upstream split. Whatever you all feel is the right call.

I really wanted Quarkus LC4J to be able to managed memory ids and conversational boundaries for you.  For Quarkus to be a real value-add programming model above LC4J.  Quarkus was just not designed for stateful services.   But, I don't see a fast way around the shutdown problem.

PR #2456, opened 2026-05-15
Allow using `ChatRequestParameters` as an AiService parameter

### LINKED ISSUE #2454: [Feature] Support passing `ChatRequestParameters` to `AiServices` methods
# Support passing `ChatRequestParameters` to `AiServices` methods

## Motivation

Upstream `dev.langchain4j:langchain4j` has introduced the ability to pass
`ChatRequestParameters` (and provider-specific variants such as
`OpenAiChatRequestParameters`) directly into `AiServices` method calls via
[#4226](https://github.com/langchain4j/langchain4j/pull/4226).

This feature allows **per-invocation override** of LLM parameters, including:

- `temperature`
- `topP`
- `modelName`
- `toolChoice`
- `responseFormat`
- provider-specific options

Currently, `quarkus-langchain4j` does not support this capability.

## Investigation & Observation

While reviewing the current implementation, it appears that **`QuarkusAiServices` creates the AI Service proxy correctly, but it does not actually read or propagate the `ChatRequestParameters` argument during method execution**.

As a result:
- The parameter is accepted in the method signature
- It is effectively ignored at runtime
- No override is applied to the underlying `ChatModel`

This PR aims to close that gap and align runtime behavior with upstream LangChain4j.

## Goal

Enable the following programming model in Quarkus:

```java
interface AssistantWithChatParams {
    String chat(@UserMessage String userMessage, ChatRequestParameters params);
}
```

geoand:
- Closes: #2454

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceStreamingResponseHandler.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceStreamingResponseHandler.java
@@ -2,6 +2,7 @@ package io.quarkiverse.langchain4j.runtime.aiservice;
 
 import static dev.langchain4j.internal.Utils.copyIfNotNull;
 import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+import static io.quarkiverse.langchain4j.runtime.aiservice.ChatRequestParametersUtil.effectiveChatRequestParameters;
 import static java.util.Objects.nonNull;
 
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/QuarkusAiServiceStreamingResponseHandler.java
```diff
@@ -532,9 +532,19 @@ public void run() {
...
 
+                    ChatRequestParameters defaultParams = parametersBuilder.build();
+                    var userParams = AiServiceMethodImplementationSupport
+                            .findChatRequestParameters(methodCreateInfo, methodArgs);
+                    ChatRequestParameters effectiveParams;
+                    if (userParams.isPresent()) {
+                        effectiveParams = userParams.get().defaultedBy(defaultParams);
+                    } else {
+                        effectiveParams = defaultParams;
+                    }
```
mariofusco: Nitpick: this logic seems duplicated, maybe it could encapsulated in a single static method?

geoand: Sure, let me take a look

mariofusco (approved):

PR #2460, opened 2026-05-15
Introduce @OnThinking annotation for reasoning output handlers

### LINKED ISSUE #1685: Introduce configuration option for setting the `thinking` option in Ollama
### Discussed in https://github.com/quarkiverse/quarkus-langchain4j/discussions/1684

omatheusmesmo:
## Describe your changes

Reasoning models return a "thinking" field alongside the final answer, but Quarkus had no way to surface that content to the application. This PR adds a provider-agnostic consumption mechanism for non-streaming AI service responses.

A new `@OnThinking` annotation marks a `static` method on the AI service interface as the handler for the model's reasoning output. The handler is invoked whenever the response carries a non-blank `AiMessage.thinking()`. The return type and signatures of the regular AI service methods do not change.

```java
@RegisterAiService
interface MathAssistant {
    String solve(String problem);

    @OnThinking
    static void onThinking(ThinkingEmitted event) {
        Log.infof("[%s] %s", event.methodName(), event.text());
    }
}
```

The annotation does not enable thinking on the model itself - that remains each provider's responsibility (e.g. `model-options.think=true` on Ollama), because langchain4j does not unify the request-side knob across providers.

### Chat Scopes integration

For `@ChatScoped` AI services that want to forward reasoning to the connected client through the existing `ChatRouteContext.response().thinking(...)` channel, the `chat-scopes` module ships a one-line helper:

```java
@RegisterAiService
@ChatScoped
interface WeaponBuilder {
    @ChatRoute("weaponBuilder")
    String build(@UserMessage String msg);

    @OnThinking
    static void onThinking(ThinkingEmitted event) {
        ChatRouteThinking.forward(event);
    }
}
```

Forwarding is opt-in by design: thinking content can be large or sensitive, so the framework never sends it to clients implicitly.

---

- Closes: #1685

### DIFF
--- a/chat-scopes/core/runtime/src/main/java/io/quarkiverse/langchain4j/chatscopes/ChatRouteThinking.java
+++ b/chat-scopes/core/runtime/src/main/java/io/quarkiverse/langchain4j/chatscopes/ChatRouteThinking.java
@@ -1,0 +1,76 @@
+package io.quarkiverse.langchain4j.chatscopes;
+
+import jakarta.enterprise.inject.Instance;
+import jakarta.enterprise.inject.spi.CDI;
+
+import io.quarkiverse.langchain4j.runtime.aiservice.ThinkingEmitted;
+
+/**
+ * Helper for forwarding model thinking output to the active chat route's
+ * response channel from within an {@link io.quarkiverse.langchain4j.OnThinking
+ * &#64;OnThinking}-annotated handler on a {@link ChatScoped &#64;ChatScoped} AI
+ * service.
+ *
+ * <pre>
+ * {
+ *     &#64;code
+ *     &#64;RegisterAiService
+ *     &#64;ChatScoped
+ *     interface WeaponBuilder {
+ *         &#64;ChatRoute("weaponBuilder")
+ *         String build(@UserMessage String msg);
+ *
+ *         @OnThinking
+ *         static void onThinking(ThinkingEmitted event) {
+ *             ChatRouteThinking.forward(event);
+ *         }
+ *     }
+ * }
+ * </pre>
+ *
+ * When called outside an active chat scope (or when no chat route context is
+ * resolvable in CDI), the call is a no-op so the same handler can be reused
+ * across services that may or may not run inside a chat route.
+ */
+public final class ChatRouteThinking {
+
+    private ChatRouteThinking() {
+    }
+
+    /**
+     * Forwards the thinking text carried by the event to the active chat
+     * route response channel as a {@link ChatRouteConstants#THINKING} event.
+     *
+     * @param event the thinking event emitted by the AI service runtime
+     */
+    public static void forward(ThinkingEmitted event) {
+        if (event == null) {
+            return;
+        }
+        forward(event.text());
+    }
+
+    /**
+     * Forwards an arbitrary message (typically a transformed or redacted
+     * version of {@code event.text()}) to the active chat route response
+     * channel.
+     *
+     * @param text the message to send; ignored when blank
+     */
+    public static void forward(String text) {
+        if (text == null || text.isBlank()) {
+            return;
+        }
+        if (!ChatScope.isActive()) {
+            return;
+        }
+        Instance<ChatRouteContext> instance = CDI.current().select(ChatRouteContext.class);
+        if (!instance.isResolvable()) {
+            return;
+        }
+        ChatRouteContext.ResponseChannel response = instance.get().response();
+        if (response != null) {
+            response.thinking(text);
+        }
+    }
+}
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
@@ -15,6 +15,7 @@ import java.lang.reflect.Type;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -594,6 +595,8 @@ public class AiServiceMethodImplementationSupport {
 
         response = ChatResponse.builder().aiMessage(response.aiMessage()).metadata(response.metadata()).build();
 
+        emitThinkingIfPresent(response, context, invocationContext);
+
         if (TypeUtil.isResult(returnType)) {
             var parsedResponse = SERVICE_OUTPUT_PARSER.parse(
                     ChatResponse.builder().aiMessage(response.aiMessage()).build(),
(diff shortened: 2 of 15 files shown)


geoand:
Thanks a lot for this, but TBH, the use of CDI events for this is weird.

IMHO, a more intuitive API would look something like:

```java
@RegisterAiService
interface MathAssistant {

    String solve(String problem);
 
    @Thinking
    static void onThinking(ThinkingEmitted event) {
        ...
    }
}
```

patriot1burke:
Would love to have this Chat scopes aware too.

omatheusmesmo:
> Thanks a lot for this, but TBH, the use of CDI events for this is weird.
> 
> IMHO, a more intuitive API would look something like:
> 
> ```java
> @RegisterAiService
> interface MathAssistant {
> 
>     String solve(String problem);
>  
>     @Thinking
>     static void onThinking(ThinkingEmitted event) {
>         ...
>     }
> }
> ```

I'll work on it, thanks for pointing out.

omatheusmesmo:
> Would love to have this Chat scopes aware too.

I'll take a look into it too, thanks.

omatheusmesmo:
Hello @geoand @patriot1burke , hope you are doing well.

I updated my PR, could you kindly take another look? I hope now is adequate to what you both expected.

I appreciate your review, thanks.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/ThinkingEmitted.java
```diff
@@ -0,0 +1,22 @@
...
+ * {@link io.quarkiverse.langchain4j.Thinking @Thinking}-annotated static
+ * handler declared on the AI service interface.
+ *
+ * @param text the thinking text returned by the model
+ * @param methodName the AI service method whose call produced the thinking
+ * @param serviceClass the AI service interface that declared the method
+ * @param memoryId the chat memory id associated with the invocation, or {@code null}
+ * @param emittedAt timestamp captured when the record was built
+ */
+public record ThinkingEmitted(String text,
```
geoand: I would prefer if this were an interface

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -635,6 +640,57 @@ private static Object doImplement0(AiServiceMethodCreateInfo methodCreateInfo, I
...
+            try {
+                m.setAccessible(true);
+                return Optional.of(MethodHandles.lookup().unreflect(m));
+            } catch (IllegalAccessException e) {
+                throw new IllegalStateException(
+                        "Unable to resolve @Thinking handler '" + serviceClass.getName() + "#" + m.getName() + "'", e);
+            }
+        }
+        return Optional.empty();
+    }
```
geoand: We definitely don't want to be resolving the method that needs to be run at runtime. We need to know at build time whether there is a `@Thinking` method at build time

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -140,6 +147,8 @@ public class AiServiceMethodImplementationSupport {
                                                                                                        // might need to
                                                                                                        // be improved
 
+    private static final ConcurrentMap<Class<?>, Optional<MethodHandle>> THINKING_HANDLER_CACHE = new ConcurrentHashMap<>();
```
geoand: If my comment about build time resolution is addressed, this is likely not going to be needed.

omatheusmesmo:
Hi @geoand , thanks, I updated the PR.

model-providers/openai/openai-vanilla/deployment/src/test/java/io/quarkiverse/langchain4j/openai/test/OpenAiThinkingEmittedTest.java
geoand, line 1: We usually use the OpenAI module for such kinds of tests

geoand (commented):
Looks nice! I just added one comment, but I also need @patriot1burke to review the chat scopes part

model-providers/openai/openai-vanilla/deployment/src/test/java/io/quarkiverse/langchain4j/openai/test/OpenAiThinkingEmittedTest.java
omatheusmesmo: I was thinking about it, I added to Ollama because of the original issue, but I can add to OpenAI as well. Do you think it's necessary to add a test to other providers too?

geoand: Just OpenAI will do

omatheusmesmo: Do you want me to remove the Ollama test or it's not necessary?

geoand: Let's remove it

core/runtime/src/main/java/io/quarkiverse/langchain4j/Thinking.java
```diff
@@ -0,0 +1,49 @@
...
+ * OpenAI, etc.).
+ * <p>
+ * For the streaming path, use the {@code onPartialThinking} handler exposed
+ * by {@code TokenStream} or observe {@code ChatEvent.PartialThinkingEvent}
+ * when returning {@code Multi<ChatEvent>}.
+ */
+@Target(ElementType.METHOD)
+@Retention(RUNTIME)
+@Documented
+public @interface Thinking {
```
geoand: Now that I think about it, wouldn't it be better to name this `@OnThinking`?

omatheusmesmo: > Now that I think about it, wouldn't it be better to name this `@OnThinking`?

 Good point.

geoand (approved):
Thanks a lot!

@patriot1burke can you also check?

chat-scopes/core/runtime/src/main/java/io/quarkiverse/langchain4j/chatscopes/ChatRouteThinking.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+        if (!ChatScope.isActive()) {
+            return;
+        }
+        ChatRouteContext context;
+        try {
+            context = CDI.current().select(ChatRouteContext.class).get();
+        } catch (Exception ignored) {
+            return;
+        }
```
geoand: Let's not use exceptions for flow control. `Instance` has methods that can be used to test if a bean is resolvable

omatheusmesmo: > Let's not use exceptions for flow control. `Instance` has methods that can be used to test if a bean is resolvable

Ok, fixed.

Issue #2457, opened 2026-05-15
The CDI behavior of Agents does not match that of AiServices


xstefank:
I ran into several surprises when I tried the new agentic framework. CDI doesn't work at all. So things like fault tolerance are a NOOP.

I talked a little with @mariofusco about this and we probably want to do this at least for top-level workflows. But I think that having this for all agents in the workflow would be better.

xstefank:
@mariofusco here is my demo if you want to test with it - https://github.com/xstefank/quarkus-smash-agents

geoand:
The first to note is that Agents are CDI beans as can be verified by:

```java
package io.xstefank.agents;

import static org.junit.jupiter.api.Assertions.assertNotNull;
import io.quarkus.test.junit.QuarkusTest;
import jakarta.inject.Inject;
import org.junit.jupiter.api.Test;

@QuarkusTest
class CdiBeansTest {

    @Inject
    AngerEvalAgent angerEvalAgent;

    @Inject
    AngerEvalWorkflow angerEvalWorkflow;

    @Inject
    DrBannerAgent drBannerAgent;

    @Inject
    HulkOutWorkflow hulkOutWorkflow;

    @Inject
    SmashSufficiencyAgent  smashSufficiencyAgent;

    @Inject
    TheHulkAgent  theHulkAgent;

    @Test
    void test() {
        assertNotNull(angerEvalAgent);
        assertNotNull(angerEvalWorkflow);
        assertNotNull(drBannerAgent);
        assertNotNull(hulkOutWorkflow);
        assertNotNull(smashSufficiencyAgent);
        assertNotNull(theHulkAgent);
    }
}

```

I need to check why Fault Tolerance is not working

geoand:
The fault tolerance issue stems from the fact that synthetic beans don't get automatic interceptor handling. There are probably a couple way to address this, none of which are great...

geoand:
https://github.com/quarkiverse/quarkus-langchain4j/pull/2467 fixes the interceptor issue

mariofusco:
Fixed by https://github.com/quarkiverse/quarkus-langchain4j/pull/2475

PR #2467, opened 2026-05-18
Add interceptor support to the agentic module

geoand:
- Relates to: #2457


cescoffier (approved):

mariofusco:
Correct me if I'm wrong, but this only solves part of the problem right? I mean that you're now enabling those interceptors, but still many annotations on agents (I tried with `@ModelName` for instance) don't work as expected because the agent is not the CDI one, but the one internally created by langchain4j. 

If I understand correctly the 2 issues are independent? Or solving the one that I mentioned would also implicitly solve the one that you're fixing with this pull request?

geoand:
Yes, I mentioned that in our chat. This solves the CDI interceptor part for agents that are already CDI beans

> If I understand correctly the 2 issues are independent

Yes.

This is why I only added `Relates to` instead of `Fixes` to the PR description

cescoffier:

PR #2468, opened 2026-05-18
Fix duplicate guardrail metrics issue

### LINKED ISSUE #2465: `guardrail` metrics double-registration with conflicting tag sets
## Summary

Micrometer meters
`guardrail.invoked`  **and** `guardrail.timed`
 are each registered **twice from inside qlc4j itself**, with
**two incompatible tag sets**. Prometheus rejects the second registration because its
contract requires all meters with the same name to share the same tag keys. The exception is
swallowed by `DefaultAiServiceListenerRegistrar` and surfaces only as a `WARN`, but the
guardrail metrics are silently lost from that point on.

## WARN log
```
WARN  [dev.langchain4j.observability.api.DefaultAiServiceListenerRegistrar]
An error occurred while firing event (DefaultOutputGuardrailExecutedEvent) to listener
(io.quarkiverse.langchain4j.observability.listener.AiServiceListenerAdapter):
Prometheus requires that all meters with the same name have the same set of tag keys.
There is already an existing meter named 'guardrail_invoked_total' containing tag keys
[class, exception, method, result, ...common tags].
The meter you are attempting to register has keys
[aiservice, guardrail, guardrail_type, operation, outcome, ...common tags].
```

(`...common tags` are user-supplied tags from a `MeterFilter.commonTags`. They are not part
of the bug — the conflict is between `[class, method, result, exception]` and
`[aiservice, guardrail, guardrail_type, operation, outcome]`.)

Truncated stack trace showing both call paths originate inside qlc4j:

```
GuardrailMetricsObserverSupport.recordMetrics(GuardrailMetricsObserverSupport.java:150)
GuardrailMetricsObserverSupport.onOutputGuardrailExecuted(...)
GuardrailMetricsObserver_Observer_onOutputGuardrailExecuted....notify(...)
EventImpl.fire(...)
AiServiceListenerAdapter.onEvent(AiServiceListenerAdapter.java:35)
DefaultAiServiceListenerRegistrar.fireEvent(DefaultAiServiceListenerRegistrar.java:66)
AbstractGuardrailExecutor.fireObservabilityEvent(AbstractGuardrailExecutor.java:123)
AbstractGuardrailExecutor.executeGuardrails(AbstractGuardrailExecutor.java:144)
OutputGuardrailExecutor.execute(OutputGuardrailExecutor.java:67)
GuardrailService.executeGuardrails(GuardrailService.java:93)
GuardrailsSupport.executeOutputGuardrails(GuardrailsSupport.java:54)
AiServiceMethodImplementationSupport.doImplement0(...)
... (around-invoke chain)
MetricsCountedWrapper.wrap(MetricsCountedWrapper.java:22)   ← legacy path also active
```

### Path 1  — bytecode-transformed `@Counted("guardrail.invoked")` and `@Timed("guardrail.timed")`

`GuardrailObservabilityProcessor#transformWithMetrics` (deployment) applies
`@io.micrometer.core.annotation.Counted("guardrail.invoked")` to guardrail `validate` methods
via Jandex bytecode transformation. The annotation is *not* present in user sources, which
makes it invisible to `grep` and explains why projects affected by this bug see no
`@Counted` anywhere.

At runtime, `MicrometerCountedInterceptor` (Quarkus-Micrometer) honours the synthesized
annotation and registers a counter with the standard `@Counted` tag set:
**`class`**, **`method`**, **`result`**, **`exception`**.

The same processor also synthesizes `@io.micrometer.core.annotation.Timed("guardrail.timed")`
on the same methods. `MicrometerTimedInterceptor` then registers a timer under
`guardrail.timed` with the standard `@Timed` tag set
(**`class`**, **`method`**, **`exception`**), producing the matching collision for the timer
meter.

Reference:
- `core/deployment/.../GuardrailObservabilityProcessor.java#transformWithMetrics`

### Path 2 — event-based `Counter.builder("guardrail.invoked")` + `Timer.builder("guardrail.timed")`

`GuardrailMetricsObserverSupport.recordMetrics` (runtime) listens to
`OutputGuardrailExecutedEvent` / `InputGuardrailExecutedEvent` via
`AiServiceListenerAdapter` and registers, in a single call:

- a **counter** under the same name `guardrail.invoked`, tags
  **`aiservice`**, **`operation`**, **`guardrail`**, **`guardrail.type`**, **`outcome`**
  (lines 142–151);
- a **timer** under the same name `guardrail.timed`, tags
  **`aiservice`**, **`operation`**, **`guardrail`**, **`guardrail.type`**, **`outcome`**
  (lines 153–163).

Both meters collide with their Path-1 counterparts.

Reference:
- `core/runtime/.../runtime/observability/GuardrailMetricsObserverSupport.java`
  lines 142–151 (qlc4j 1.9.2 sources jar):

  ```java
  Counter.builder("guardrail.invoked")
      .description("Number of guardrail invocations")
      .tags("aiservice", aiServiceName,
            "operation", methodName,
            "guardrail", guardrailName,
            "guardrail.type", guardrailType.getValue(),
            "outcome", outcome.getValue())
      .register(Metrics.globalRegistry)
      .increment();
  ```

## What I expected

Either:

- **A**: Only one of the two paths registers `guardrail.invoked`. Preferably the new
  event-based one, since its tag schema (`aiservice`, `guardrail`, `outcome`, …) is strictly
  richer than `[class, method, result, exception]`.
- **B**: The two paths use different metric names (e.g. `guardrail.invoked.legacy`
  vs `guardrail.invoked`).
- **C**: A config property (`quarkus.langchain4j.metrics.legacy.enabled=false` or similar)
  to disable the legacy `@Counted` injection. As of 1.10.0 no such property exists; the
  only gate is Micrometer-capability detection.

geoand:
- Fixes: #2465


cescoffier (approved):

cescoffier:

PR #2478, opened 2026-05-20
Pass tool executions in results

filipaggoncalves:
Fix: surface tool executions in Result<T> for normal tool calls

Previously, when an AI service method returned Result<T> and used tools with the default ReturnBehavior.DEFAULT (i.e., non-IMMEDIATE), the Result.toolExecutions() list was always empty: tool execution data was collected per iteration of the internal tool-calling loop but never passed to the final result. This change introduces an accumulator list (allToolExecutions) that collects every ToolExecution across all rounds of tool calling and passes it to the Result builder when the service completes, making tool execution history accessible to callers after the service returns. Also added a test to ToolExecutionModelTest to cover this scenario.

This is useful for programatically doing things based on what tools were called or what their result was AFTER the AI service executes.

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionModelTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionModelTest.java
@@ -302,6 +302,16 @@ public class ToolExecutionModelTest {
         }
     }
 
+    @Test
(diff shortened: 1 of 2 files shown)


geoand (approved):
Thanks!

+1 for me, but I would also like to ask @mariofusco if he sees any reason to not do this

mariofusco (approved):

PR #2479, opened 2026-05-20
Add support for multiple named Infinispan embedding store instances

### LINKED ISSUE #2438: Add support for multiple named Infinispan embedding store instances
Parent: #2415

## Description

Add support for configuring multiple named Infinispan embedding store instances, each with independent configuration and injectable via `@EmbeddingStoreName`.

Infinispan uses a CDI-managed `RemoteCacheManager` (via the `infinispan-client` extension), so named stores can map to a separate Infinispan client (different servers) or use the default client with different cache and schema configurations. Each named store may use a different `dimension`, which requires distinct Protobuf schema definitions and marshallers per store.

## Configuration example

```properties
quarkus.langchain4j.infinispan.products.dimension=1536
quarkus.langchain4j.infinispan.products.cache-name=product_embeddings
quarkus.langchain4j.infinispan.products.client-name=products-infinispan
```

## Injection

```java
@Inject @EmbeddingStoreName("products") EmbeddingStore<TextSegment> productsStore;
```

## Implementation pattern (suggestion)

Follow the same approach as pgvector (#2409):

- Refactor config to composition (`defaultConfig`/`namedConfig` with `@WithParentName`)
- Add `client-name` as a build-time property to `InfinispanNamedStoreBuildTimeConfig` (required for SmallRye `@WithDefaults` to create named config map entries, and needed to request the correct `InfinispanClientNameBuildItem` per store)
- Add `default-store-enabled` to the default build-time config
- Produce `SyntheticBeanBuildItem` with `@EmbeddingStoreName` qualifier
- Update the recorder to handle per-store configuration
- Update `SchemaAndMarshallerProducer` to produce Protobuf schemas and marshallers per named store (each `dimension` requires a different Protobuf message type)

cc @mcruzdev

mcruzdev:
Closes #2438


embedding-stores/infinispan/deployment/src/main/java/io/quarkiverse/langchain4j/infinispan/InfinispanEmbeddingStoreProcessor.java
```diff
@@ -39,43 +50,119 @@ FeatureBuildItem feature() {
...
+                .supplier(recorder.itemMarshallerSupplier(storeName))
+                .done();
+    }
+
+    private SyntheticBeanBuildItem buildMarshallerMetadataSyntheticBean(InfinispanEmbeddingStoreRecorder recorder,
+            String storeName) {
+        return SyntheticBeanBuildItem
+                .configure(LANGCHAIN_METADATA_MARSHALLER)
+                .identifier("langchain_metadata_" + storeName)
+                .types(ClassType.create(BaseMarshaller.class))
```
mcruzdev, line 154: @karesti The [`InfinispanClientProducer.handleProtoStreamMarshaller` calls beanManager.getBeans(BaseMarshaller.class) with no qualifier](https://github.com/quarkusio/quarkus/blob/24725bb3310948949c6cd7fe9a7226990959b6fa/extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientProducer.java#L424-L425), which per the CDI spec implicitly requires `@Default`. 

Any marshaller bean that carries an explicit qualifier loses `@Default` and becomes invisible to that lookup. (And I do not want multiple `@Default` beans).

Passing `Any.Literal.INSTANCE` instead (`beanManager.getBeans(BaseMarshaller.class, Any.Literal.INSTANCE)`) would return every registered marshaller regardless of qualifier, removing the need for this workaround. Is this `@Default` only lookup intentional, or an oversight from before multi-store scenarios existed?

karesti: no, it is not intentional it is probably a missing case if I understand well the issue.

PR #2481, opened 2026-05-20
Let tools return their own ToolExecutionResult

filipaggoncalves:
Right now, `QuarkusToolExecutor` always serialized a tool's return value through `Json.toJson()` to produce `resultText`, meaning that even if a tool returned a `ToolExecutionResult` directly, the LLM would just receive a JSON-serialized dump of the wrapper object rather than benefiting over the manual return type. This made it impossible to control the structured result and the LLM-facing text — for example, returning a valid complex object as the result of a tool while sending a formatted, human-readable summary as `resultText` to the model.

This change adds a passthrough: if the invocation result (or the resolved value of a Uni) is already a `ToolExecutionResult`, it is returned as-is, giving programmers control over the fields if they want/need it. It still preserves the existing serialization behavior for all other return types.

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolReturningToolExecutionResultTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolReturningToolExecutionResultTest.java
@@ -1,0 +1,127 @@
+package io.quarkiverse.langchain4j.test.tools;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+import java.util.function.Supplier;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.context.control.ActivateRequestContext;
+import jakarta.inject.Inject;
+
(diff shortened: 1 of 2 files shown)


geoand:
I can't say I understand this, so I would like to see an example. Also, a test will be needed

filipaggoncalves:
> I can't say I understand this, so I would like to see an example. Also, a test will be needed

@geoand Thanks, added the tests. As for the example, it could be something like: having a tool that calls an external API to fetch some sort of data that comes as a JSON object. You may want to keep that full json as a result to use it programatically for other things after the AI service executes, while not really wanting to provide that entire json to the model and instead feeding it a curated version of those results, or even together with some directives like "fetched items succesfully, item 1 is xxx, item 2 is yyy, you can tell the user to click button A for item1, (...)". As it is, you can only have one or the other, a curated tool result for the model or access to the "real" tool result after the AI service executes. There is the option of guardrails but it kind of isn't their goal, imo. This would give the option to manually curate tool results more in detail

geoand (approved):
Thanks!

Issue #2068, opened 2026-01-15
Provide a way to intercept and store chat history


chirino:
It would be nice if we could annotate the `@RegisterAiService` class methods with something like `@StoreHistory` annotation.

I ended building something like that in: https://github.com/chirino/memory-service?tab=readme-ov-file#5-conversation-history-recording using the `@RecordConversation`, but I think you guys could do an even better job.

Now the tricky part might be designing a good pluggable SPI.  A simple approach might be to just call some store interface with the user message and then again with the final agent response.   But my store also wants to make in flight  streaming responses resumable/cancelable so it needs to wrap the stream response too.   Oh and we need a way to make sure the user identity is propagated to those store interfaces.

geoand:
IIUC, the solution you came up with is https://github.com/chirino/memory-service/blob/bfd26f7d4690d180abb09706daa64ab15c2df80f/quarkus/memory-service-extension/runtime/src/main/java/io/github/chirino/memory/history/runtime/ConversationInterceptor.java, right?

Can you describe a little what the various pieces do?

Thanks

PR #2483, opened 2026-05-21
Add chat history recording for AI services

omatheusmesmo:
## Describe your changes

Applications often need to persist the conversation between a user and the model (to a database, an audit pipeline, a log, or a remote API) but Quarkus had no built-in way to intercept that exchange without coupling the recording logic to the AI service itself. This PR adds a pluggable recording mechanism.

A new `@RecordChatHistory` annotation marks a `@RegisterAiService` interface to opt in. The application provides a `ChatHistoryStore` CDI bean, and the runtime calls it at each step of the conversation turn. The two required callbacks (`onUserMessage`, `onAgentMessage`) are enough to record every turn; the remaining callbacks are optional `default` methods for finer-grained streaming behavior.

```java
@RecordChatHistory
@RegisterAiService
public interface SupportAgent {
    String chat(@MemoryId String conversationId, @UserMessage String message);
}

@ApplicationScoped
public class JpaChatHistoryStore implements ChatHistoryStore {

    @Inject ConversationRepository repository;

    @Override
    public void onUserMessage(Object memoryId, String userMessage) {
        repository.append(memoryId, Role.USER, userMessage);
    }

    @Override
    public void onAgentMessage(Object memoryId, String agentMessage) {
        repository.append(memoryId, Role.AGENT, agentMessage);
    }
}
```

Recording works for synchronous methods and for all streaming return types (`Multi<String>`, `Multi<ChatEvent>`, `TokenStream`). For streaming, partial chunks are buffered internally and the full text is delivered through `onAgentMessage` on completion; the optional `onAgentPartial`, `onCompleted`, `onCancelled`, and `onError` callbacks expose the rest of the stream lifecycle.

A store that throws is never allowed to break the conversation: every callback is invoked defensively and any `RuntimeException` is logged as a warning.

### Global mode

To record every AI service without annotating each one, set the build-time property:

```properties
quarkus.langchain4j.chat-history-enabled=true
```

### Deferred scope

The issue also describes two capabilities that are intentionally **not** part of this PR, since both are larger and orthogonal to recording:

- **Resumption**: replaying an in-flight streaming response to a new subscriber. This is a different mechanism (buffering + multicast + a replay SPI + durable storage) and is left as a complementary follow-up.
- **Automatic user identity propagation into the store**: The store is a CDI bean, so it can `@Inject SecurityIdentity` directly. This works on the calling thread for synchronous methods, but streaming callbacks run on Mutiny worker threads where the request scope may not be active; the documentation describes capturing identity at `onUserMessage` or relying on SmallRye Context Propagation. Automatic propagation could be added later.

This is why the PR references the issue without closing it.

---

- Relates to #2068

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -379,6 +379,7 @@ public class AiServicesProcessor {
     @BuildStep
     public void findDeclarativeServices(CombinedIndexBuildItem indexBuildItem,
             CustomScopeAnnotationsBuildItem customScopes,
+            LangChain4jBuildConfig config,
             List<AnnotationsImpliesAiServiceBuildItem> annotationsImpliesAiServiceItems,
             BuildProducer<RequestChatModelBeanBuildItem> requestChatModelBeanProducer,
             BuildProducer<RequestModerationModelBeanBuildItem> requestModerationModelBeanProducer,
@@ -589,7 +590,8 @@ public class AiServicesProcessor {
                             // we need to make these @DefaultBean because there could be other CDI beans of the same type that need to take precedence
                             impliedRegisterAiServiceTarget.contains(declarativeAiServiceClassInfo.name()),
                             shouldThrowExceptionOnEventError,
-                            chatMemoryFlushStrategySupplierClassDotName));
+                            chatMemoryFlushStrategySupplierClassDotName,
+                            hasRecordChatHistory(declarativeAiServiceClassInfo, config.chatHistoryEnabled())));
 
         }
         toolProviderProducer.produce(new ToolProviderMetaBuildItem(toolProviderInfos));
@@ -665,6 +667,10 @@ public class AiServicesProcessor {
             beanName = Optional.ofNullable(namedAnno.value().asString());
         }
         return beanName;
+    }
+
+    private static boolean hasRecordChatHistory(ClassInfo declarativeAiServiceClassInfo, boolean globalEnabled) {
+        return globalEnabled || declarativeAiServiceClassInfo.hasAnnotation(LangChain4jDotNames.RECORD_CHAT_HISTORY);
     }
 
     private static DotName cdiScope(CustomScopeAnnotationsBuildItem customScopes, ClassInfo declarativeAiServiceClassInfo) {
@@ -935,6 +941,7 @@ public class AiServicesProcessor {
         boolean needsModerationModelBean = false;
         boolean needsImageModelBean = false;
         boolean needsToolProviderBean = false;
+        boolean needsChatHistoryStoreBean = false;
         Set<DotName> allToolNames = new HashSet<>();
         Set<DotName> allToolProviders = new HashSet<>();
         Set<DotName> allToolHallucinationStrategies = new HashSet<>();
@@ -1098,7 +1105,8 @@ public class AiServicesProcessor {
                                     bi.getMaxToolCallsPerResponse(),
                                     allowContinuousForcedToolCalling,
                                     bi.isShouldThrowExceptionOnEventError(),
-                                    defaultMemoryIdProviderClassName)))
+                                    defaultMemoryIdProviderClassName,
+                                    bi.isRecordChatHistory())))
                     .setRuntimeInit()
                     .addQualifier()
                     .annotation(LangChain4jDotNames.QUARKUS_AI_SERVICE_CONTEXT_QUALIFIER).addValue("value", serviceClassName)
@@ -1219,6 +1227,11 @@ public class AiServicesProcessor {
                             new Type[] { ClassType.create(OutputGuardrail.class) }, null))
                     .done();
 
+            if (bi.isRecordChatHistory()) {
+                configurator.addInjectionPoint(ClassType.create(LangChain4jDotNames.CHAT_HISTORY_STORE));
+                needsChatHistoryStoreBean = true;
+            }
+
             syntheticBeanProducer.produce(configurator.done());
         }
 
@@ -1254,6 +1267,9 @@ public class AiServicesProcessor {
         }
         if (!allToolHallucinationStrategies.isEmpty()) {
             unremovableProducer.produce(UnremovableBeanBuildItem.beanTypes(allToolHallucinationStrategies));
+        }
+        if (needsChatHistoryStoreBean) {
+            unremovableProducer.produce(UnremovableBeanBuildItem.beanTypes(LangChain4jDotNames.CHAT_HISTORY_STORE));
         }
     }
 
@@ -1818,8 +1834,13 @@ public class AiServicesProcessor {
                         .map(AiServicesProcessor::classOutputGuardrails)
                         .orElse(null);
 
+                var recordChatHistory = aiServiceBuildItem
+                        .map(DeclarativeAiServiceBuildItem::isRecordChatHistory)
+                        .orElse(false);
+
                 perClassMetadata.put(ifaceName,
-                        new AiServiceClassCreateInfo(perMethodMetadata, implClassName, inputGuardrails, outputGuardrails));
+                        new AiServiceClassCreateInfo(perMethodMetadata, implClassName, inputGuardrails, outputGuardrails,
+                                recordChatHistory));
                 // make the constructor accessible reflectively since that is how we create the instance
                 reflectiveClassProducer.produce(ReflectiveClassBuildItem.builder(implClassName).build());
             }
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/DeclarativeAiServiceBuildItem.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/DeclarativeAiServiceBuildItem.java
@@ -43,6 +43,7 @@ public final class DeclarativeAiServiceBuildItem extends MultiBuildItem {
     private final boolean makeDefaultBean;
     private final boolean shouldThrowExceptionOnEventError;
     private final DotName chatMemoryFlushStrategySupplierClassDotName;
+    private final boolean recordChatHistory;
 
     public DeclarativeAiServiceBuildItem(
             ClassInfo serviceClassInfo,
@@ -71,7 +72,8 @@ public final class DeclarativeAiServiceBuildItem extends MultiBuildItem {
             Integer maxToolCallsPerResponse,
             boolean allowContinuousForcedToolCalling,
             boolean makeDefaultBean, boolean shouldThrowExceptionOnEventError,
-            DotName chatMemoryFlushStrategySupplierClassDotName) {
+            DotName chatMemoryFlushStrategySupplierClassDotName,
+            boolean recordChatHistory) {
         this.serviceClassInfo = serviceClassInfo;
         this.chatLanguageModelSupplierClassDotName = chatLanguageModelSupplierClassDotName;
         this.streamingChatLanguageModelSupplierClassDotName = streamingChatLanguageModelSupplierClassDotName;
@@ -100,6 +102,7 @@ public final class DeclarativeAiServiceBuildItem extends MultiBuildItem {
         this.makeDefaultBean = makeDefaultBean;
         this.shouldThrowExceptionOnEventError = shouldThrowExceptionOnEventError;
         this.chatMemoryFlushStrategySupplierClassDotName = chatMemoryFlushStrategySupplierClassDotName;
+        this.recordChatHistory = recordChatHistory;
     }
 
     public ClassInfo getServiceClassInfo() {
@@ -241,5 +244,9 @@ public final class DeclarativeAiServiceBuildItem extends MultiBuildItem {
 
     public void setDefaultMemoryIdProviderClassDotName(DotName defaultMemoryIdProviderClassDotName) {
         this.defaultMemoryIdProviderClassDotName = defaultMemoryIdProviderClassDotName;
+    }
+
+    public boolean isRecordChatHistory() {
+        return recordChatHistory;
     }
 }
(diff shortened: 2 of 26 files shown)


geoand:
@chirino is this similar to what you had in mind?

chirino:
It's very close to what we needed in the [memory-service history recording](https://chirino.github.io/memory-service/docs/quarkus/conversation-history/).  The SPI need to be able get some more metadata from the agent: agentId, fork/start ids, attachments, indexed content, bearer token, and user ID.

Th PR's recording attaches directly to @RegisterAiService; the memory-service impl use a wrapper like HistoryRecordingAgent, which lets you add non-LangChain4j parameters such as fork ids and attachments.

The Multi<ChatEvent> result stream path also needs some thought.  In the memory-service we basically wrap the ChatEvent into a richer even adding things attachment references (the memory-service handles storing generated files in storage for later retrieval).

I would need to think about it to see if I could add the additional features we provide around this simpler SPI.

geoand:
Thanks @chirino

omatheusmesmo:
Thanks @chirino. I went through memory-service to understand the gap, and the extra metadata you listed splits into two kinds, neither of which the core can synthesize from a plain `@RegisterAiService` method:

- per-call values your `HistoryRecordingAgent` wrapper passes as method parameters: fork/start ids (`@ForkedAtConversationId` etc.) and an `Attachments` parameter (plus an `@AgentId` slot the interceptor already supports);
- request/security-scoped values you resolve inside your store (user id, bearer token, indexed content).

So this PR keeps the recording SPI minimal (memory id + message), which is enough to record every turn, and leaves that richer metadata to the layer on top.

The open question is whether we want the core SPI to carry agent-level metadata at all, or leave that to the wrapper layer.

I'd be glad to follow up on this if you both agree it's worthwhile. I wanted your feedback before expanding it into a public SPI that all providers can use.

WDYT @geoand @chirino ?

sabob:
The description mentions "audit pipeline". 
But the PR only emits the User and LLM response messages?
For Auditing one would also need the ToolRequests and ToolResponses from the LLM, so we know the full exchange.
We currently use Listeners for auditing, which works well. Was just wondering if this PR could be exspanded to be used for auditing purposes.

Kind regards
Bob

chirino:
@sabob Keep in mind that the ChatEvent  response stream includes the tool calls and tool results. 

That's another reason why history should a way of filtering/transforming the ChatEvent stream into a different representation which is what actually gets streamed to end users.  You might want to hide tool call parameters from user for example.

I think history recording should be focused on capturing the data the user sent and was shown.  Which can be different from what is happening in LLM interactions.

In the https://chirino.github.io/memory-service/ we capture all of those User/Agent interactions as history entries in the conversation.  An then it provides ChatMemoryStore to capture all the messages sent to the LLM as context entries in the conversation.  You end up with a nice auditable view what was exchanged between the user and agent and what was exchanged between the agent and the LLM.

omatheusmesmo:
Hello @sabob ,

I would like to add more enhancements to this PR, but I'm still waiting for some kind of approval/guidance here, not sure if I should invest more time on adding more features to the current PR or if we should just let `memory-service` handle everything... 

I was expecting to be worth add something about it on Quarkus LangChain4j, but if it's not, I can just drop this PR without problems, but I think we should close the issue as not planned.

geoand:
https://github.com/quarkiverse/quarkus-langchain4j/issues/2068 was originally opened because @chirino needed that feature, so I'd really like him to weigh in on whether what is provided here is good enough or needs to be changed and/or enhanced

chirino:
I do think a standardized interface would be a good thing.  But I think it's important that we validate that in the standardized interface theres a way to pass additional per-request data to the SPI (passing bearer token would be good example), and that you can record a agent defined translation of the `Multi<ChatEvent>` stream.

omatheusmesmo:
Following up here and on #2499 (cc @chirino @sabob @geoand). Before expanding the SPI I wanted to map what data the core already exposes, so we can agree on the shape instead of inventing new event types.

Most of what's been asked for already exists in three places:

**1. The `ChatEvent` stream (`Multi<ChatEvent>`).** This already carries the full exchange, not just text: `BeforeToolExecutionEvent` (the `ToolExecutionRequest`), `ToolExecutedEvent` (request + result), `ContentFetchedEvent` (RAG), `IntermediateResponseEvent` (per-turn thinking + tool requests), plus the partial thinking/tool-call events for streaming. So @chirino's point about recording a translation of the `Multi<ChatEvent>` stream and @sabob's need for tool calls/results both land here.

**2. The observability events** already wired through `AiServiceEvents`: `AiServiceStartedEvent` (system + user message), `AiServiceResponseReceivedEvent` (request/response with token metadata), `ToolExecutedEvent`, `AiServiceErrorEvent`, etc. These give the per-round-trip view and the system prompt that the `ChatEvent` stream doesn't.

**3. `InvocationContext`,** carried by every observability event. This looks like the answer to @chirino's "pass additional per-request data" point. It carries a per-request channel, `invocationParameters()`, threaded through when the method declares an `InvocationParameters` parameter (the same mechanism guardrails use via `ToolInvocationContext`). It's captured on the calling thread, so a snapshot survives into the streaming callbacks. Plus `chatMemoryId()`, `invocationId()`, `interfaceName()`/`methodName()`, `methodArguments()` and `timestamp()`. So an app puts a bearer token / user id in `invocationParameters()` at the edge and the store reads it back, with no provider-specific coupling.

What the core *can't* synthesize is the typed business params your `HistoryRecordingAgent` declares (`@ForkedAtConversationId`, `@AgentId`, an `Attachments` parameter) since those only exist because they're method parameters on the wrapper. So my suggestion is: the core SPI carries the raw/generic data plus a stream-transform hook, and the typed wrapper stays a layer on top, the way memory-service already does it.

Concretely that means changing the current callbacks to take an immutable context snapshot instead of a bare `Object memoryId`:

```java
void onUserMessage(RecordingContext ctx, String userMessage);
void onAgentMessage(RecordingContext ctx, String agentMessage);

// observe / translate the full event stream
default Multi<ChatEvent> onAgentEvents(RecordingContext ctx, Multi<ChatEvent> events) {
    return events;
}
```

where `RecordingContext` is built once on the calling thread (so identity is still valid when later callbacks run on Mutiny worker threads) and exposes `memoryId()`, `invocationId()`, `interfaceName()`, `methodName()`, `timestamp()` and an `attributes()` map sourced from `invocationParameters()`.

For this PR that means the context object plus the `onAgentEvents` transform hook. The typed business params stay in the wrapper layer, and resumption remains separate as already noted.

Two things I'd like your input on before I touch the code:

- @chirino is the free-form `attributes()` map enough for the bearer token / user id, or would you want a typed identity accessor?
- @sabob for tool auditing: the observability `ToolExecutedEvent` (the one you already use) is fired for every tool execution, including immediate-return ones, before any return-behavior branch, so they do produce a `ToolExecutedEvent` today. On top of that the synchronous path already collects the full tool-execution list and carries it in the final `Result`, so if we wanted tool recording in this SPI it's feasible there too, without forcing `Multi<ChatEvent>`. If you saw the event not fire, could you share the return type / scenario? That tells us whether tool recording belongs in this SPI or stays with the observability events you already rely on.

chirino:
re: free-form attributes() .. is that a `Map<String, Object>` ? How would my wrapper API add stuff to it?

Also would your proposal support doing something like:

```
Multi<MyCustomEvent> onAgentEvents(RecordingContext ctx, Multi<ChatEvent> events) {
    return translate(events);
}
```

I don't want to be restricted to transforming to `Multi<ChatEvent> `

omatheusmesmo:
> is that a `Map<String, Object>`? How would my wrapper API add stuff to it?

Yes @chirino, it's LangChain4j's `InvocationParameters`, a thread-safe `Map<String, Object>`. Your wrapper writes to it the documented way: declare an `InvocationParameters` parameter and pass it in, e.g. `InvocationParameters.from(Map.of("bearerToken", token))`. The runtime threads that same instance into the `InvocationContext`, so the store reads it back and the LLM never sees it. The one caveat is that it's stringly-typed, so typed params like `@ForkedAtConversationId` or an `Attachments` arg don't land there automatically: the wrapper copies them in, or we add a typed accessor if you'd rather.

> would your proposal support doing something like `Multi<MyCustomEvent> onAgentEvents(...)` [...] I don't want to be restricted to transforming to `Multi<ChatEvent>`

For recording you're not restricted at all: you get the live `Multi<ChatEvent>`, read whatever you need and persist it in any shape you want, which is what your event adapter already does. What the hook can't do is change the type the subscriber receives, because an AI service method only builds as `Multi<String>` or `Multi<ChatEvent>`, and the runtime needs `ChatEvent` to accumulate the final text and fire the terminal callbacks. The `Multi<MyCustomEvent>` mapping is really a presentation concern and fits at the SSE/REST edge, where memory-service already does it. If you want a sanctioned in-pipeline transform separate from recording, I think that's its own extension point rather than this SPI.

chirino:
I think recording is a presentation concern.  It's about recording the chat input and output.  If the chat output is a Multi<MyCustomEvent>, then that's what should be getting recorded.  Similarly, Input can be more complex.  How do you record a chat input message that has file type that is not supported by the existing Content types?

omatheusmesmo:
> I think recording is a presentation concern.  It's about recording the chat input and output.  If the chat output is a Multi<MyCustomEvent>, then that's what should be getting recorded.  Similarly, Input can be more complex.  How do you record a chat input message that has file type that is not supported by the existing Content types? 
> 

Hi @chirino , I'm so sorry, I missed the notification entirely, I'll answer you properly in a few days.

omatheusmesmo:
> How do you record a chat input message that has file type that is not supported by the existing Content types?

Short answer on the file question: you can't record it through the core SPI, and I think that's the right outcome. A file with no matching `Content` type never becomes part of the `UserMessage` the AI service builds, there's no `Content` for it and no declarative parameter that yields one. At the recording point it's simply not there - which is exactly why memory-service records it in its own interceptor, and I'd keep it there.

What I will fix is the input being too thin: today it flattens to `String`, dropping even the supported multimodal `Content`. I'll change `onUserMessage` to carry the full `UserMessage` , since the message is already built before recording.

Does that boundary work for you?

chirino:
I can see the perspective your taking on this of just recording the Input and output of what's going in and out of langchain4j api's but I don't think that's sufficient.

What user want to record is all the data that was exchanged between the user and the agent for a turn.   And this data will likely be filtered down stripping out event details, and supplemented  (recording additional data like files submitted in the turn, maybe git shas etc).  But what is included or excluded is going to be 100% controlled by the agent app and using the types it defines.

chirino:
I'm thinking we want users to call something like:

```java
public interface Recorder {

    <IN, OUT> Multi<OUT> record(
            IN recordedInput,
            Supplier<Multi<ChatEvent>> agentCall,
            Function<Multi<ChatEvent>, Multi<OUT>> responseStreamMapper);
}
```

This would allow user to write resource methods like:

```java
public Multi<MyResponse> chat(FancyRequest request) {
    return recorder.record(
            request,
            () -> agent.chat(
                    request.memoryId(),
                    request.message(),
                    request.attachments()
            ),
            AppAgentMapping::toResponseStream
    );
}
```

For the simple cases where there is no transformation, I'm sure we can provide some simpler method overloads.

omatheusmesmo:
Hi @chirino .

`memory-service` is a solid piece of work, and I'm not trying to win a design argument against it. What I want is a recording capability in the extension that serves everyone: memory-service, an audit log, a JPA table, a Kafka topic, a remote API. That's the pluggable SPI you asked for when you opened #2068.

> What user want to record is all the data that was exchanged between the user and the agent for a turn ... what is included or excluded is going to be 100% controlled by the agent app and using the types it defines.

I agree with the control part: what gets stripped, kept, or supplemented (files, git shas) should be the app's call, in the app's own types. But that filtering is downstream of a complete turn. To filter the exchange down you first need the whole exchange to filter from, and to attach supplemental data to the right turn you need a per-request channel to correlate it. So the split I'd draw is: the extension hands over the complete canonical turn plus that channel, and the app decides what to keep and what to add.

Concretely, that's:

- The full `UserMessage` as input, not flattened text, so supported multimodal content survives.
- The live `Multi<ChatEvent>` stream, so the consumer sees the whole exchange including tool calls and results.
- A per-request `RecordingContext` captured on the calling thread (so identity is valid on the Mutiny worker threads), with an `attributes()` channel sourced from `InvocationParameters`, so the app passes a bearer token, user id, git sha, etc. through with no provider coupling.

(The code on the PR today is a first, primitive cut of this; the points above are the refinement I want to build.)

On the `Recorder`:

```java
<IN, OUT> Multi<OUT> record(
        IN recordedInput,
        Supplier<Multi<ChatEvent>> agentCall,
        Function<Multi<ChatEvent>, Multi<OUT>> responseStreamMapper);
```

The extension already has the piece that transforms a streaming response, and it's worth looking at its shape. `AiResponseAugmenter<T>`, selected with `@ResponseAugmenter`, rewrites the response including the `Multi<T>` form, and the runtime applies it inside the AI service invocation (`ResponseAugmenterSupport` resolves the bean and calls `augment(...)`). Every extension point here is driven the same way: the framework invokes your code inside the flow, you don't assemble the call. You implement an interface and annotate (`AiResponseAugmenter`, guardrails, `SystemMessageProvider`, `ChatMemorySeeder`), or you observe a fired event (the observability events @sabob already uses for audit). There is no API where the caller assembles the invocation and calls in.

`record()` is the opposite shape: the caller supplies the agent call as a `Supplier`, maps the stream with a `Function`, and calls `record()` above the AI service. It touches only public types (`Multi`, `ChatEvent`, and the app's `IN`/`OUT`), uses no extension internals, and the framework applies nothing. That's a generic reactive tee-and-map. It doesn't match anything the extension exposes today, and it would work just as well as a small user utility with no LangChain4j dependency.

So the canonical-turn recording fits the existing idiom (annotate, implement, applied inside the invocation, streaming form included, the same shape as `AiResponseAugmenter`), and `record()` doesn't. There's a design consequence as well: because `record()` records the app-mapped `OUT`, the recorded shape is app-defined, so @sabob's audit case in #2499 can't rely on it, since the mapper may have dropped the tool calls, results and token metadata an audit log needs.

@geoand this is really a scope call, and you and Clement Escoffier both liked the feature on #2068. Two shapes are on the table: a declarative SPI over the canonical turn, in the same shape as `AiResponseAugmenter` and applied inside the invocation (this PR's direction), and @chirino's `Recorder`, a reactive wrapper the app assembles over public types. My read is that the first matches how the extension exposes everything else and the second doesn't, but the call is yours. I have an early prototype of the first. Which should I build toward?

PR #2486, opened 2026-05-21
Introduce a way to resolve agentic method parameters from CDI

geoand:
Requires https://github.com/langchain4j/langchain4j/pull/5286

### DIFF
--- a/agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/CdiChatSupplierParameterResolverTest.java
+++ b/agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/CdiChatSupplierParameterResolverTest.java
@@ -1,0 +1,148 @@
+package io.quarkiverse.langchain4j.agentic.deployment;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import jakarta.enterprise.inject.Produces;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import dev.langchain4j.agentic.Agent;
+import dev.langchain4j.agentic.declarative.ChatModelSupplier;
+import dev.langchain4j.agentic.declarative.SequenceAgent;
+import dev.langchain4j.model.chat.ChatModel;
+import dev.langchain4j.service.UserMessage;
+import dev.langchain4j.service.V;
+import io.quarkiverse.langchain4j.ModelName;
+import io.quarkiverse.langchain4j.agentic.runtime.CdiBean;
+import io.quarkiverse.langchain4j.openai.testing.internal.OpenAiBaseTest;
+import io.quarkiverse.langchain4j.testing.internal.WiremockAware;
+import io.quarkus.test.QuarkusUnitTest;
+
+public class CdiChatSupplierParameterResolverTest extends OpenAiBaseTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest unitTest = new QuarkusUnitTest()
+            .setArchiveProducer(
+                    () -> ShrinkWrap.create(JavaArchive.class)
+                            .addClasses(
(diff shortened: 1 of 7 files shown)


geoand:
@mariofusco this can now be reviewed

agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/CdiChatSupplierParameterResolverTest.java
```diff
@@ -0,0 +1,123 @@
...
+    static final String SELECTED_RESPONSE = "selected-by-cdi-bean";
+
+    @ApplicationScoped
+    public static class ModelSelector {
+
+        public ChatModel select(String input) {
+            if (input.contains("a")) {
+                return new Agents.FixedResponseChatModel(SELECTED_RESPONSE);
+            }
+            return new Agents.EchoResponseChatModel();
```
mariofusco, line 73: I guess this is not very relevant for the test purposes, but maybe it is better to use the best possible setup also in our tests, even because users could take them as examples for their own code.

The problem here is that at each invocation of this method it will create a new `ChatModel`, while each one of them should be a singleton also injected via CDI and referenced through the `@ModelName` annotation as for instance has been done here https://github.com/quarkiverse/quarkus-langchain4j/blob/e7f6c38367222b8e232526b1366dfe5d35885098/agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/DynamicModelSelectionTest.java#L81

agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/CdiChatSupplierParameterResolverTest.java
```diff
@@ -0,0 +1,123 @@
...
+    public interface EchoAgent {
+
+        @UserMessage("{{text}}")
+        @Agent(description = "An agent echoing its input", outputKey = "echo")
+        String echo(String text);
+
+        @ChatModelSupplier
+        static ChatModel chatModel() {
+            return new Agents.EchoResponseChatModel();
+        }
```
mariofusco, line 86: Using the pattern I mentioned above, this could also be removed, and a fixed model can be assigned to an agent also via the `@ModelName` annotation as done here https://github.com/quarkiverse/quarkus-langchain4j/blob/e7f6c38367222b8e232526b1366dfe5d35885098/agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/DynamicModelSelectionTest.java#L58

agentic/deployment/pom.xml
```diff
@@ -9,6 +9,11 @@
   <artifactId>quarkus-langchain4j-agentic-deployment</artifactId>
   <name>Quarkus LangChain4j - Agentic - Deployment</name>
   <dependencies>
+    <dependency>
+      <groupId>dev.langchain4j</groupId>
+      <artifactId>langchain4j-agentic</artifactId>
+      <version>1.16.0-beta26-SNAPSHOT</version>
+    </dependency>
```
mariofusco: This can be removed now.

geoand: Ah, I missed it, thanks!

agentic/runtime/pom.xml
```diff
@@ -17,6 +17,7 @@
     <dependency>
       <groupId>dev.langchain4j</groupId>
       <artifactId>langchain4j-agentic</artifactId>
+      <version>1.16.0-beta26-SNAPSHOT</version>
```
mariofusco: Remove this.

geoand: Oops, now I see that I forgot to push the changes I had already made...

agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/CdiChatSupplierParameterResolverTest.java
geoand: Good point. Fixed

mariofusco (approved):
Ok for me, but is that huge change in documentation related to this or an unwanted leftover? It is so big that I even cannot visualize it.

geoand:
I'll revert that doc change. I have yet to understand what the hell happens with the docs update every now and then...

PR #2496, opened 2026-05-25
Pass missing fields from ToolExecutionResult to ToolExecutionResultMessage

filipaggoncalves:
Currently, `isError` and `attributes` aren't being propagated from `ToolExecutionResult` to `ToolExecutionResultMessage`, always leaving these empty and losing that information in persistence. This fix passes these fields' values through while keeping all other behavior the same.

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/ToolExecutionResultMessageUtil.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/ToolExecutionResultMessageUtil.java
@@ -1,0 +1,25 @@
+package io.quarkiverse.langchain4j.runtime.aiservice;
+
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+import java.util.List;
+
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.data.message.TextContent;
+import dev.langchain4j.data.message.ToolExecutionResultMessage;
+import dev.langchain4j.service.tool.ToolExecutionResult;
+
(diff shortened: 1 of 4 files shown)


core/runtime/src/test/java/io/quarkiverse/langchain4j/runtime/aiservice/ToolExecutionResultMessageUtilTest.java
geoand, line 1: I would much prefer we have a proper integration test instead of this unit test

geoand (commented):
Thanks a lot for this!

I added a comment about the test

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/ToolExecutionResultMessageUtil.java
```diff
@@ -0,0 +1,25 @@
...
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+import java.util.List;
+
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.data.message.TextContent;
+import dev.langchain4j.data.message.ToolExecutionResultMessage;
+import dev.langchain4j.service.tool.ToolExecutionResult;
+
+public class ToolExecutionResultMessageUtil {
```
geoand: This can be made package-private, right?

filipaggoncalves: absolutely, fixed

geoand (approved):
Thanks!

PR #2497, opened 2026-05-25
Add audio transcription support to the OpenAI extension

### LINKED ISSUE #1502: Is OpenAI able to send AudioContent?
I am trying to use OpenAI Speech to text https://platform.openai.com/docs/guides/speech-to-text. Is the OpenAI supporting it?

```java
@RegisterAiService
public interface TimelessAiService {
    String expense(AudioContent audioMessage);
}
```

I expected to receive the audio transcription, is it possible?

omatheusmesmo:
## Describe your changes

Exposes `AudioTranscriptionModel` (Whisper) as an injectable CDI bean in `quarkus-langchain4j-openai`, backed by langchain4j's `OpenAiAudioTranscriptionModel`.

It follows the pattern of the other auxiliary models (`EmbeddingModel`, `ImageModel`, `ModerationModel`): the bean is selected by provider/config and used directly, rather than through a declarative `@RegisterAiService` method.

Compared to building `OpenAiAudioTranscriptionModel` directly, the bean reuses the Quarkus REST client (native-image friendly), the TLS and proxy registries, externalized configuration and the CDI lifecycle, instead of langchain4j's default HTTP stack.

The non-trivial part is in `openai-common`: `QuarkusOpenAiClient` did not implement `audioTranscription`, so the upstream default threw `UnsupportedOperationException` and the bean would fail on first call. This PR adds the `audio/transcriptions` endpoint to `OpenAiRestApi` and builds the multipart body programmatically through `ClientMultipartForm`, so the `file` part carries the real filename derived from the audio's mime type (Whisper uses the extension as a format hint).

Per-request options (`language`, `prompt`, `temperature`) are passed via `AudioTranscriptionRequest`; only `model-name`, `log-requests` and `log-responses` are configuration keys, plus the build-time `enabled` flag that mirrors the other models.

Validated end-to-end against an OpenAI-compatible Whisper endpoint (Groq `whisper-large-v3`), with both `mp3` and `ogg` inputs.

## Configuration

```properties
quarkus.langchain4j.openai.api-key=sk-...
quarkus.langchain4j.openai.audio-transcription-model.model-name=whisper-1
```

## Injection

```java
@ApplicationScoped
public class TranscriptionService {

    @Inject
    AudioTranscriptionModel audioTranscriptionModel;

    public String transcribe(byte[] audioBytes, String mimeType) {
        AudioTranscriptionRequest request = AudioTranscriptionRequest.builder()
                .audio(Audio.builder().binaryData(audioBytes).mimeType(mimeType).build())
                .build();
        return audioTranscriptionModel.transcribe(request).text();
    }
}
```

The `mimeType` determines the filename sent in the multipart request (e.g. `audio.mp3`), which Whisper uses as a format hint.

---

- Closes #1502

### DIFF
--- a/docs/modules/ROOT/pages/openai-audio-transcription-model.adoc
+++ b/docs/modules/ROOT/pages/openai-audio-transcription-model.adoc
@@ -1,0 +1,67 @@
+= OpenAI Audio Transcription Models
+
+include::./includes/attributes.adoc[]
+include::./includes/customization.adoc[]
+
+OpenAI provides audio transcription models (Whisper) capable of converting spoken audio into text.
+These models are useful for use cases such as building voice assistants, transcribing meetings, or indexing podcast content.
+
+== Prerequisites
+
+include::./openai-chat-model.adoc[tags=openai-prerequisites]
+
+=== OpenAI Quarkus Extension
+
+To use OpenAI audio transcription models in your Quarkus application, add the `quarkus-langchain4j-openai` extension:
+
+:provider-artifact: quarkus-langchain4j-openai
+include::./includes/quarkus-langchain4j-maven-dependencies.adoc[]
+
+== Configuration
+
+include::includes/quarkus-langchain4j-openai.adoc[leveloffset=+1,opts=optional]
+
+The minimum configuration to use audio transcription is the OpenAI API key (the model name defaults to `whisper-1`):
+
+[source,properties]
+----
+quarkus.langchain4j.openai.api-key=sk-...
+----
+
+Any OpenAI-compatible transcription endpoint can be used by overriding `quarkus.langchain4j.openai.base-url`.
+
+== Using Audio Transcription Models
+
+Inject the `AudioTranscriptionModel` bean and call `transcribe(...)` with an `AudioTranscriptionRequest`:
+
+[source,java]
+----
+package org.acme;
+
+import dev.langchain4j.data.audio.Audio;
+import dev.langchain4j.model.audio.AudioTranscriptionModel;
+import dev.langchain4j.model.audio.AudioTranscriptionRequest;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+
+@ApplicationScoped
+public class TranscriptionService {
+
+    @Inject
+    AudioTranscriptionModel audioTranscriptionModel;
+
+    public String transcribe(byte[] audioBytes, String mimeType) {
+        AudioTranscriptionRequest request = AudioTranscriptionRequest.builder()
+                .audio(Audio.builder().binaryData(audioBytes).mimeType(mimeType).build())
+                .build();
+        return audioTranscriptionModel.transcribe(request).text();
+    }
+}
+----
+
+Per-request options such as `language`, `prompt`, and `temperature` are passed via the `AudioTranscriptionRequest` builder.
+
+== Additional Resources
+
+[.lead]
+* Explore xref:ai-services.adoc[AI Services] for orchestrating multiple models
--- a/model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/common/QuarkusOpenAiClient.java
+++ b/model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/common/QuarkusOpenAiClient.java
@@ -19,10 +19,12 @@ import jakarta.enterprise.inject.Instance;
 import jakarta.enterprise.inject.spi.CDI;
 import jakarta.ws.rs.client.ClientRequestContext;
 import jakarta.ws.rs.client.ClientRequestFilter;
+import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.MultivaluedHashMap;
 import jakarta.ws.rs.core.MultivaluedMap;
 
 import org.eclipse.microprofile.rest.client.ext.ClientHeadersFactory;
+import org.jboss.resteasy.reactive.client.api.ClientMultipartForm;
 import org.jboss.resteasy.reactive.client.api.LoggingScope;
 
 import dev.langchain4j.model.chat.response.StreamingHandle;
@@ -35,6 +37,8 @@ import dev.langchain4j.model.openai.internal.StreamingCompletionHandling;
 import dev.langchain4j.model.openai.internal.StreamingResponseHandling;
 import dev.langchain4j.model.openai.internal.SyncOrAsync;
 import dev.langchain4j.model.openai.internal.SyncOrAsyncOrStreaming;
+import dev.langchain4j.model.openai.internal.audio.transcription.OpenAiAudioTranscriptionRequest;
+import dev.langchain4j.model.openai.internal.audio.transcription.OpenAiAudioTranscriptionResponse;
 import dev.langchain4j.model.openai.internal.chat.ChatCompletionRequest;
 import dev.langchain4j.model.openai.internal.chat.ChatCompletionResponse;
 import dev.langchain4j.model.openai.internal.completion.CompletionRequest;
@@ -46,6 +50,7 @@ import dev.langchain4j.model.openai.internal.image.GenerateImagesResponse;
 import dev.langchain4j.model.openai.internal.moderation.ModerationRequest;
 import dev.langchain4j.model.openai.internal.moderation.ModerationResponse;
 import dev.langchain4j.model.openai.internal.spi.OpenAiClientBuilderFactory;
+import io.netty.buffer.Unpooled;
 import io.quarkiverse.langchain4j.auth.ModelAuthProvider;
 import io.quarkiverse.langchain4j.openai.common.runtime.AdditionalPropertiesHack;
 import io.quarkus.rest.client.reactive.QuarkusRestClientBuilder;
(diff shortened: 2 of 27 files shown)


omatheusmesmo:
Hello @mcruzdev ,

Could you kindly see if is it what you expected? I would appreciate.

geoand:
Very cool, thanks!

I am wondering if we could have a better test than the one added....
Moreover, we'll need some docs for this

omatheusmesmo:
> Very cool, thanks!
> 
> I am wondering if we could have a better test than the one added.... Moreover, we'll need some docs for this

Thanks! 

I updated the test to assert the multipart request itself (endpoint, model field, mime-derived filename), not just the response. 

About the docs: do you think I need add more than this one?

https://github.com/omatheusmesmo/quarkus-langchain4j/blob/2be9cbdfe59fc0cd642b63e43358e9c32462bd6a/docs/modules/ROOT/pages/openai-audio-transcription-model.adoc

omatheusmesmo:
Hello @geoand , could you please trigger the CI again? Looks related to that GitHub incident.

https://www.githubstatus.com/incidents/gnftqj9htp0g

docs/modules/ROOT/pages/openai-audio-transcription-model.adoc
```diff
@@ -0,0 +1,75 @@
...
+AudioTranscriptionModel audioTranscriptionModel;
+
+// ...
+byte[] audioBytes = Files.readAllBytes(Path.of("meeting.mp3"));
+AudioTranscriptionRequest request = AudioTranscriptionRequest.builder()
+        .audio(Audio.builder().binaryData(audioBytes).mimeType("audio/mpeg").build())
+        .build();
+AudioTranscriptionResponse response = audioTranscriptionModel.transcribe(request);
+String text = response.text();
+----
```
geoand, line 60: It would be best if the example contained the full class so users can copy and paste

docs/modules/ROOT/pages/openai-audio-transcription-model.adoc
```diff
@@ -0,0 +1,75 @@
...
+Per-request options such as `language`, `prompt`, and `temperature` are passed via the `AudioTranscriptionRequest` builder.
+
+== Groq compatibility
+
+Groq exposes an OpenAI-compatible Whisper endpoint, so the same extension can target it by overriding the base URL and using one of Groq's audio models:
+
+[source,properties]
+----
+quarkus.langchain4j.openai.base-url=https://api.groq.com/openai/v1/
+quarkus.langchain4j.openai.api-key=${GROQ_API_KEY}
```
geoand: `GROQ_API_KEY`?

omatheusmesmo: Sorry about that. Groq is OpenAi Compatible and I tested using it since it has a good free tier, but should not be the example on documentation. I'll update.

docs/modules/ROOT/pages/openai-audio-transcription-model.adoc
omatheusmesmo: Done.

docs/modules/ROOT/pages/openai-audio-transcription-model.adoc
omatheusmesmo: Sorry about the Groq Section, @geoand , I added a small line about the OpenAI Compatible though.

```adoc
Any OpenAI-compatible transcription endpoint can be used by overriding `quarkus.langchain4j.openai.base-url`.
```

docs/modules/ROOT/pages/openai-audio-transcription-model.adoc
```diff
@@ -0,0 +1,68 @@
...
+== Configuration
+
+include::includes/quarkus-langchain4j-openai.adoc[leveloffset=+1,opts=optional]
+
+The minimum configuration to use audio transcription is the OpenAI API key and the model name (defaults to `whisper-1`):
+
+[source,properties]
+----
+quarkus.langchain4j.openai.api-key=sk-...
+quarkus.langchain4j.openai.audio-transcription-model.model-name=whisper-1
```
geoand: Isn't `whisper-1`  the default?

model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/common/QuarkusOpenAiClient.java
```diff
@@ -380,6 +385,65 @@ public Uni<GenerateImagesResponse> get() {
...
+    }
+
+    private static ClientMultipartForm buildAudioTranscriptionForm(OpenAiAudioTranscriptionRequest request) {
+        String fileName = request.file().fileName();
+        String mimeType = request.file().mimeType();
+        if (mimeType == null || mimeType.isBlank()) {
+            mimeType = MediaType.APPLICATION_OCTET_STREAM;
+        }
+        ClientMultipartForm form = ClientMultipartForm.create()
+                .binaryFileUpload("file", fileName, Buffer.buffer(request.file().content()), mimeType)
```
geoand: `Buffer.buffer(request.file().content())` will add the entire file's content in memory, which could be really problematic. We definitely need to avoid it

docs/modules/ROOT/pages/openai-audio-transcription-model.adoc
omatheusmesmo: I'm sorry about this.

model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/common/QuarkusOpenAiClient.java
omatheusmesmo: Thanks for pointing out.

model-providers/openai/openai-common/runtime/src/main/java/io/quarkiverse/langchain4j/openai/common/QuarkusOpenAiClient.java
```diff
@@ -380,6 +386,66 @@ public Uni<GenerateImagesResponse> get() {
...
+
+    private static ClientMultipartForm buildAudioTranscriptionForm(OpenAiAudioTranscriptionRequest request) {
+        String fileName = request.file().fileName();
+        String mimeType = request.file().mimeType();
+        if (mimeType == null || mimeType.isBlank()) {
+            mimeType = MediaType.APPLICATION_OCTET_STREAM;
+        }
+        ClientMultipartForm form = ClientMultipartForm.create()
+                .binaryFileUpload("file", fileName, Buffer.buffer(Unpooled.wrappedBuffer(request.file().content())),
+                        mimeType)
```
geoand, line 441: This still buffers in memory. You need to use the variant that simply passes the file path, not the content

omatheusmesmo: Hi, @geoand

I'm away from my machine right now, but I want to check something: I was reviewing this with Claude earlier, and another alternative given was to make a temp copy of the file on disk and use that path because langchain4j already uses 'byte[]'. However, you're suggesting another approach as well, right? You really don't need to make a tutorial for me, haha. I just want to confirm this before I dig into it later tonight.

I think I need to debug more carefully to confirm this before sending to you again.

P.S. I'm writing by heart, so maybe I forgot something about it.

geoand: If the request is a file, we should not do anything but pass the file on to `ClientMultipartForm`. Only if the contents are already in memory should a `Buffer` be used.

omatheusmesmo: Hello @geoand ,

It's in-memory only, LangChain4j hand us the audio as a `byte[]` (`Audio` exposes only `binaryData()` / `base64Data()` / `url()`, never a file or path, and `AudioFile.content()` just returns that `byte[]`), so there's no path to give `ClientMultipartForm`.

Confirmed in the debugger:

`QuarkusOpenAiClient.buildAudioTranscriptionForm`
 
screenshot-2026-05-28_21-40-23

`AudioFile.content()`
 
screenshot-2026-05-28_21-39-18

Audio has no path field: `url/base64Data` are null and the content is the in-memory `byte[]`. The buffer wraps that array (Unpooled.wrappedBuffer) instead of copying it, so there's no second copy.

If Audio ever exposes a file/stream source upstream, we can branch to the path-based binaryFileUpload then.

---

https://github.com/langchain4j/langchain4j/blob/b2df847b0fbb9d19c55407683cacc57d9f4e4125/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/audio/transcription/AudioFile.java#L23-L41

https://github.com/langchain4j/langchain4j/blob/main/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/audio/transcription/OpenAiAudioTranscriptionRequest.java

omatheusmesmo: P.S. I also added a javadoc about using the deprecated method, so we can easily migrate when we move to Vert.x 5 .

geoand: Okay, thanks

geoand (approved):
Thanks a lot!

PR #1994, opened 2025-12-04
Add a basic A2A server extension

geoand:
For now, the extension does two things:

- Simplifies the AgentCard creation
- Creates an implementation of AgentExecutor

Compared to how this improves the developer experience for writing an A2A server,  consider the [Weather Agent](https://github.com/a2aproject/a2a-samples/tree/main/samples/java/agents/weather_mcp) from the [A2A samples](https://github.com/a2aproject/a2a-samples).
Using this PR, the following diff shows how much easier the task now becomes (as writing the `AgentCard and the `AgentExecutor` manually, is no longer necessary):

```diff
Index: samples/java/agents/weather_mcp/src/main/java/com/samples/a2a/WeatherAgentCardProducer.java
===================================================================
diff --git a/samples/java/agents/weather_mcp/src/main/java/com/samples/a2a/WeatherAgentCardProducer.java b/samples/java/agents/weather_mcp/src/main/java/com/samples/a2a/WeatherAgentCardProducer.java
deleted file mode 100644
--- a/samples/java/agents/weather_mcp/src/main/java/com/samples/a2a/WeatherAgentCardProducer.java	(revision d4fa006438e521b63a8c8145676f6df0c6b0aafa)
+++ /dev/null	(revision d4fa006438e521b63a8c8145676f6df0c6b0aafa)
@@ -1,68 +0,0 @@
-package com.samples.a2a;
-
-import io.a2a.server.PublicAgentCard;
-import io.a2a.spec.AgentCapabilities;
-import io.a2a.spec.AgentCard;
-import io.a2a.spec.AgentSkill;
-import jakarta.enterprise.context.ApplicationScoped;
-import jakarta.enterprise.inject.Produces;
-import jakarta.inject.Inject;
-import java.util.Collections;
-import java.util.List;
-import org.eclipse.microprofile.config.inject.ConfigProperty;
-
-/**
- * Producer for weather agent card configuration.
- * This class is final and not designed for extension.
- */
-@ApplicationScoped
-public final class WeatherAgentCardProducer {
-
-  /** The HTTP port for the agent service. */
-  @Inject
-  @ConfigProperty(name = "quarkus.http.port")
-  private int httpPort;
-
-  /**
-   * Gets the HTTP port.
-   *
-   * @return the HTTP port
-   */
-  public int getHttpPort() {
-    return httpPort;
-  }
-
-  /**
-   * Produces the agent card for the weather agent.
-   *
-   * @return the configured agent card
-   */
-  @Produces
[...]

PR #2501, opened 2026-05-26
Add A2A Agent Card discovery via Apicurio Registry

### LINKED ISSUE #2335: Feature Proposal: A2A Agent Card discovery via Apicurio Registry
## Summary

I'd like to propose integrating **A2A (Agent-to-Agent) protocol** agent discovery into Quarkus LangChain4j, powered by **[Apicurio Registry](https://www.apicur.io/registry/)**.

Quarkus LangChain4j supports agent patterns and multi-turn tool invocation, but currently lacks a standardized mechanism for agents to discover each other. Apicurio Registry already has native support for the A2A protocol with a dedicated `AGENT_CARD` artifact type and `/.well-known/agent.json` endpoints, including capability-based agent search.

## How it would

1. AI services built with Quarkus LangChain4j **publish** their agent cards to Apicurio Registry on startup (via a CDI startup observer or Quarkus extension auto-configuration).
2. An orchestrator agent **discovers** peer agents by querying the registry's agent search API, filtering by skills, capabilities, or input/output modes.
3. The orchestrator delegates tasks to discovered agents using the contact information from the agent card.

Example agent card registration:
```java
@ApplicationScoped
public class AgentCardPublisher {
    @Inject RegistryClient registryClient;

    void onStart(@Observes StartupEvent ev) {
        registryClient.createArtifact("agents", "support-agent", 
            ArtifactType.AGENT_CARD, agentCardJson);
    }
}
```

Example agent discovery:
```java
@Inject RegistryClient registryClient;

List<AgentCard> translationAgents = registryClient.searchAgents(
    AgentSearchCriteria.builder()
        .skill("translation")
        .inputMode("text")
        .build()
);
```

## Value

- **Multi-agent orchestration** — agents find and delegate to each other through a central registry instead of hard-coded service URLs.
- **Standardized discovery** — follows the [A2A protocol](https://google.github.io/A2A/), ensuring interoperability beyond the Java/Quarkus ecosystem.
- **Governance** — agent card versioning and validation ensure agents advertise accurate capabilities. Compatibility rules catch breaking changes when agent capabilities evolve.
- **Already built on the registry side** — the `AGENT_CARD` artifact type, `/.well-known/agent.json` endpoints, and capability-based search are already implemented in Apicurio Registry.

## Context

The blog post ["Apicurio Registry: AI-Native Evolution"](https://www.apicur.io/blog/2026/02/05/apicurio-registry-ai-natural-evolution) describes a multi-agent context chaining use case where specialized agents (sentiment analysis, issue extraction, response generation, translation) register their capabilities via agent cards and are orchestrated through the registry. The analogy is: just as API schemas need governance, AI agent capabilities need the same level of management.

Phase 2 of the Apicurio AI roadmap includes advanced A2A features like verifiable credentials for federated trust and JSON-LD discovery mechanisms.

## Contribution

I'm happy to implement and contribute this integration if the people see value in it. The registry-side infrastructure already exists — the work would primarily be on the Quarkus LangChain4j side (auto-publishing, discovery APIs, and potentially a `@DiscoverAgent` annotation or similar declarative mechanism).

carlesarnal:
New `a2a-apicurio-registry` extension for discovering and publishing A2A agents through Apicurio Registry, using the `AgentsRegistry` SPI from langchain4j-agentic.

**Discovery:** `ApicurioAgentsRegistry` implements `AgentsRegistry`, queries `AGENT_CARD` artifacts in the registry and creates `AgentInstance` objects via the A2A client builder. Agents can be referenced declaratively with `@RegistryAgent("name")`.

**Publishing:** `@PublishToAgentRegistry` on a `@RegisterAiService` interface auto-publishes the application's agent card on startup via a `@Record(RUNTIME_INIT)` build step.

**Sample:** Support ticket router that classifies requests locally and delegates to specialist agents (billing, technical, shipping) discovered from the registry.

- Closes: #2335


geoand:
Thanks a lot.

Please update the description as per https://github.com/quarkiverse/quarkus-langchain4j/blob/main/AI_POLICY.md#signal-to-noise-ratio

a2a-apicurio-registry/runtime/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-agentic-a2a</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.apicurio</groupId>
+            <artifactId>apicurio-registry-java-sdk</artifactId>
+            <version>3.2.1</version>
```
geoand: Let's create a property for this version and move it into the base `pom.xml`

a2a-apicurio-registry/runtime/src/main/resources/META-INF/beans.xml
geoand, line 1: This file should be completely empty

a2a-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/a2a/runtime/apicurio/ApicurioRegistryA2ATools.java
geoand, line 1: Can you outline a scenario where the tools in this class would be useful?

cc @mariofusco @kabir

a2a-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/a2a/runtime/apicurio/config/A2AApicurioRegistryRuntimeConfig.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Optional;
+
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithDefault;
+
+@ConfigRoot(phase = ConfigPhase.RUN_TIME)
+@ConfigMapping(prefix = "quarkus.langchain4j.a2a.apicurio-registry")
+public interface A2AApicurioRegistryRuntimeConfig {
```
geoand, line 12: Wouldn't it make sense for some of these properties to be defined in an annotation?

I had done something similar in https://github.com/quarkiverse/quarkus-langchain4j/pull/1994

a2a-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/a2a/runtime/apicurio/A2AAgentCardPublisher.java
```diff
@@ -0,0 +1,116 @@
...
+
+            registryClient.groups().byGroupId(groupId).artifacts().post(createArtifact);
+
+            log.infof("Published agent card '%s' to Apicurio Registry (group=%s, artifact=%s)",
+                    agentName, groupId, artifactId);
+        } catch (JsonProcessingException e) {
+            log.errorf(e, "Failed to serialize agent card for '%s'", agentName);
+        } catch (Exception e) {
+            log.errorf(e, "Failed to publish agent card '%s' to Apicurio Registry", agentName);
+        }
```
geoand: Instead of a giant try-catch, can't we make this more focused?

geoand (commented):
Thanks for this.

I have added various comments

carlesarnal:
Re: the use-case for `ApicurioRegistryA2ATools` —

An orchestrator agent has an AI service with these tools available. When a user asks "translate this to French," the LLM calls `searchA2AAgents("translation")` to find agents in the registry that have translation skills, then `delegateToA2AAgent(groupId, artifactId, "translate this to French")` to send the task to the discovered agent.

The key value is that the orchestrator does not need hardcoded agent URLs — it discovers peers dynamically through the registry. Agents can be added, removed, or updated independently.

geoand:
I can't say I'm convinced, but I'd like to hear from @mariofusco as well

a2a-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/a2a/runtime/apicurio/ApicurioRegistryA2ATools.java
mariofusco: I believe that the idea here would be to provide a search discovery tool, similar to what we have for skills. If so, the problem is that we are not discovering skills or other tools, but A2A-based remote agents, so what can you do with them once they have been discovered? One possibility in this instance would be to use the agent-as-tool pattern and make the newly discovered agents available to the main agent as tools. However I must admit that I never particularly liked that pattern and in this situation is seems even more stretched. 

More in detail, I believe that the agent-as-tool pattern is a simple agentic system in disguise, so why not using this new feature to create a proper agentic system allowing to dynamically discover and register other agents while running? I believe that this idea could work well with the supervisor pattern for example, but this would require some improvement also on the langchain4j-agentic side.

In essence what I'm proposing is to introduce in langchain4j-agentic an `AgentsRegistry` SPI, check how this works and can be integrated with some agentic patterns like the supervisor one, and finally leverage the features provided in this new module to provide an Apicurio-based implementation of that SPI. 

WDYT? /cc @cescoffier @dliubarskyi

carlesarnal: Makes sense. I went with the `@Tool` approach because it was the most obvious path for me, but I see how it's just wrapping an agentic system in tool calls. The AgentsRegistry SPI in langchain4j-agentic with an Apicurio implementation sounds like a much cleaner fit.

Happy to help with that once the SPI shape is clearer or I can try to draft it. Waiting on input from others in the meantime.

a2a-apicurio-registry/deployment/src/main/java/io/quarkiverse/langchain4j/a2a/deployment/apicurio/A2AApicurioRegistryProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+    public void registerApicurioRegistryA2ATools(
+            A2AApicurioRegistryBuildTimeConfig buildTimeConfig,
+            BuildProducer<SyntheticBeanBuildItem> beanProducer,
+            CoreVertxBuildItem vertxBuildItem,
+            A2AApicurioRegistryRecorder recorder) {
+        if (!buildTimeConfig.enabled()) {
+            return;
+        }
+
+        log.info("Apicurio Registry A2A integration enabled, registering ApicurioRegistryA2ATools bean");
```
geoand: This should be a debug log

a2a-apicurio-registry/deployment/src/main/java/io/quarkiverse/langchain4j/a2a/deployment/apicurio/A2AApicurioRegistryProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+        if (skillsValue != null) {
+            for (AnnotationInstance skillAnnotation : skillsValue.asNestedArray()) {
+                skills.add(new A2AAgentCardPublisher.SkillInfo(
+                        skillAnnotation.value("id").asString(),
+                        skillAnnotation.value("name").asString(),
+                        skillAnnotation.value("description").asString()));
+            }
+        }
+
+        log.infof("Found @PublishToAgentRegistry on %s, will publish agent card '%s'",
```
geoand: Same here

a2a-apicurio-registry/deployment/src/main/java/io/quarkiverse/langchain4j/a2a/deployment/apicurio/A2AApicurioRegistryProcessor.java
```diff
@@ -0,0 +1,150 @@
...
+
+        reflectiveClass.produce(ReflectiveClassBuildItem.builder(
+                "io.apicurio.registry.rest.client.models.ArtifactMetaData",
+                "io.apicurio.registry.rest.client.models.SearchedArtifact",
+                "io.apicurio.registry.rest.client.models.ArtifactSearchResults",
+                "io.apicurio.registry.rest.client.models.CreateArtifact",
+                "io.apicurio.registry.rest.client.models.CreateVersion",
+                "io.apicurio.registry.rest.client.models.VersionContent",
+                "io.apicurio.registry.rest.client.models.Labels")
+                .fields(true).methods(true).build());
```
geoand: Why are these needed?

carlesarnal: They were more preventive than anything else based on my experience. I've removed them and I'll them only if needed.

carlesarnal:
I have updated this now to build on the SPI raised on https://github.com/langchain4j/langchain4j/pull/5491.

a2a-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/a2a/runtime/apicurio/ApicurioAgentsRegistry.java
```diff
@@ -0,0 +1,82 @@
...
+                    if (agentUrl == null) {
+                        continue;
+                    }
+
+                    String name = artifact.getName() != null ? artifact.getName() : artifact.getArtifactId();
+                    UntypedAgent a2aAgent = AgenticServices.a2aBuilder(agentUrl)
+                            .outputKey(name)
+                            .build();
+
+                    if (a2aAgent instanceof AgentInstance instance) {
```
mariofusco: Is this `instanceof` necessary? Did you find situation where the `a2aAgent` was not an `AgentInstance`? I believe this should be pretty impossible, but even if so we should fail fast instead of hiding the problem and move forward.

carlesarnal:
This is waiting on a release of langchain4j.

mariofusco:
> This is waiting on a release of langchain4j.

LangChain4j 1.17.2 already contains the new `AgentsRegistry` and the quarkus extension has been already bumped to that release. Maybe you just need to rebase this pull request?

carlesarnal:
@mariofusco @geoand this is ready for review. Unfortunately, I think I'm facing the same problem with the documentation as in https://github.com/quarkiverse/quarkus-langchain4j/pull/2338/changes.

geoand:
I force pushed to fix the docs

a2a-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/a2a/runtime/apicurio/config/A2AApicurioRegistryRuntimeConfig.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * The base URL of the Apicurio Registry instance
+     * (e.g. http://localhost:8080/apis/registry/v3).
+     */
+    Optional<String> url();
+
+    /**
+     * Optional authentication token for the registry.
+     */
+    Optional<String> authToken();
```
mariofusco, line 23: I don't see where this is used. For instance, shouldn't it be passed to the `RegistryClientOptions` in `A2AApicurioRegistryRecorder.createRegistryClient()`?

a2a-apicurio-registry/deployment/src/main/java/io/quarkiverse/langchain4j/a2a/deployment/apicurio/A2AApicurioRegistryProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+        log.debugf("Found @PublishToAgentRegistry on %s, will publish agent card '%s'",
+                annotation.target(), agentName);
+
+        beanProducer.produce(SyntheticBeanBuildItem
+                .configure(A2A_PUBLISHER)
+                .setRuntimeInit()
+                .defaultBean()
+                .unremovable()
+                .scope(ApplicationScoped.class)
+                .createWith(recorder.agentCardPublisherFunction(
```
mariofusco: I have the feeling that this function and the subsequent call to `recorder.publishAgentCard` at line 113 create a duplicated `A2AAgentCardPublisher`. I believe that this twofold creation is wasteful and in reality that the synthetic bean is created but never actually used (it's unremovable, so it stays around).

a2a-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/a2a/runtime/apicurio/ApicurioAgentsRegistry.java
```diff
@@ -0,0 +1,97 @@
...
+    public ApicurioAgentsRegistry(RegistryClient registryClient) {
+        this.registryClient = registryClient;
+    }
+
+    @Override
+    public Map<String, AgentInstance> allAgents() {
+        if (agents == null) {
+            synchronized (this) {
+                if (agents == null) {
+                    agents = discoverAgents();
```
mariofusco: I'm not sure if this is wanted but, if the registry is temporarily unreachable at the first call, `discoverAgents()` returns an empty Map anyway, preventing any other subsequent re-discover. Also shouldn't there be a mechanism to cleanup the existing agents and force a re-discovery?

a2a-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/a2a/runtime/apicurio/ApicurioAgentsRegistry.java
```diff
@@ -0,0 +1,97 @@
...
+        return agent;
+    }
+
+    private Map<String, AgentInstance> discoverAgents() {
+        Map<String, AgentInstance> discovered = new HashMap<>();
+
+        try {
+            var results = registryClient.search().artifacts().get(config -> {
+                config.queryParameters.artifactType = AGENT_CARD_TYPE;
+                config.queryParameters.limit = 100;
```
mariofusco: This hardcodes the maximum number of discovered agents to 100, silently dropping any further agents. Could we make this limit configurable or at least emit a warning if some agents are dropped from the lookup?

a2a-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/a2a/runtime/apicurio/A2AApicurioRegistryRecorder.java
```diff
@@ -0,0 +1,81 @@
...
+public class A2AApicurioRegistryRecorder {
+
+    private final RuntimeValue<A2AApicurioRegistryRuntimeConfig> runtimeConfig;
+
+    public A2AApicurioRegistryRecorder(RuntimeValue<A2AApicurioRegistryRuntimeConfig> runtimeConfig) {
+        this.runtimeConfig = runtimeConfig;
+    }
+
+    public Function<SyntheticCreationalContext<ApicurioAgentsRegistry>, ApicurioAgentsRegistry> agentsRegistryFunction(
+            Supplier<io.vertx.core.Vertx> vertx) {
```
mariofusco: I'm not getting why you are passing around this Supplier since you never use it.

a2a-apicurio-registry/runtime/src/main/java/io/quarkiverse/langchain4j/a2a/runtime/apicurio/A2AApicurioRegistryRecorder.java
```diff
@@ -0,0 +1,81 @@
...
+            public ApicurioAgentsRegistry apply(SyntheticCreationalContext<ApicurioAgentsRegistry> context) {
+                A2AApicurioRegistryRuntimeConfig config = runtimeConfig.getValue();
+                RegistryClient registryClient = createRegistryClient(config);
+                return new ApicurioAgentsRegistry(registryClient);
+            }
+        };
+    }
+
+    public Function<SyntheticCreationalContext<A2AAgentCardPublisher>, A2AAgentCardPublisher> agentCardPublisherFunction(
+            Supplier<io.vertx.core.Vertx> vertx,
```
mariofusco: Same here.

a2a-apicurio-registry/deployment/src/main/java/io/quarkiverse/langchain4j/a2a/deployment/apicurio/A2AApicurioRegistryProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+            return;
+        }
+
+        IndexView index = combinedIndex.getIndex();
+        Collection<AnnotationInstance> annotations = index.getAnnotations(PUBLISH_ANNOTATION);
+        if (annotations.isEmpty()) {
+            return;
+        }
+
+        AnnotationInstance annotation = annotations.iterator().next();
```
mariofusco: This implies that only the first `@PublishToAgentRegistry` annotation is processed, so if multiple AiServices have this annotation, only the first one is published. I believe that this should either be documented as "only one per application" (and fail-fast if more are found), or support multiple.

mariofusco (changes requested):
Please give a look at my comments. 

Also I see an example of how the new `@PublishToAgentRegistry` annotation is integrated with `@RegisterAiService`, but maybe it could make sense to provide and test the same level of integration for a whole agentic system, publishing to the registry its top level agent.

mariofusco (approved):
It seems that a complete end-to-end example demonstrating how the new `@PublishToAgentRegistry` annotation can be used and integrated with an agentic system, publishing to the registry its top level agent is still missing and I would like to see it at work before merging this pull request. Anyway the implementation looks ok to me now.

geoand (approved):

carlesarnal:
> It seems that a complete end-to-end example demonstrating how the new `@PublishToAgentRegistry` annotation can be used and integrated with an agentic system, publishing to the registry its top level agent is still missing and I would like to see it at work before merging this pull request. Anyway the implementation looks ok to me now.

Ok, it's been merged, but I'll add it as a follow-up.

Issue #2400, opened 2026-04-28
Mermaid diagrams render as raw source on docs.quarkiverse.io (kroki-error)


rolfedh:
## Summary

Every `[mermaid]` block on the published `dev` docs renders as raw `flowchart …` source instead of as an image. The HTML wraps it in `<div class="listingblock kroki-error">`, which `asciidoctor-kroki` emits as a fallback when the diagram conversion call fails.

Reproducing locally and probing `kroki.io` directly shows the failure originates in `kroki.io`, not in this repo — see "Root cause" below.

**Live example:** https://docs.quarkiverse.io/quarkus-langchain4j/dev/models.html#_orchestration_through_ai_services — the diagram under "Orchestration through AI services" appears as plain text.

The HTML around that block:

```html
<div class="listingblock kroki-error">
  <div class="content">
    <pre>flowchart LR
    App[Quarkus Application]
    AIS[AI Service]
    ...
    </pre>
  </div>
</div>
```

## Affected pages

Same symptom on every `dev` page that contains `[mermaid]`:

- [models.html](https://docs.quarkiverse.io/quarkus-langchain4j/dev/models.html#_orchestration_through_ai_services)
- [rag.html](https://docs.quarkiverse.io/quarkus-langchain4j/dev/rag.html)
- [quickstart-rag.html](https://docs.quarkiverse.io/quarkus-langchain4j/dev/quickstart-rag.html)
- [rag-ingestion.html](https://docs.quarkiverse.io/quarkus-langchain4j/dev/rag-ingestion.html)
- [rag-query.html](https://docs.quarkiverse.io/quarkus-langchain4j/dev/rag-query.html)
- [testing.html](https://docs.quarkiverse.io/quarkus-langchain4j/dev/testing.html)

In every case the rendered page contains the diagram source as a `<pre>` block and no corresponding `<img>` tag.

## Reproduction

Confirmed locally by building the docs from a clean checkout of `main`:

```bash
git clone https://github.com/quarkiverse/quarkus-langchain4j
cd quarkus-langchain4j/docs
npm ci
npx antora --stacktrace antora-playbook.yml
grep -c kroki-error target/site/quarkus-langchain4j/dev/models.html
# → 1
```

The Antora log shows `asciidoctor-kroki` calling `kroki.io` and receiving HTTP 504 after ~60 seconds for every Mermaid block:

```
[...]

PR #2503, opened 2026-05-26
docs(rag): render RAG flow diagram from committed SVG

rolfedh:
Replace the [mermaid, RAG flow, svg] block in rag.adoc with an image::rag-flow.svg macro pointing at a pre-rendered SVG committed under modules/ROOT/assets/images/. The mermaid source moves to diagrams/rag-flow.mmd for future regeneration via `npx mmdc`.

This removes the build-time kroki dependency for this page so it cannot regress to a kroki-error fallback when kroki.io is degraded.

First page of six; the other five mermaid blocks (in models.adoc, quickstart-rag.adoc, rag-ingestion.adoc, rag-query.adoc, and testing.adoc) can follow the same pattern in a follow-up once the maintainers confirm the approach.

Relates to #2400.


geoand (approved):
Thanks!

PR #2505, opened 2026-05-27
Decouple Ollama streaming from Mutiny Context binary signature

omatheusmesmo:
## Describe your changes

`OllamaStreamingChatLanguageModel.doChat` used `io.smallrye.mutiny.Context` only as a state holder across the streaming callbacks. Mutiny 3.2.0 (Quarkus 3.36) changed `Context.put`/`get` from `(String, Object)` to `(Object, Object)` and removed the old overloads. The extension is compiled against Quarkus 3.33 (Mutiny 3.1.1), so its bytecode calls a method that is gone on 3.36: a native build fails at analysis time (`unresolved method ... Context.put`), while in JVM mode it would only show up as a `NoSuchMethodError` if the streaming path runs.

Replaced the `Context` with plain locals and the context-free `MultiSubscribe.with(...)` overload, so the bytecode no longer depends on the Mutiny `Context` signature. Behavior is unchanged.

---

- Reported via the quarkus-flow nightly native build (Quarkus 3.36, quarkus-langchain4j 1.10.0) on Zulip: [#dev > Quarkus Langchain4j Native Build errors](https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/Quarkus.20Langchain4j.20Native.20Build.20errors/with/597932036) .

---

P.S. I choose to not update the Quarkus Version of the project to not deal with all changes needed for it, but perhaps we need to take a look into it further.

### DIFF
--- a/model-providers/ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/OllamaStreamingChatLanguageModel.java
+++ b/model-providers/ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/OllamaStreamingChatLanguageModel.java
@@ -10,6 +10,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
@@ -29,7 +30,6 @@ import dev.langchain4j.model.chat.request.ChatRequestParameters;
 import dev.langchain4j.model.chat.response.ChatResponseMetadata;
 import dev.langchain4j.model.chat.response.StreamingChatResponseHandler;
 import dev.langchain4j.model.output.TokenUsage;
-import io.smallrye.mutiny.Context;
 
 /**
  * Use to have streaming feature on models used trough Ollama.
@@ -38,10 +38,6 @@ public class OllamaStreamingChatLanguageModel implements StreamingChatModel {
 
     private static final Logger log = Logger.getLogger(OllamaStreamingChatLanguageModel.class);
 
-    private static final String TOOLS_CONTEXT = "TOOLS";
-    private static final String TOKEN_USAGE_CONTEXT = "TOKEN_USAGE";
-    private static final String RESPONSE_CONTEXT = "RESPONSE";
-    private static final String MODEL_ID = "MODEL_ID";
     private final OllamaClient client;
     private final String model;
     private final String format;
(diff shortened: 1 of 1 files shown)


geoand:
I am not sure this is the proper solution to the issue raised in https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/Quarkus.20Langchain4j.20Native.20Build.20errors/near/597931670

cescoffier:
Yeah, I would like to hear @jponge as 1) I'm not sure of the solution, 2) I'm not totally sure of the root cause either (breaking a mutiny API requires a justification, I'm surprised revapi did not report it).

geoand:
As far as I can tell, there is no breakage, which is why I asked on the original issue for the exact way to reproduce the issue

jponge:
First off sorry if it did break code.

This was a conscious change in Mutiny (RevAPI had been told a justification), going from `String` to `Object` being quite safe given type erasure rules (you can still use `String` keys and the code will compile fine, and at runtime types are erased anyway).

geoand:
I guess my IDE was tricking me then into thinking there was no breakage.

@jponge the code in question is:

https://github.com/quarkiverse/quarkus-langchain4j/blob/04c0ae077aed091431b6a517a74bc22dddc7bb07/model-providers/ollama/runtime/src/main/java/io/quarkiverse/langchain4j/ollama/OllamaStreamingChatLanguageModel.java#L80-L83

How would that be rewritten?

jponge:
It doesn't have to be modified, a `String` is a `Object`, see https://github.com/smallrye/smallrye-mutiny/commit/7ee9346722be0d7111be9ce546438a6b0b469a29 for the upstream change.

geoand:
Okay, that's what I hoped, but we still need to figure out (with @ricardozanini) why Quarkus Flow is failing

geoand:
I was able to reproduce the original issue, so I'm going to check and see what can be done about it

omatheusmesmo:
I think the reason Flow hits it: it consumes the published quarkus-langchain4j-ollama:1.10.0 binary, and its bytecode calls Context.put(String, Object), that descriptor is gone in Mutiny 3.2.0 (only put(Object, Object) now), so it's a NoSuchMethodError at link time, which native-image reports at build time.

Recompiling against 3.2.0 fixes it (a String is an Object), but a release cut from main today would still ship the old bytecode, since main is on Quarkus 3.33 / Mutiny 3.1.1.

So either this PR (drops Context, works on any Mutiny version) or a Quarkus 3.36 bump on our side would fix it. I'm happy to follow either way.

geoand:
Actually, what happens is straightforward: The change in mutiny is source compatible (i.e. Quarkus LangChain4j sources do not need any changes to compile against Mutiny 3.2), but it's not binary compatible as  

```
invokevirtual #141                // Method io/smallrye/mutiny/Context.put:(Ljava/lang/String;Ljava/lang/Object;)Lio/smallrye/mutiny/Context;
```

is no longer resolvable.

My suggestion is that the String variants be added back to Mutiny so existing code can be forward compatible =, WDYT @jponge @cescoffier ?

jponge:
I could do a 3.2.1 with those methods added as deprecated.

geoand:
That would be great, thanks a lot!

omatheusmesmo:
I think this PR is no longer necessary then, right? May I just close it?

geoand:
Yeah, let's close. Thanks for looking into the issue nonetheless!

PR #2507, opened 2026-05-27
docs: render remaining mermaid diagrams from committed SVGs

### LINKED ISSUE #2400: Mermaid diagrams render as raw source on docs.quarkiverse.io (kroki-error)

rolfedh:
Follow-up to #2503. Replaces the five remaining `[mermaid, ...]` blocks in models.adoc, quickstart-rag.adoc, rag-ingestion.adoc, rag-query.adoc, and testing.adoc with `image::` macros pointing at pre-rendered SVGs under `modules/ROOT/assets/images/`. The mermaid sources live in `diagrams/*.mmd`.

Since the last push, this PR also:

* **Fixes the diagram display widths** that kept it in draft. Root cause: mmdc's default output sets `width="100%"` on the SVG root, so when the file is referenced via `` the browser has no intrinsic width and stretches it to the full docs column, making narrow diagrams disproportionately large. All SVGs are now rendered with `useMaxWidth: false` (`diagrams/mermaid-config.json`), which emits explicit pixel width/height. Narrow diagrams render at natural size, and the site stylesheet's `max-width: 100%` still scales down diagrams wider than the column. This includes a regenerated `rag-flow.svg` from #2503, which had the same latent issue.
* **Adds `npm run diagrams`** (`diagrams/generate.sh`), so future diagram edits regenerate every SVG with the correct settings in one command. The script also stamps each SVG with a "generated file, do not edit" comment, the `.mmd` headers point at the command, and the README documents the workflow.
* **Removes `asciidoctor-kroki`** from the Antora playbook and package.json. With no `[mermaid]` blocks left, the docs build no longer contacts kroki.io at all, so it cannot regress to the kroki-error fallback when kroki.io is degraded.

Verified with a full local Antora build: exit 0, zero `kroki-error` divs, and the built pages render the diagrams at natural size.

* Fixes #2400


rolfedh:
I've set this as a Draft PR while I investigate and fix issues with the display widths of the diagrams.

geoand:
🆗

rolfedh:
@geoand The display-width issue is resolved, so this is ready for review. Root cause: mmdc's default output sets `width="100%"` on the SVG root, which makes browsers stretch narrow diagrams to the full docs column. The SVGs are now rendered with explicit pixel dimensions and display at natural size.

The PR also adds `npm run diagrams` so future diagram edits regenerate the SVGs the same way, and removes `asciidoctor-kroki` from the build since no `[mermaid]` blocks remain. That takes kroki.io out of the picture entirely, so merging this fully closes #2400.

geoand (approved):
Thanks!

Issue #2396, opened 2026-04-25
Align max sequential tool invocations naming with upstream API


omatheusmesmo:
The naming for the max sequential tool invocations/execution concept is inconsistent across the Quarkus integration:

| Layer | Current name | User-facing? |
|---|---|---|
| Upstream API | `maxSequentialToolsInvocations` | Yes (builder method) |
| Annotation attribute | `maxSequentialToolInvocations` | Yes |
| Config property | `quarkus.langchain4j.ai-service.max-tool-executions` | Yes |
| Internal record field | `maxSequentialToolInvocations` | No |
| Internal runtime field | `maxSequentialToolExecutions` | No |

The builder method was recently fixed in #2393 to properly override the upstream `maxSequentialToolsInvocations(int)`, and the old non-overriding `maxSequentialToolInvocations(Integer)` method was removed. This issue was initially discussed there.

However, the annotation attribute and config property still use different names from the upstream API, which is confusing for users and was the root cause of #2373.

A possible approach would be to align everything with the upstream naming (`maxSequentialToolsInvocations`), deprecating the old names where they are user-facing (annotation attribute and config property) and renaming the internal fields outright.

Open questions:
- From which version should the old names be deprecated?
- How many versions should pass before the deprecated names are removed?
- Should the config property also be renamed to `max-sequential-tools-invocations` to match the upstream naming, or is `max-tool-executions` acceptable as-is?

jmartisk:
I guess I wouldn't mind unifying to `maxSequentialToolInvocations` (better than `maxSequentialToolsInvocations`) but that would mean also renaming it in the upstream API, so we'd have to discuss it there too :)

Generally, deprecation for ~4-6 months would be enough? I don't think we have any very specific policy for that.

omatheusmesmo:
> I guess I wouldn't mind unifying to `maxSequentialToolInvocations` (better than `maxSequentialToolsInvocations`) but that would mean also renaming it in the upstream API, so we'd have to discuss it there too :)
> 
> Generally, deprecation for ~4-6 months would be enough? I don't think we have any very specific policy for that.

Issue opened, @jmartisk ! 😃 

https://github.com/langchain4j/langchain4j/issues/5063

omatheusmesmo:
Completed on #2509

PR #2393, opened 2026-04-25
Make QuarkusAiServices override upstream maxSequentialToolsInvocations method

omatheusmesmo:
The QuarkusAiServices builder defined `maxSequentialToolInvocations(Integer)` which does not override the upstream `AiServices#maxSequentialToolsInvocations(int)`. This caused the value set via the programmatic builder to be ignored at runtime, always falling back to the default of 10.

This PR adds the missing `@Override` and removes the old non-overriding method.

Note: the naming across the integration is inconsistent (annotation attribute, config property, internal field — all use different names for the same concept). This PR only fixes the immediate bug. A broader unification would be a breaking change and is better discussed separately.

- Fixes: #2373

PR #2509, opened 2026-05-27
refactor: align max sequential tool invocations naming with upstream API maxToolCallingRoundTrips

JeffersonSantos29:
## Related Issues
* **Resolves:** #2396
* **Related to:** #2393, [Upstream Issue #5063](https://github.com/langchain4j/langchain4j/issues/5063), [Upstream PR #5148](https://github.com/langchain4j/langchain4j/pull/5148)

## Change
Renames `maxSequentialToolInvocations` (and its configuration counterparts) to `maxToolCallingRoundTrips` across the Quarkus extension (`RegisterAiService`, `AiServiceConfig`, `AiServicesProcessor`, and internal runtime components).

Following the upstream library changes, the old name was semantically misleading — it limits LLM request/response round trips, not the total number of individual tool invocations. This PR permanently fixes the naming fragmentation across the annotation, configuration, and internal layers, aligning them perfectly with the upstream API.

Old methods and config properties (like `max-tool-executions`) are kept as `@Deprecated(forRemoval = true)` and act as safe fallbacks. Existing applications will continue to compile and work without changes.

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -517,9 +517,16 @@ public class AiServicesProcessor {
                 throw new IllegalArgumentException("Tool usage requires chat memory. Offending AiService is '"
                         + declarativeAiServiceClassInfo.name() + "'");
             }
-            Integer maxSequentialToolInvocations = instance.value("maxSequentialToolInvocations") != null
-                    ? instance.value("maxSequentialToolInvocations").asInt()
-                    : 0;
+            Integer maxToolCallingRoundTrips = 0;
+            AnnotationValue newAnnoValue = instance.value("maxToolCallingRoundTrips");
+            AnnotationValue legacyAnnoValue = instance.value("maxSequentialToolInvocations");
+
+            if (newAnnoValue != null) {
+                maxToolCallingRoundTrips = newAnnoValue.asInt();
+            } else if (legacyAnnoValue != null) {
+                // Fallback to deprecated property
+                maxToolCallingRoundTrips = legacyAnnoValue.asInt();
+            }
 
             Integer maxToolCallsPerResponse = instance.value("maxToolCallsPerResponse") != null
                     ? instance.value("maxToolCallsPerResponse").asInt()
@@ -584,7 +591,7 @@ public class AiServicesProcessor {
                             classOutputGuardrails(declarativeAiServiceClassInfo, index),
                             toolArgumentsErrorHandlerDotName(declarativeAiServiceClassInfo, generatedBeanProducer),
                             toolExecutionErrorHandlerDotName(declarativeAiServiceClassInfo, generatedBeanProducer),
-                            maxSequentialToolInvocations,
+                            maxToolCallingRoundTrips,
                             maxToolCallsPerResponse,
                             allowContinuousForcedToolCalling,
                             // we need to make these @DefaultBean because there could be other CDI beans of the same type that need to take precedence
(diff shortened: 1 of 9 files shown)


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/config/AiServiceConfig.java
```diff
@@ -18,7 +20,15 @@ public interface AiServiceConfig {
     /**
      * Maximum number of sequential tool executions while handling a single chat request.
      * If this number is exceeded, the chat request will fail.
+     *
+     * @deprecated Use {@link #maxToolCallingRoundTrips()} instead.
+     */
+    @Deprecated(forRemoval = true, since = "999-SNAPSHOT")
```
geoand: `since` is wrong here; it should be `1.11.0`

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/config/AiServiceConfig.java
```diff
@@ -18,7 +20,15 @@ public interface AiServiceConfig {
...
+    Optional<Integer> maxToolExecutions();
+
+    /**
+     * Maximum number of LLM request/response round trips while handling a single chat request.
+     * If this number is exceeded, the chat request will fail.
      */
-    @WithDefault("10")
```
geoand, line 22: Why was this removed?

core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
```diff
@@ -90,9 +90,22 @@
      * If not specified (left to zero) for a specific AI service,
      * the AI service will use the value of the common {@code quarkus.langchain4j.ai-service.max-tool-executions} property.
      * If that property is unset too, the default is 10 invocations.
+     *
+     * @deprecated Use {@link #maxToolCallingRoundTrips()} instead.
      */
+    @Deprecated(forRemoval = true, since = "999-SNAPSHOT")
```
geoand: Same comment as above for `since`

geoand (commented):
Thanks for this.

I've added some comments.

Also cc @jmartisk

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/config/AiServiceConfig.java
jmartisk: I guess the new property (`maxToolCallingRoundTrips`) should have a default of 10. But then it also shouldn't be an `Optional`, no?

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -517,9 +517,16 @@ public void findDeclarativeServices(CombinedIndexBuildItem indexBuildItem,
                 throw new IllegalArgumentException("Tool usage requires chat memory. Offending AiService is '"
                         + declarativeAiServiceClassInfo.name() + "'");
             }
-            Integer maxSequentialToolInvocations = instance.value("maxSequentialToolInvocations") != null
-                    ? instance.value("maxSequentialToolInvocations").asInt()
-                    : 0;
+            Integer maxSequentialToolInvocations = 0;
```
jmartisk: should we also rename this variable to `maxToolCallingRoundTrips`?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/config/AiServiceConfig.java
JeffersonSantos29: Fixed! Updated to 1.11.0.

core/runtime/src/main/java/io/quarkiverse/langchain4j/RegisterAiService.java
JeffersonSantos29: Fixed! Updated to 1.11.0.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/config/AiServiceConfig.java
JeffersonSantos29: Good catch, @jmartisk!  I've reverted both properties back to int with @WithDefault("10") to maintain the configuration contract.

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
JeffersonSantos29: Makes sense! Renamed to maxToolCallingRoundTrips.

jmartisk (approved):

PR #2515, opened 2026-05-29
Chatscopes CDI lifecycle events and route exception handling

patriot1burke:
Added two new features to Chat Scopes:

* CDI Lifecycle events for start, end, activate and deactivate of a Chat Scope.
* Default and route-based Exception handlers that can be set up via annotations.

Docs have more details.


chat-scopes/core/runtime/src/main/java/io/quarkiverse/langchain4j/chatscopes/ChatScopeCDIEvent.java
```diff
@@ -0,0 +1,14 @@
+package io.quarkiverse.langchain4j.chatscopes;
+
+/**
+ * Various classes implement this to send chat scope lifecycle events like
+ * chat scope starting, ending, etc.
+ */
+public interface ChatScopeCDIEvent {
```
edeandrea, line 7: Does it have to have `CDI` in the name of the class? The fact that its CDI is more an implementation detail.

chat-scopes/core/runtime/src/main/java/io/quarkiverse/langchain4j/chatscopes/AbstractChatScopeCDIEvent.java
```diff
@@ -0,0 +1,14 @@
+package io.quarkiverse.langchain4j.chatscopes;
+
+public abstract class AbstractChatScopeCDIEvent implements ChatScopeCDIEvent {
```
edeandrea, line 3: Does it have to have `CDI` in the name of the class? The fact that its CDI is more an implementation detail.

patriot1burke: Well, chat scopes is a CDI-based framework.  And I don't want somebody getting confused between these CDI events and chat scope client events which are a different thing.

chat-scopes/core/deployment/src/main/java/io/quarkiverse/langchain4j/chatscopes/deployment/ChatScopeProcessor.java
```diff
@@ -4,10 +4,7 @@
 import static io.quarkus.deployment.annotations.ExecutionTime.STATIC_INIT;
 
 import java.lang.reflect.Modifier;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
geoand, line 7: Let's try and avoid these next time. 

I know the IDE adds them automatically, so it needs some tweaking to avoid it

patriot1burke: @geoand ugh.  Recently went back to a fresh Intellij.  I hate .* too.

Ladicek:
I don't know anything about ChatScopes, but if there's anything in here that is actual CDI scope/context, I would encourage you to follow https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1.html#custom_contexts

PR #2516, opened 2026-05-30
Add audio transcription support to the Azure OpenAI extension

### LINKED ISSUE #2514: Add audio transcription support to the Azure OpenAI extension
## Description

#2497 added `AudioTranscriptionModel` (Whisper) as an injectable CDI bean to the `quarkus-langchain4j-openai` extension. The same capability is missing from the Azure OpenAI extension.

Upstream langchain4j already provides the implementation: `AzureOpenAiAudioTranscriptionModel` (in `langchain4j-azure-open-ai`, currently `@Experimental`). The `quarkus-langchain4j-openai-azure` extension wraps Azure's Chat, StreamingChat, Embedding and Image models, but not transcription, so `@Inject AudioTranscriptionModel` does not resolve for an Azure-configured application.

This was anticipated in #1502, where it was noted that Azure OpenAI exposes essentially the same transcription API as vanilla OpenAI.

## Scope

The quarkiverse Azure models are not built on the langchain4j Azure SDK; they reuse `QuarkusOpenAiClient` from `openai-common` (see `AzureOpenAiImageModel`, which calls `client.imagesGeneration(...)`). Since #2497 already added `audioTranscription(...)` to `QuarkusOpenAiClient`, the transcription call (including the multipart handling) is reused as-is, and the work is mostly config and wiring:

- Add a quarkiverse `AzureOpenAiAudioTranscriptionModel` implementing `dev.langchain4j.model.audio.AudioTranscriptionModel`, built on `QuarkusOpenAiClient.audioTranscription(...)`, mirroring `AzureOpenAiImageModel`.
- Add `AudioTranscriptionModelConfig` and reference it from `LangChain4jAzureOpenAiConfig`, plus add `AUDIO_TRANSCRIPTION` to the `EndpointType` enum and to the per-type resolution methods (`endPointFor`, `resourceNameFor`, `domainNameFor`, `deploymentNameFor`).
- Add an `audioTranscriptionModel(...)` method to `AzureOpenAiRecorder`, the same way `imageModel(...)` is wired (endpoint via `getEndpoint(..., EndpointType.AUDIO_TRANSCRIPTION)`).
- Produce the `SyntheticBeanBuildItem` for `AudioTranscriptionModel` in `AzureOpenAiProcessor`, reusing the candidate/selected build items already introduced in core by #2497.
- Add a `DisabledAudioTranscriptionModel` fallback and a deployment test, following the OpenAI module.
- Document the new config keys and usage, mirroring `openai-audio-transcription-model.adoc`.

## Notes

- The core wiring (`AudioTranscriptionModelProviderCandidateBuildItem`, `SelectedAudioTranscriptionModelProviderBuildItem`, the `AudioTranscriptionModel` selection in `BeansProcessor`) already exists from #2497, so this is provider-side work only.
- Transcription against Azure already works today by pointing `quarkus.langchain4j.openai.base-url` at the Azure endpoint (with `api-version` and `api-key`), since `QuarkusOpenAiClient` already carries the Azure auth and api-version. So this is not about adding a missing capability, but about exposing it through the idiomatic `quarkus.langchain4j.azure-openai.*` namespace (resource/domain/deployment name resolution, per-model enable flag), consistent with the other Azure models.
- OpenAI and Azure OpenAI are currently the only providers with an upstream `AudioTranscriptionModel`, so no other provider is in scope.

omatheusmesmo:
## Describe your changes

Exposes `AudioTranscriptionModel` (Whisper) as an injectable CDI bean in the `quarkus-langchain4j-openai-azure` extension.

Like the other Azure models (`AzureOpenAiChatModel`, `AzureOpenAiImageModel`, ...), the new `AzureOpenAiAudioTranscriptionModel` is built on `QuarkusOpenAiClient` rather than the langchain4j Azure SDK. Since #2497 already added `audioTranscription(...)` to `QuarkusOpenAiClient`, the transcription call (including the multipart handling) is reused as-is, so this PR is mostly Azure config and bean wiring.

The shared WireMock mapping for `/v1/audio/transcriptions` had a `Bearer` matcher; it is relaxed so the same endpoint also matches Azure's `api-key` auth, consistent with the chat mapping which does not match on auth.

The second commit clarifies in both transcription doc pages that the model is used directly (it is a specialized single-shot model with no messages/tools/guardrails) and is not part of the AI Services abstraction.

## Configuration

```properties
quarkus.langchain4j.azure-openai.api-key=...
quarkus.langchain4j.azure-openai.resource-name=...
quarkus.langchain4j.azure-openai.audio-transcription-model.deployment-name=whisper
```

## Injection

```java
@ApplicationScoped
public class TranscriptionService {

    @Inject
    AudioTranscriptionModel audioTranscriptionModel;

    public String transcribe(byte[] audioBytes, String mimeType) {
        Audio audio = Audio.builder().binaryData(audioBytes).mimeType(mimeType).build();
        return audioTranscriptionModel.transcribeToText(audio);
    }
}
```

---

- Closes #2514

### DIFF
--- a/model-providers/openai/testing-internal/src/main/resources/openai/mappings/completions.json
+++ b/model-providers/openai/testing-internal/src/main/resources/openai/mappings/completions.json
@@ -108,9 +108,6 @@
         "headers": {
           "Content-Type": {
             "matches": "multipart/form-data.*"
-          },
-          "Authorization": {
-            "matches": "Bearer .+"
           }
         }
       },
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -76,6 +76,7 @@
 *** xref:mistral-moderation-model.adoc[Mistral AI]
 *** xref:openai-moderation-model.adoc[OpenAI]
 ** Audio Transcription Models
+*** xref:azure-openai-audio-transcription-model.adoc[Azure OpenAI]
 *** xref:openai-audio-transcription-model.adoc[OpenAI]
 ** xref:enable-disable-integrations.adoc[Enabling and Disabling Integrations]
 * xref:testing.adoc[Testing & Evaluation]
--- a/docs/modules/ROOT/pages/azure-openai-audio-transcription-model.adoc
+++ b/docs/modules/ROOT/pages/azure-openai-audio-transcription-model.adoc
@@ -1,0 +1,82 @@
+= Azure OpenAI Audio Transcription Models
+
+include::./includes/attributes.adoc[]
+include::./includes/customization.adoc[]
+
+Azure OpenAI provides audio transcription models (Whisper) that convert spoken audio into text. These models are useful for use cases such as building voice assistants, transcribing meetings, or indexing podcast content.
+
+== Prerequisites
+
+include::./azure-openai-chat-model.adoc[tags=azure-openai-prerequisites]
+
+=== Azure OpenAI Quarkus Extension
+
+To use Azure OpenAI audio transcription models in your Quarkus application, add the following extension:
+
+:provider-artifact: quarkus-langchain4j-azure-openai
+include::./includes/quarkus-langchain4j-maven-dependencies.adoc[]
+
+[NOTE]
+====
+This extension also supports Azure OpenAI chat, embedding and image models. Moderation models are not currently supported by Azure OpenAI.
+====
+
+== Configuration
+
+include::includes/quarkus-langchain4j-azure-openai.adoc[leveloffset=+1,opts=optional]
+
+The transcription model is reached through your Azure OpenAI Whisper deployment, named by `deployment-name`:
+
+[source,properties,subs=attributes+]
+----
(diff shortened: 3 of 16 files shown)


model-providers/openai/testing-internal/src/main/resources/openai/mappings/completions.json
```diff
@@ -108,9 +108,6 @@
         "headers": {
           "Content-Type": {
             "matches": "multipart/form-data.*"
-          },
-          "Authorization": {
-            "matches": "Bearer .+"
```
geoand, line 113: Why this change?

omatheusmesmo: Azure authenticates with an `api-key` header, not `Authorization: Bearer`, so the shared mapping's Bearer matcher never matched the Azure request and the test failed. Dropped it so the mapping serves both, consistent with the `chat/completions` mapping which also doesn't match on auth.

geoand (approved):
Thanks!

PR #2526, opened 2026-06-03
fix(agentic): guard invokeAgent against arbitrary class reflection

### LINKED ISSUE #2525: security: AgenticJsonRpcService.invokeAgent allows arbitrary class reflection via DevUI
## Summary

`AgenticJsonRpcService.invokeAgent` (lines 67–97) accepts `agentClassName` as a raw string from the browser and passes it directly to `Class.forName()` with no allow-list check. Any class on the classpath can be instantiated and any method invoked.

## Affected file

`agentic/runtime/src/main/java/io/quarkiverse/langchain4j/agentic/runtime/devui/AgenticJsonRpcService.java`

```java
public JsonObject invokeAgent(String agentClassName, String methodName, String inputJson) {
    Class<?> agentClass = Class.forName(agentClassName, true, Thread.currentThread().getContextClassLoader());
    Object agent = Arc.container().select(agentClass).get();
    ...
    Object result = targetMethod.invoke(agent, args);
```

## Impact

- **Context:** Dev mode only (DevUI JSON-RPC endpoint)
- **Vector:** Any client able to reach the DevUI endpoint (localhost by default, but exposed in Docker, remote dev, or misconfigured firewall setups)
- **Effect:** Arbitrary class instantiation + method invocation on the server — effectively an RCE primitive

## Fix

The build-time data already available via `addBuildTimeData("agents", agentInfos)` contains the set of valid agent class names. The fix is a server-side allow-list check before `Class.forName`:

```java
if (!knownAgentClassNames.contains(agentClassName)) {
    return new JsonObject().put("success", false).put("error", "Unknown agent class: " + agentClassName);
}
```

The frontend already restricts selection to known agents — this adds the missing backend enforcement.

## Severity

Low-Medium (dev mode only, localhost by default) — but the pattern is unconditionally wrong and should be fixed regardless of default exposure.

mdproctor:
> **PR chain — Agentic module Quarkus integration**
> 1. **#2526** ← this PR — invokeAgent allowlist
> 2. #2534 (draft) — build-time safety checks and CDI foundation
> 3. #2555 (draft) — CDI wiring tiers and guardrails
> 4. #2544 (draft) — parallel context propagation
> 5. #2550 (draft) — OTel, Micrometer, CDI events, health checks
>
>
> 6. #2572 (planned) — @RegisterAiService simplification and composition annotations
>
> Each PR is standalone-compilable. Merge in order — each builds on the previous.

## Summary

`AgenticJsonRpcService.invokeAgent` accepted `agentClassName` as a raw string from the browser and passed it directly to `Class.forName()` with no validation — allowing any class on the classpath to be instantiated via the DevUI JSON-RPC endpoint.

### Changes

- **Allow-list check** — server-side guard in `invokeAgent` rejects unknown classes immediately
- **Build-time registration** — `AgenticDevUIProcessor` collects user agent class names at build time, registers via `AgenticRecorder.setDevUIAllowedAgentClassNames`
- **runtime-dev module** — `AgenticJsonRpcService` moved to `agentic/runtime-dev` so it never appears in production binaries
- **Conditional dev dependency** — `conditionalDevDependencies` registered in runtime `pom.xml`

### Impact

- **Context:** Dev mode only (`AgenticJsonRpcService` lives in `runtime-dev`)
- **Vector:** Any client able to reach the DevUI JSON-RPC endpoint
- **Effect without fix:** Arbitrary class instantiation + method invocation

---

- Fixes: #2525

### DIFF
--- a/agentic/runtime/pom.xml
+++ b/agentic/runtime/pom.xml
@@ -39,6 +39,9 @@
             </goals>
             <configuration>
               <deployment>${project.groupId}:${project.artifactId}-deployment:${project.version}</deployment>
+              <conditionalDevDependencies>
+                <artifact>${project.groupId}:quarkus-langchain4j-agentic-dev:${project.version}</artifact>
+              </conditionalDevDependencies>
             </configuration>
           </execution>
         </executions>
--- a/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/devui/AgenticDevUIProcessor.java
+++ b/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/devui/AgenticDevUIProcessor.java
@@ -103,6 +103,15 @@ public class AgenticDevUIProcessor {
 
     @BuildStep(onlyIf = IsDevelopment.class)
     @Record(ExecutionTime.RUNTIME_INIT)
+    void registerDevUIAllowedAgentClassNames(List<DetectedAiAgentBuildItem> agents, AgenticRecorder recorder) {
+        Set<String> classNames = filterUserAgents(agents).stream()
+                .map(a -> a.getIface().name().toString())
+                .collect(Collectors.toSet());
+        recorder.setDevUIAllowedAgentClassNames(classNames);
+    }
+
+    @BuildStep(onlyIf = IsDevelopment.class)
+    @Record(ExecutionTime.RUNTIME_INIT)
     @Consume(SyntheticBeansRuntimeInitBuildItem.class)
     void enableDevModeMonitoring(List<DetectedAiAgentBuildItem> agents,
             AgenticRecorder recorder) {
--- a/agentic/pom.xml
+++ b/agentic/pom.xml
@@ -14,6 +14,7 @@
   <modules>
     <module>deployment</module>
     <module>runtime</module>
+    <module>runtime-dev</module>
   </modules>
 
 
--- a/agentic/runtime-dev/pom.xml
+++ b/agentic/runtime-dev/pom.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.quarkiverse.langchain4j</groupId>
+    <artifactId>quarkus-langchain4j-agentic-parent</artifactId>
+    <version>999-SNAPSHOT</version>
+  </parent>
+  <artifactId>quarkus-langchain4j-agentic-dev</artifactId>
+  <name>Quarkus LangChain4j - Agentic - Dev</name>
+
+  <description>
+    Dev-mode-only classes for the agentic module. This module is included in the deployment
+    classpath (which is on the runtime classpath in dev mode) but is excluded from production
+    binaries. JSON-RPC services for Dev UI live here.
+  </description>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.quarkiverse.langchain4j</groupId>
+      <artifactId>quarkus-langchain4j-agentic</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-core</artifactId>
+    </dependency>
+    <dependency>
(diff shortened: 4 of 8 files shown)


.gitignore
```diff
@@ -80,3 +80,4 @@ release.properties
 /docs/node_modules/
 node_modules/
 antora-cache/
+wksp
```
mariofusco: What is this? I don't recognize this pattern.

mariofusco (approved):
I believe that this check is very reasonable and a good improvement, but I'm not an expert of this area.

@geoand @phillip-kruger can you please give a look at this?

geoand (commented):
I guess we can do this, but the same thing applies to other places WRT DevUI, right @phillip-kruger ?

phillip-kruger:
You should also move the JsonRPC service to runtime-dev, then it never end up in the prod binary. See https://quarkus.io/guides/dev-ui#communicating-to-the-backend. I am not sure if this happens anywhere else in Dev UI, I assumed it was just in this extension ?

.gitignore
mdproctor: removed

mdproctor:
> You should also move the JsonRPC service to runtime-dev, then it never end up in the prod binary. See https://quarkus.io/guides/dev-ui#communicating-to-the-backend. I am not sure if this happens anywhere else in Dev UI, I assumed it was just in this extension ?

doing it.

geoand:
@phillip-kruger can you give this another look?

phillip-kruger (changes requested):
You need to register the runtime-dev module. See https://quarkus.io/guides/dev-ui#jsonrpc-against-the-runtime-classpath : 

"If you are creating a new runtime-dev module, you must also add the conditional Dev dependency in the runtime module pom.xml:"

```
<plugin>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-extension-maven-plugin</artifactId>
    <configuration>
        <conditionalDevDependencies>
            <artifact>${project.groupId}:${project.artifactId}-dev:${project.version}</artifact>
        </conditionalDevDependencies>
        ...
    </configuration>
</plugin>
```

phillip-kruger:
Also see https://quarkus.io/guides/conditional-extension-dependencies

mdproctor:
Part of C1 alongside #2534 (quick wins + safety). Followed by #2542 (C2 — CDI auto-wiring) and #2544 (C3 — parallel safety).

mdproctor:
Added `conditionalDevDependencies` in the runtime `pom.xml` per the [Quarkus Dev UI guide](https://quarkus.io/guides/dev-ui#jsonrpc-against-the-runtime-classpath). The runtime-dev module is now properly registered as a conditional dev dependency.

Commits squashed to a single commit.

geoand:
PR needs to be updated to make CI pass

agentic/runtime/pom.xml
```diff
@@ -18,6 +18,21 @@
...
+    <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-core</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-smallrye-health</artifactId>
+      <optional>true</optional>
+    </dependency>
```
geoand: Are these necessary for this PR?

agentic/runtime/pom.xml
```diff
@@ -39,6 +54,9 @@
             </goals>
             <configuration>
               <deployment>${project.groupId}:${project.artifactId}-deployment:${project.version}</deployment>
+              <conditionalDevDependencies>
+                <artifact>${project.groupId}:quarkus-langchain4j-agentic-dev:${project.version}</artifact>
+              </conditionalDevDependencies>
```
geoand, line 44: @phillip-kruger is this the proper pattern? It looks right to me, just to make sure

geoand:
Once my two small comments have been answered, we can get this in

agentic/runtime/pom.xml
phillip-kruger: yes

geoand: 🙏🏽

agentic/runtime/pom.xml
mdproctor: No — removed. These are for the observability PR (#2550), not this one.

agentic/runtime/pom.xml
mdproctor: This registers the `runtime-dev` module as a conditional dev dependency per the [Quarkus Dev UI guide](https://quarkus.io/guides/dev-ui#jsonrpc-against-the-runtime-classpath). The `AgenticJsonRpcService` lives in `runtime-dev` so it never ends up in production builds.

geoand (approved):
Thanks!

geoand:
The CI failure has nothing to do with the change, so I'll merge this

PR #2529, opened 2026-06-03
Support IMMEDIATE_IF_LAST and IMMEDIATE returntype not Result

### LINKED ISSUE #2461: returnBehavior=IMMEDIATE should not require a Result method return type in certain conditions
An AiService must return a Result if there is a tool with returnBehavior=IMMEDIATE called.  IMO this should not be required in certain situations, specifically:

* If there is only one tool invoked and the tool result can be converted to the method return type
* If all tool results are null and null is a valid response

We shouldn't be forcing the user to unpack a Result object.

https://github.com/quarkiverse/quarkus-langchain4j/blob/cc7de1116459dfff513830ec87dd38095c7bd716/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java#L495

### LINKED ISSUE #2528: Support IMMEDIATE_IF_LAST
Support IMMEDIATE_IF_LAST

patriot1burke:
- Closes: #2528 
- Closes: #2461 

If tool execution is IMMEDIATE or IMMEDIATE_IF_LAST and return type is not a Result
* If ai service method return type is void, just return null
* If any tool executions are TO_LLM throw exception (IMMEDIATE_IF_LAST case)
* If all tool executions have a null result and return type is nullable (not a primitive), then return null
* If there is one non-null tool execution result, and the result can be resolved to the return type, then return that object

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionModelTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolExecutionModelTest.java
@@ -4,6 +4,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.fail;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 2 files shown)


patriot1burke:
@mariofusco If you haven't implemented #2461 yet here's a PR and tests for it.

mariofusco:
The implementation is ok for me, but the problem is that this pull request makes the quarkus extension to behave in a different way than vanilla langchain4j and I don't like it.

In my opinion either we do the same also in langchain4j or we leave everything as it is. I want to avoid that they could have a divergent behavior on this.

@geoand @dliubarskyi wdyt?

geoand:
Ideally the behavior would be the same with upstream. Obviously that's not the case for all features, but when doable (as in this case I pressume), we should do it

mariofusco:
> Ideally the behavior would be the same with upstream. Obviously that's not the case for all features, but when doable (as in this case I pressume), we should do it

I totally agree.

@dliubarskyi is it ok for you to implement the same behavior also in langchain4j?

patriot1burke:
I can implement it in LC4J too.

@dliubarskyi  Has already agreed to this issue?  

https://github.com/quarkiverse/quarkus-langchain4j/issues/2461#issuecomment-4565187805

patriot1burke:
@mariofusco Good with this PR?  I'll merge it and work on LC4J patch.

mariofusco:
> @mariofusco Good with this PR? I'll merge it and work on LC4J patch.

Please let me check @dliubarskyi's feedback on this before moving forward.

patriot1burke:
@mariofusco @dliubarskyi  Just updated OP PR Comment to describe behavior better.

patriot1burke:
@mariofusco 

PR for LC4J:

https://github.com/langchain4j/langchain4j/pull/5367

dliubarskyi:
Sorry for the late response (there was a public holiday yesterday)... I have reviewed https://github.com/langchain4j/langchain4j/pull/5367, looks good, mostly nitpicks, but there is [one bigger issue](https://github.com/langchain4j/langchain4j/pull/5367#discussion_r3361235849)

patriot1burke:
@mariofusco going to merge this as https://github.com/langchain4j/langchain4j/pull/5367 is now merged and closed.

PR #2531, opened 2026-06-04
Fix NPE for single shared ChatMemory in active request scope

### LINKED ISSUE #2381: NullPointerException in ChatMemoryService when programmatic AiServices.builder() uses .chatMemory(...) inside an active request scope

omatheusmesmo:
## Describe your changes

When an AiService is built programmatically with a single shared `ChatMemory` (no `ChatMemoryProvider` and no `@MemoryId` parameter) and invoked inside an active request scope, the `DefaultMemoryIdProvider` chain produced a non-default memory id (the request-scope state). A single-memory `ChatMemoryService` cannot resolve that id, so `getChatMemory(memoryId)` returned `null` and the streaming response handler threw a `NullPointerException`.

`memoryId()` now consults the `DefaultMemoryIdProvider` chain only when a `ChatMemoryProvider` is actually configured, tracked via a new `hasChatMemoryProvider` flag on `QuarkusAiServiceContext`.

---

- Closes #2381

---

> Validated locally against the reporter's reproducer (https://github.com/Colprit/issue-2381-repro) built against this branch (`999-SNAPSHOT`): `GET /ask` now returns `200 ok` with no NPE. A streaming regression test reproduces the original SSE scenario and the existing memory/scope/streaming suite passes.

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/StreamingSingleSharedChatMemoryInRequestScopeTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/StreamingSingleSharedChatMemoryInRequestScopeTest.java
@@ -1,0 +1,86 @@
+package io.quarkiverse.langchain4j.test;
+
+import static io.restassured.RestAssured.get;
+import static org.hamcrest.Matchers.equalTo;
+
+import java.util.function.Supplier;
+
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.memory.ChatMemory;
+import dev.langchain4j.memory.chat.MessageWindowChatMemory;
+import dev.langchain4j.model.chat.StreamingChatModel;
+import dev.langchain4j.model.chat.request.ChatRequest;
+import dev.langchain4j.model.chat.response.ChatResponse;
+import dev.langchain4j.model.chat.response.StreamingChatResponseHandler;
+import dev.langchain4j.service.AiServices;
+import io.quarkiverse.langchain4j.RegisterAiService;
+import io.quarkus.test.QuarkusUnitTest;
+import io.smallrye.mutiny.Multi;
+
+/**
+ * Verifies that a programmatically built streaming AiService with a single shared ChatMemory
+ * (no ChatMemoryProvider, no @MemoryId) can be invoked from within an active request scope, using
+ * the shared memory instead of resolving a non-default request-scope memory id it cannot handle.
+ */
+public class StreamingSingleSharedChatMemoryInRequestScopeTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest unitTest = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(StreamingAssistant.class, ChatResource.class, FakeStreamingChatModelSupplier.class));
+
+    @Test
+    public void streamingServiceWithSingleSharedMemoryDoesNotFailInsideRequestScope() {
+        get("/streaming-shared-memory")
+                .then()
+                .statusCode(200)
+                .body(equalTo("42"));
+    }
+
+    @RegisterAiService(streamingChatLanguageModelSupplier = FakeStreamingChatModelSupplier.class, chatMemoryProviderSupplier = RegisterAiService.NoChatMemoryProviderSupplier.class)
+    public interface StreamingAssistant {
+
+        Multi<String> chat(String message);
+    }
+
+    @Path("/streaming-shared-memory")
+    public static class ChatResource {
+
+        private static final ChatMemory SHARED_MEMORY = MessageWindowChatMemory.withMaxMessages(20);
+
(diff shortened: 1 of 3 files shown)


omatheusmesmo:
Hello @Colprit ,

Could you please give a try? 

Thanks.

cescoffier:

omatheusmesmo:
Hello @cescoffier @geoand , do you mind test using the reproducer and review if is it a proper solution?

I think @Colprit is unavailable for it. I know it's not a bug affecting too much users, but would be really nice if you could help me, I would really appreciate.

Thanks.

geoand (approved):

PR #2539, opened 2026-06-05
Fix native mode substitution for AiServices.builder() to use context lookup

### LINKED ISSUE #2538: Native mode substitution for AiServices.builder() bypasses CDI context lookup
Follows https://github.com/quarkiverse/quarkus-langchain4j/issues/2344

  The GraalVM substitution for AiServices.builder() in Substitutions.java creates a bare  QuarkusAiServiceContext directly, bypassing the CDI lookup that QuarkusAiServiceContextFactory performs  in JVM mode.

  JVM mode path:
  AiServices.builder(class)
    → AiServiceContext.create(class)
      → ServiceLoader finds QuarkusAiServiceContextFactory
        → Arc.container().instance(QuarkusAiServiceContext.class, qualifier)
          → returns CDI-managed context with chatMemoryService initialized
          
  Native mode path (current):
  AiServices.builder(class)  [substituted]
    → new QuarkusAiServiceContext(aiService)
      → bare context, no CDI lookup, chatMemoryService is null

  This causes hasChatMemory() to return false in native mode even when a ChatMemoryProvider bean is configured. The AiServiceMethodImplementationSupport then falls back to NoopChatMemory, which discards all messages. Any AI service that uses tools fails in native mode with "messages cannot be null or  empty" because the follow-up chat request after tool execution has no conversation history.

  Fix:

  Change Substitutions.Target_AiServices.builder() from:
  return new QuarkusAiServicesFactory.QuarkusAiServices<>(new QuarkusAiServiceContext(aiService));
  to:
  return new QuarkusAiServicesFactory.QuarkusAiServices<>(new
  QuarkusAiServiceContextFactory().create(aiService));

  This aligns native mode with JVM mode by using the same CDI lookup path.

  Discovered during: Apache Camel Quarkus integration testing — promoting langchain4j-agent-ql4j from JVM-only to native mode support (https://github.com/apache/camel-quarkus/issues/8690).

JiriOndrusek:
Fix `AiServices.builder()` GraalVM substitution to use CDI context lookup, aligning native mode behavior
  with JVM mode.

  The native substitution in `Substitutions.Target_AiServices` was creating a bare
  `QuarkusAiServiceContext` directly (`new QuarkusAiServiceContext(aiService)`), bypassing the
  `QuarkusAiServiceContextFactory` CDI lookup that JVM mode uses via ServiceLoader. This caused
  `chatMemoryService` to be null in native mode, making `hasChatMemory()` return false even when a
  `ChatMemoryProvider` bean was configured. As a result, `NoopChatMemory` was used instead, breaking tool
  execution loops with `"messages cannot be null or empty"`.

  The fix changes the substitution to delegate to `QuarkusAiServiceContextFactory().create(aiService)`,
  which performs the same `Arc.container().instance()` lookup as JVM mode.

- Fixes: https://github.com/quarkiverse/quarkus-langchain4j/issues/2538

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/graalvm/Substitutions.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/graalvm/Substitutions.java
@@ -15,13 +15,13 @@ import dev.langchain4j.spi.prompt.PromptTemplateFactory;
 import dev.langchain4j.spi.prompt.structured.StructuredPromptFactory;
 import dev.langchain4j.store.embedding.inmemory.InMemoryEmbeddingStore;
 import dev.langchain4j.store.embedding.inmemory.InMemoryEmbeddingStoreJsonCodec;
+import io.quarkiverse.langchain4j.QuarkusAiServiceContextFactory;
 import io.quarkiverse.langchain4j.QuarkusAiServicesFactory;
 import io.quarkiverse.langchain4j.QuarkusChatMessageJsonCodecFactory;
 import io.quarkiverse.langchain4j.QuarkusInMemoryEmbeddingJsonCodecFactory;
 import io.quarkiverse.langchain4j.QuarkusJsonCodecFactory;
 import io.quarkiverse.langchain4j.QuarkusPromptTemplateFactory;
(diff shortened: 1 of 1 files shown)


JiriOndrusek:
@geoand FYI

geoand (approved):
Thanks!

PR #2543, opened 2026-06-06
Expose LangChain4j Tool Search in `@RegisterAiService`

### LINKED ISSUE #2533: Expose LangChain4j Tool Search (`toolSearchStrategy`) in the `@RegisterAiService` model
## Summary

LangChain4j core added the **Tool Search** feature (`AiServices.toolSearchStrategy(ToolSearchStrategy)`, package `dev.langchain4j.service.tool.search`) in [langchain4j #4570](https://github.com/langchain4j/langchain4j/pull/4570) (merged 2026-02-23), based on Anthropic's tool-search-tool pattern. It lets the LLM discover tools dynamically via dedicated search tools instead of exposing the full tool catalog upfront, which reduces token usage and improves model performance when many tools are registered. `SimpleToolSearchStrategy` and `VectorToolSearchStrategy` are provided out of the box.

This capability is currently **not reachable** from the declarative `@RegisterAiService` programming model in quarkus-langchain4j.

## Current state

`AiServicesRecorder` wires core builder capabilities into the AI service one by one (`tools`, `toolProvider`, `toolHallucinationStrategy`, `maxToolCallingRoundTrips`, `maxToolCallsPerResponse`, etc.). `toolSearchStrategy` is not among them. Concretely, there is:

- no `@RegisterAiService` attribute for it,
- no CDI bean pickup (unlike `toolProvider`, which has `BeanIfExistsToolProviderSupplier`),
- no general AI-service builder customizer (`ModelBuilderCustomizer` only targets model builders, not the `AiServices` builder).

As a result, the only way to use Tool Search today is to build the AI service fully programmatically with `AiServices.builder()...toolSearchStrategy(...)`, which bypasses `@RegisterAiService` and the CDI integration most Quarkus apps rely on.

## Proposal

Make `toolSearchStrategy` configurable for `@RegisterAiService` interfaces, so the Quarkus declarative model reaches parity with the core builder.

There is an open design choice I would like maintainer input on before implementing:

1. **Targeted wiring** - follow the existing `toolProviderSupplier` pattern: pick up a `ToolSearchStrategy` CDI bean if present (and/or add a supplier attribute on `@RegisterAiService`), then call `quarkusAiServices.toolSearchStrategy(...)` in `AiServicesRecorder`. `VectorToolSearchStrategy` needs an `EmbeddingModel`, which fits naturally as a CDI bean.

2. **General hook** - introduce an `AiServices` builder customizer (analogous to `ModelBuilderCustomizer`, but for the AI service builder). This would close this gap and also cover other core builder capabilities that are not yet exposed declaratively (for example `executeToolsConcurrently`), avoiding one-off wiring each time a new core feature lands.

---

### Relation to the Working Group

This is part of the **Cost & Observability** epic (#1899), under the WG proposal in [quarkusio/quarkus#50582](https://github.com/quarkusio/quarkus/discussions/50582), which explicitly lists "select required tools dynamically via `AiServices.toolSelector` (to be implemented)" as a token-minimization technique. The core side of that item is now implemented (as Tool Search); this issue tracks exposing it in the Quarkus declarative model.

#### Contribution

I am willing to implement this and open a PR once the preferred direction (1 vs 2) is agreed.

omatheusmesmo:
## Describe your changes

This submission exposes langchain4j's Tool Search (`ToolSearchStrategy`) as a declarative option on `@RegisterAiService`, letting the model discover tools dynamically at inference time instead of receiving the whole tool catalog upfront.

The design follows the established `toolProviderSupplier` pattern. A new `toolSearchStrategySupplier()` attribute defaults to `BeanIfExistsToolSearchStrategySupplier`: if exactly one `ToolSearchStrategy` CDI bean exists it is applied automatically, otherwise no tool search happens. A custom `Supplier<ToolSearchStrategy>` selects a specific instance, and `NoToolSearchStrategySupplier` opts out explicitly even when such a bean is present.

The non-obvious part is the runtime integration. Quarkus drives its own tool-calling loop rather than going through core's `ToolService`, so wiring the strategy into the AI service context is not enough: the loop has to apply the search service itself. The strategy is wrapped in core's `ToolSearchService` and stored on `QuarkusAiServiceContext`, then applied at two points in both the blocking and streaming paths:

- `adjust()` narrows the upfront catalog to the search tools before the first request, so the model only sees the search tool;
- `addFoundTools()` re-adds the tools the model surfaces through the search tool after each round, so they become callable on the next turn.

Blocking support lives in `AiServiceMethodImplementationSupport`; streaming support spans `QuarkusAiServiceTokenStream` and `QuarkusAiServiceStreamingResponseHandler`. The full catalog survives in the returned context across rounds, so discovered tools accumulate as the conversation progresses. The feature works with core's built-in strategies (for example `VectorToolSearchStrategy`) and with custom ones.

## Usage

Declare a `ToolSearchStrategy` as a CDI bean and it is picked up automatically:

```java
@ApplicationScoped
public class MyToolSearchStrategy implements ToolSearchStrategy {
    // returns the search tool(s) and resolves which tools to surface per request
}

@RegisterAiService(tools = BookingTools.class)
public interface BookingAssistant {
}
```

To use a specific instance, point the attribute at a supplier:

```java
@RegisterAiService(tools = BookingTools.class, toolSearchStrategySupplier = MyStrategySupplier.class)
public interface BookingAssistant {
}
```

To opt out when a strategy bean exists:

```java
@RegisterAiService(tools = BookingTools.class, toolSearchStrategySupplier = NoToolSearchStrategySupplier.class)
```

---

- Closes #2533

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/LangChain4jDotNames.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/LangChain4jDotNames.java
@@ -35,6 +35,7 @@ import dev.langchain4j.service.guardrail.OutputGuardrails;
 import dev.langchain4j.service.tool.ToolErrorContext;
 import dev.langchain4j.service.tool.ToolErrorHandlerResult;
 import dev.langchain4j.service.tool.ToolProvider;
+import dev.langchain4j.service.tool.search.ToolSearchStrategy;
 import dev.langchain4j.web.search.WebSearchEngine;
 import dev.langchain4j.web.search.WebSearchTool;
 import io.quarkiverse.langchain4j.AudioUrl;
@@ -131,6 +132,12 @@ public class LangChain4jDotNames {
     static final DotName NO_TOOL_PROVIDER_SUPPLIER = DotName.createSimple(
             RegisterAiService.NoToolProviderSupplier.class);
 
+    static final DotName BEAN_IF_EXISTS_TOOL_SEARCH_STRATEGY_SUPPLIER = DotName.createSimple(
+            RegisterAiService.BeanIfExistsToolSearchStrategySupplier.class);
+
+    static final DotName NO_TOOL_SEARCH_STRATEGY_SUPPLIER = DotName.createSimple(
+            RegisterAiService.NoToolSearchStrategySupplier.class);
+
     static final DotName QUARKUS_AI_SERVICE_CONTEXT_QUALIFIER = DotName.createSimple(
             QuarkusAiServiceContextQualifier.class);
 
@@ -156,6 +163,8 @@ public class LangChain4jDotNames {
     static final DotName VIDEO = DotName.createSimple(dev.langchain4j.data.video.Video.class);
     static final DotName RESULT = DotName.createSimple(Result.class);
     public static final DotName TOOL_PROVIDER = DotName.createSimple(ToolProvider.class);
+    public static final DotName TOOL_SEARCH_STRATEGY = DotName
+            .createSimple(ToolSearchStrategy.class);
     // Using the class name to keep the McpToolBox annotation in the mcp module
     public static final DotName MCP_TOOLBOX = DotName.createSimple("io.quarkiverse.langchain4j.mcp.runtime.McpToolBox");
     public static final DotName CHAT_EVENT = DotName.createSimple(ChatEvent.class);
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
@@ -104,6 +104,8 @@ import dev.langchain4j.service.tool.ToolExecutor;
 import dev.langchain4j.service.tool.ToolProvider;
 import dev.langchain4j.service.tool.ToolProviderRequest;
 import dev.langchain4j.service.tool.ToolProviderResult;
+import dev.langchain4j.service.tool.ToolServiceContext;
+import dev.langchain4j.service.tool.search.ToolSearchService;
 import dev.langchain4j.spi.ServiceHelper;
 import io.quarkiverse.langchain4j.AudioUrl;
 import io.quarkiverse.langchain4j.ImageUrl;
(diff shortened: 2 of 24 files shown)


core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/LangChain4jDotNames.java
```diff
@@ -156,6 +162,8 @@ public class LangChain4jDotNames {
     static final DotName VIDEO = DotName.createSimple(dev.langchain4j.data.video.Video.class);
     static final DotName RESULT = DotName.createSimple(Result.class);
     public static final DotName TOOL_PROVIDER = DotName.createSimple(ToolProvider.class);
+    public static final DotName TOOL_SEARCH_STRATEGY = DotName
+            .createSimple(dev.langchain4j.service.tool.search.ToolSearchStrategy.class);
```
geoand: No need to use the full name here

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -392,6 +394,22 @@ private static Object doImplement0(AiServiceMethodCreateInfo methodCreateInfo, I
...
+        if (toolSearchService != null) {
+            toolSearchContext = toolSearchService.adjust(
+                    new ToolServiceContext.Builder()
+                            .effectiveTools(toolSpecifications)
+                            .availableTools(toolSpecifications)
+                            .toolExecutors(toolExecutors)
+                            .build(),
+                    messagesToSend, invocationContext);
+            toolSpecifications = new ArrayList<>(toolSearchContext.effectiveTools());
+            toolExecutors = new HashMap<>(toolSearchContext.toolExecutors());
```
geoand, line 410: How does this play with the existing code that sets these fields?

omatheusmesmo: > How does this play with the existing code that sets these fields?

Hi @geoand ,

By L409 these locals hold the fully assembled catalog for the call: specs and executors from the method tools or `context.toolService`, plus whatever the tool providers (e.g. MCP) appended (L260-277). That full set is fed to `adjust()` as both its `effectiveTools` and its `availableTools`.

`adjust()` then splits the two apart: `availableTools` stays the full catalog (untouched, kept in the returned context), while `effectiveTools` is narrowed to just the search tool. L409 stores that narrowed `effectiveTools` into `toolSpecifications`, and L410 stores the executors (the original set merged with the search-tool executor) into `toolExecutors`. So from here `availableTools` is the fixed backing catalog and `effectiveTools` is the moving subset actually offered on each request.

That is why the per-round update at L582 refreshes only `toolSpecifications`: `addFoundTools` grows `effectiveTools` by pulling discovered specs out of the preserved `availableTools`, while the executor map from L410 is already complete and never needs refreshing. Nothing the existing assembly produced is dropped, and when no strategy is configured the block is skipped, so these keep exactly what L260-277 set.

`ToolSearchStrategyTest` covers it end to end: the mock answers `TOOL_NOT_NARROWED` if the catalog wasn't narrowed on the first request and `FOUND_TOOL_NOT_AVAILABLE` if the searched tool wasn't callable afterwards, so it passes only when both hold.

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/LangChain4jDotNames.java
omatheusmesmo: I'm sorry, I hate using FQN imports and I missed this one.

geoand (approved):
Thanks!

PR #2552, opened 2026-06-08
Bump to LangChain4j 1.16.1 + implement tools inheritance

### LINKED ISSUE #2493: @Tool methods not inherited from superclass
@Tool methods not inherited from superclass. I think this is a bug as this is something that should work.

I didn't 100% verify this, but my inherited tools weren't being called and the LLM was stating that it didn't know what tool to call. Once I put the tool methods back in the concrete class, things worked, so I'm assuming that this bug report is true.

This should be done in LC4J too if it already doesn't work there.

mariofusco:
- Fixes: https://github.com/quarkiverse/quarkus-langchain4j/issues/2493

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,7 +39,7 @@
   <properties>
     <quarkus.version>3.33.2</quarkus.version>
     <quarkus.extension-processor.version>${quarkus.version}</quarkus.extension-processor.version>
-    <langchain4j.version>1.15.1</langchain4j.version>
+    <langchain4j.version>1.16.1</langchain4j.version>
     <langchain4j-community.version>1.6.0-beta12</langchain4j-community.version>
     <quarkus-antora.version>3.25.0</quarkus-antora.version>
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -39,7 +39,7 @@
   <properties>
     <quarkus.version>3.33.2</quarkus.version>
     <quarkus.extension-processor.version>${quarkus.version}</quarkus.extension-processor.version>
-    <langchain4j.version>1.15.1</langchain4j.version>
+    <langchain4j.version>1.16.0</langchain4j.version>
```
geoand: `1.16.1` should also be available, so let's use that

mariofusco: Done.

geoand (approved):
Thanks!

geoand:
The failure seems related. Do you want me to look into it @mariofusco ?

mariofusco:
> The failure seems related. Do you want me to look into it @mariofusco ?

I'm giving a look right now, will keep you posted.

geoand:
🆗

geoand (approved):

mariofusco:
It should be ok now.

jmartisk (approved):

PR #2553, opened 2026-06-08
fix(mistral): guard null delta content in streaming responses

HakimMohammed:
Mistral emits tool-call deltas with delta.content == null.
The unguarded for-each over getContent() threw a NullPointerException, breaking streaming when calling functions. Adds a regression test.


geoand (approved):
Thanks!

PR #2554, opened 2026-06-08
Marshal AgenticScope through the managed ObjectMapper

omatheusmesmo:
## Describe your changes

A langchain4j `DefaultAgenticScope` can't be serialized by the Quarkus-managed Jackson `ObjectMapper` (it exposes no discoverable bean properties), so `writeValueAsString` throws `InvalidDefinitionException`. langchain4j already serializes it correctly through its own `AgenticScopeSerializer`; it just isn't wired into the application's `ObjectMapper`.

This adds an `ObjectMapperCustomizer` that registers a `SimpleModule` delegating `AgenticScope` (de)serialization to `AgenticScopeSerializer`, plus a build step registering the classes the codec marshals reflectively for native image (it uses Jackson default typing, which embeds concrete class names). A consumer like `quarkus-flow` reaches this because it injects the managed `ObjectMapper` bean into the serverlessworkflow SDK.

On where this belongs: langchain4j keeps its Jackson codec `@Internal` and exposes only the String-based `AgenticScopeSerializer` as public API, so bridging it into a framework-managed `ObjectMapper` is by design the integrator's job, which is what this PR does. (I can ask upstream whether they'd consider a public Jackson `Module`, which would shrink the serializer here to a one-line registration, but the `ObjectMapperCustomizer` wiring stays regardless.)

Two regression tests: `AgenticScopeSerializationTest` (runtime) pins the contract (vanilla mapper throws, `AgenticScopeSerializer` round-trips), and `AgenticScopeObjectMapperTest` (deployment) checks the managed mapper round-trips a scope, including when nested.

Tested locally at module level (runtime + deployment); relying on CI for the full and native builds.

---

- Relates to quarkiverse/quarkus-flow#613 , quarkiverse/quarkus-flow#614 

(the Quarkus LangChain4j side of it; the flow workflow model still needs to stop swallowing the failure).

### DIFF
--- a/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
+++ b/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
@@ -65,6 +65,7 @@ import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
 import io.quarkus.deployment.builditem.IndexDependencyBuildItem;
 import io.quarkus.deployment.builditem.ServiceStartBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.NativeImageProxyDefinitionBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
 
 public class AgenticProcessor {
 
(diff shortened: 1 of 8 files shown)


geoand:
Thanks for this, but I really don't understand why the `AgenticScope` would ever need to be serialized, so I need input from @ricardozanini and @mariofusco on this

mariofusco:
I'm probably a bit out of context as well. I understand that it is necessary to serialize the `AgenticScope` in order to persist it, but I don't see why you may want to do so through the the Quarkus-managed Jackson `ObjectMapper`. 

For instance in [this PoC](https://github.com/mariofusco/quarkus-langchain4j/commit/7e59fb397b116538c9e21f1fa7c9cee4ce6f0a4a) I created an Infinispan-based reference implementation for persistence of an agentic system by directly serializing the `AgenicScope` in json using directly the mechanism provided out-of-the-box by langchain4j. Wouldn't be enough to do something similar? 

That said, considering that the `AgenticScope` already contains the complete state of the agentic system, can you please also clarify which features `quarkus-flow` based persistence provides in addition to what I sketched in that PoC?

omatheusmesmo:
> Thanks for this, but I really don't understand why the `AgenticScope` would ever need to be serialized, so I need input from @ricardozanini and @mariofusco on this

> I'm probably a bit out of context as well. I understand that it is necessary to serialize the `AgenticScope` in order to persist it, but I don't see why you may want to do so through the the Quarkus-managed Jackson `ObjectMapper`.
> 
> For instance in [this PoC](https://github.com/mariofusco/quarkus-langchain4j/commit/7e59fb397b116538c9e21f1fa7c9cee4ce6f0a4a) I created an Infinispan-based reference implementation for persistence of an agentic system by directly serializing the `AgenicScope` in json using directly the mechanism provided out-of-the-box by langchain4j. Wouldn't be enough to do something similar?
> 
> That said, considering that the `AgenticScope` already contains the complete state of the agentic system, can you please also clarify which features `quarkus-flow` based persistence provides in addition to what I sketched in that PoC?

Thanks @mariofusco, and @geoand. I hadn't realized this yesterday, I did this PR rushing during some work shift breaks.

`AgenticScopeSerializer` is already available out of the box. So there is no need to route `AgenticScope` serialization through the Quarkus-managed Jackson `ObjectMapper`.

I'm moving the fix into the quarkus-flow persistence marshaller, where it serializes the `AgenticScope` directly with `AgenticScopeSerializer.toJson/fromJson`, in the same spirit as the `AgenticScopeStore` PoC. The marshaller is verified on PostgreSQL. That means this change is not needed for the persistence use case, so I'm happy to close this if we agree it is not needed elsewhere.

@ricardozanini FYI

PR #2558, opened 2026-06-09
Replace proxy-based agents with deployment time generated ones

mariofusco:
~~This pull request depends on the work done here https://github.com/langchain4j/langchain4j/pull/5397~~

### DIFF
--- a/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
+++ b/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
@@ -5,6 +5,8 @@ import static io.quarkiverse.langchain4j.agentic.deployment.ValidationUtil.valid
 import static io.quarkiverse.langchain4j.agentic.deployment.ValidationUtil.validateRequiredParameterTypes;
 import static io.quarkiverse.langchain4j.agentic.deployment.ValidationUtil.validateStaticMethod;
 
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -31,17 +33,20 @@ import org.jboss.jandex.IndexView;
 import org.jboss.jandex.MethodInfo;
 import org.jboss.jandex.MethodParameterInfo;
 import org.jboss.jandex.ParameterizedType;
+import org.jboss.jandex.PrimitiveType;
 import org.jboss.jandex.Type;
 import org.jboss.logging.Logger;
 
 import dev.langchain4j.agentic.agent.ChatMessagesAccess;
 import dev.langchain4j.agentic.declarative.ParallelMapperAgent;
-import dev.langchain4j.agentic.internal.AgenticScopeOwner;
 import dev.langchain4j.agentic.internal.InternalAgent;
+import dev.langchain4j.agentic.scope.AgenticScopeAccess;
 import dev.langchain4j.observability.api.listener.AiServiceResponseReceivedListener;
 import dev.langchain4j.service.IllegalConfigurationException;
 import dev.langchain4j.service.memory.ChatMemoryAccess;
 import io.quarkiverse.langchain4j.ModelName;
+import io.quarkiverse.langchain4j.agentic.runtime.AbstractQuarkusAgent;
+import io.quarkiverse.langchain4j.agentic.runtime.AgentClassCreateInfo;
 import io.quarkiverse.langchain4j.agentic.runtime.AgenticRecorder;
 import io.quarkiverse.langchain4j.agentic.runtime.AiAgentCreateInfo;
 import io.quarkiverse.langchain4j.deployment.AnnotationsImpliesAiServiceBuildItem;
@@ -53,6 +58,8 @@ import io.quarkiverse.langchain4j.deployment.RequestChatModelBeanBuildItem;
 import io.quarkiverse.langchain4j.deployment.SkipOutputFormatInstructionsBuildItem;
 import io.quarkiverse.langchain4j.runtime.NamedConfigUtil;
 import io.quarkus.arc.deployment.BeanDiscoveryFinishedBuildItem;
+import io.quarkus.arc.deployment.GeneratedBeanBuildItem;
+import io.quarkus.arc.deployment.GeneratedBeanGizmoAdaptor;
 import io.quarkus.arc.deployment.InterceptorResolverBuildItem;
 import io.quarkus.arc.deployment.SyntheticBeanBuildItem;
 import io.quarkus.arc.deployment.UnremovableBeanBuildItem;
@@ -64,11 +71,22 @@ import io.quarkus.deployment.annotations.Record;
 import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
 import io.quarkus.deployment.builditem.IndexDependencyBuildItem;
 import io.quarkus.deployment.builditem.ServiceStartBuildItem;
-import io.quarkus.deployment.builditem.nativeimage.NativeImageProxyDefinitionBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
+import io.quarkus.gizmo.CatchBlockCreator;
+import io.quarkus.gizmo.ClassCreator;
+import io.quarkus.gizmo.ClassOutput;
+import io.quarkus.gizmo.FieldDescriptor;
+import io.quarkus.gizmo.MethodCreator;
+import io.quarkus.gizmo.MethodDescriptor;
+import io.quarkus.gizmo.ResultHandle;
+import io.quarkus.gizmo.TryBlock;
 
 public class AgenticProcessor {
 
     private static final Logger log = Logger.getLogger(AgenticProcessor.class);
+
+    private static final MethodDescriptor HANDLER_INVOKE = MethodDescriptor.ofMethod(
+            InvocationHandler.class, "invoke", Object.class, Object.class, Method.class, Object[].class);
 
     @BuildStep
     void indexDependencies(BuildProducer<IndexDependencyBuildItem> producer) {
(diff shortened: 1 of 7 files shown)


agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
```diff
@@ -64,12 +76,29 @@
...
 
 public class AgenticProcessor {
 
     private static final Logger log = Logger.getLogger(AgenticProcessor.class);
 
+    private static final Map<String, String> PRIMITIVE_TO_WRAPPER = Map.of(
+            "boolean", "java.lang.Boolean", "byte", "java.lang.Byte",
+            "char", "java.lang.Character", "short", "java.lang.Short",
+            "int", "java.lang.Integer", "long", "java.lang.Long",
+            "float", "java.lang.Float", "double", "java.lang.Double");
```
geoand: Do we have this in Quarkus core or Quarkus LangChain4j core already?

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
```diff
@@ -64,12 +76,29 @@
...
+
+    private static final MethodDescriptor HANDLER_INVOKE = MethodDescriptor.ofMethod(
+            InvocationHandler.class, "invoke", Object.class, Object.class, Method.class, Object[].class);
```
geoand, line 89: As mentioned in https://github.com/langchain4j/langchain4j/pull/5397, I don't believe `InvocationHandler` should be part of the signature

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
```diff
@@ -503,6 +542,22 @@ void cdiSupport(List<DetectedAiAgentBuildItem> detectedAiAgentBuildItems, Agenti
...
+        if (annClass != null) {
+            return annClass.hasAnnotation(INTERCEPTOR_BINDING);
+        }
+        try {
+            Class<?> annClazz = Class.forName(ann.name().toString(), false,
+                    Thread.currentThread().getContextClassLoader());
+            return annClazz.isAnnotationPresent(jakarta.interceptor.InterceptorBinding.class);
+        } catch (ClassNotFoundException e) {
+            return false;
+        }
```
geoand, line 547: How come we need to fallback to this?

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
```diff
@@ -520,16 +575,370 @@ private static boolean hasAnyInterceptorBindings(DetectedAiAgentBuildItem agent,
...
+            generateAgentClass(classOutput, bi.getIface(), implClassName, isLeaf, index);
+
+            metadata.put(ifaceName, new AgentClassCreateInfo(implClassName));
+            reflectiveClassProducer.produce(ReflectiveClassBuildItem.builder(implClassName).build());
+        }
+
+        recorder.setAgentClassMetadata(metadata);
+    }
+
+    private void generateAgentClass(ClassOutput classOutput, ClassInfo agentIface, String implClassName,
```
geoand, line 640: Please add a comment showing an example generated class

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
mariofusco: When running the `FaultToleranceTest` the `@Timeout` annotation cannot be found on the `IndexView`, but conversely it can be loaded with that `Class.forName`. Is this expected? Am I doing something wrong?

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
mariofusco: Actually this `HANDLER_INVOKE` is not related with the other comment you linked, but it simply used internally to the `AgenticProcessor` to generate the code invoking the langchain4j agent.

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
mariofusco: Removed.

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
mariofusco: Done.

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
mariofusco: Claude's more detailed response: "The Class.forName fallback is necessary. The FaultToleranceTest proved it: interceptor binding annotations like @Timeout (from MicroProfile Fault Tolerance) come from external libraries that aren't in the Jandex index. Without the fallback, those annotations wouldn't be recognized
  as interceptor bindings, wouldn't be copied to the generated methods, and interception wouldn't work. It's the same pattern used by AiServicesProcessor.hasInterceptorBinding() for the same reason."

geoand: I'll have to run it myself and see

mariofusco:
@geoand I should have addressed all your comments, thanks for review. Please give it a second look.

geoand:
Thanks. I'll run it tomorrow

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
geoand: I pushed an update for this

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
```diff
@@ -520,16 +560,268 @@ private static boolean hasAnyInterceptorBindings(DetectedAiAgentBuildItem agent,
...
+     *         try {
+     *             Class[] var0 = new Class[] { CarInfo.class, Integer.class, FeedbackAnalysisResults.class, String.class };
+     *             METHOD_agent_analyzeForCondition_1 = CarConditionFeedbackAgent.class.getMethod("analyzeForCondition", var0);
+     *         } catch (Exception var2) {
+     *             throw (Throwable) (new RuntimeException("Failed to initialize agent method references", var2));
+     *         }
+     *     }
+     *
+     *     public CarConditions analyzeForCondition(CarInfo var1, Integer var2, FeedbackAnalysisResults var3, String var4) {
+     *         InvocationHandler var10 = (InvocationHandler) this.handler;
```
geoand: Is this correct? I don't see a `handler` field anywhere

mariofusco: Yes, it is. The `handler` is in the `AbstractQuarkusAgent` parent class that the generated agent extends.

geoand: Ah okay, I missed the extends part. thanks

agentic/runtime/src/main/java/io/quarkiverse/langchain4j/agentic/runtime/AbstractQuarkusAgent.java
```diff
@@ -0,0 +1,348 @@
...
+            METHOD_lastChatResponse = ChatMessagesAccess.class.getMethod("lastChatResponse", Object.class);
+            METHOD_removeLastResponseEvent = ChatMessagesAccess.class.getMethod("removeLastResponseEvent", Object.class);
+            METHOD_getEventClass = AiServiceListener.class.getMethod("getEventClass");
+            METHOD_onEvent = AiServiceListener.class.getMethod("onEvent", AiServiceEvent.class);
+            METHOD_getAgenticScope = AgenticScopeAccess.class.getMethod("getAgenticScope", Object.class);
+            METHOD_evictAgenticScope = AgenticScopeAccess.class.getMethod("evictAgenticScope", Object.class);
+        } catch (NoSuchMethodException e) {
+            throw new ExceptionInInitializerError(e);
+        }
+    }
```
geoand, line 69: Is all this reflection really needed? Is it needed all the time?  If we really do need it, can't we use `MethodHandle` instead?

mariofusco: The reflection is unfortunately necessary since both `AgentInvocationHandler` and `PlannerBasedInvocationHandler` only implement `InvocationHandler` + `InternalAgent`. Methods declared by other interfaces are all dispatched inside their `invoke()` method based on the Method object, they're not exposed as direct interface implementations and I wouldn't change this design for now.

Also I cannot use `MethodHandle` since `InvocationHandler.invoke(Object proxy, Method method, Object[] args)` requires a `java.lang.reflect.Method` as its second argument. The `invokeOnAgent` method passes these Method objects directly to the handler's `invoke()`, so they must remain `Method` instances. `MethodHandle` would only help if we were calling the methods directly, but here the handler's `invoke()` dispatches internally based on the `Method` it receives.

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
mariofusco: I renamed it to `agent`.

agentic/runtime/src/main/java/io/quarkiverse/langchain4j/agentic/runtime/AbstractQuarkusAgent.java
geoand: Okay, that's unfortunate

mariofusco: I could improve this, by changing the design of those langchain4j `InvocationHandler` implementation, but if you don't mind I would do this in a follow up work and leave out of scope of this pull request.

Except for this, if you're happy with this solution, please do a final review of the [langchain4j pull request](https://github.com/langchain4j/langchain4j/pull/5397), so if it's ok for you we could create another patch release and finalize also this one.

geoand: Sure yeah, let's make the improvements in steps

mariofusco:
@geoand this should be ready to be merged now.

geoand (approved):
Thanks!

mariofusco:
@geoand the failure seems to be unrelated, but please double check.

PR #2562, opened 2026-06-09
Add default ToolExecutionErrorHandler and way to override it.

patriot1burke:
This PR does two things:

* The default tool execution error handling eats the exception and doesn't even log the exception.  It is impossible to diagnosis problems thrown from a tool method.  So, I added a simple OOTB default to log the exception and to then perform the default behavior LC4J does.

* If devs want to override the default behavior they annotation a class with `@DefaultToolExecutionErrorHandler`

```java
@DefaulToolExecutionErrorHandler
public class MyHandler implements ToolExecutionErrorHandler {
...
}
```

See docs in PR for more information.


geoand (approved):
Thanks!

PR #2563, opened 2026-06-10
Allow the system message to depend on the model in use

### LINKED ISSUE #2466: Allow for System/User message (prompt) be based on LLM+model configuration
It would be useful to be able to have LLM+model specific messages / prompts.
e.g. one for OpenAI+gpt-4o, other for Claude+Opus, third for Gemini+xyz, Watson+abc

They would be stored in different txt files, similar to resource bundles (or similar to current `fromResource`, but more flexible)
e.g. mysysmsg_openai_gpt-40.txt

Or we could combine all LLM+model specific prompts in one property file, `openai_gpt-40.properties`:
```
mysysmsg=Do this and that
```

Where @SystemMessage would have a new property
```
@SystemMessage(fromLLMandModelPrefix="mysysmsg")
```
and if the combination is not found on classpath or properties, use previous mechanism (`fromResource` and `value`).

omatheusmesmo:
## Describe your changes

Introduces `SystemMessageProviderWithContext`, a system message provider that receives the `InvocationContext` and can inspect the model serving each call (`modelProvider()` and `defaultRequestParameters()`), so the system message can be chosen per model. It works for both blocking and streaming services and honors per-call `@ModelName` selection, since the context is built from the effective model.

It is a separate interface rather than a second method on `SystemMessageProvider`, so an implementation only writes the method it needs. Both interfaces extend the `BaseSystemMessageProvider` marker so the single `systemMessageProviderSupplier` attribute accepts either. `InvocationContext` also exposes `chatMemoryId()`, so the context-aware provider can do everything the memory-id one does, plus inspect the model.

Resolution precedence is `@SystemMessage` annotation > provider. Backward compatible: the existing `SystemMessageProvider`, with its single `getSystemMessage(Object)` method, is unchanged.

---

- Closes #2466

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ModelAwareSystemMessageProviderModelNameTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ModelAwareSystemMessageProviderModelNameTest.java
@@ -1,0 +1,138 @@
+package io.quarkiverse.langchain4j.test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.context.control.ActivateRequestContext;
+import jakarta.enterprise.inject.Produces;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.data.message.SystemMessage;
+import dev.langchain4j.invocation.InvocationContext;
+import dev.langchain4j.model.ModelProvider;
+import dev.langchain4j.model.chat.ChatModel;
+import dev.langchain4j.model.chat.request.ChatRequest;
+import dev.langchain4j.model.chat.request.ChatRequestParameters;
+import dev.langchain4j.model.chat.response.ChatResponse;
+import dev.langchain4j.service.UserMessage;
+import io.quarkiverse.langchain4j.ModelName;
+import io.quarkiverse.langchain4j.RegisterAiService;
+import io.quarkiverse.langchain4j.runtime.aiservice.SystemMessageProviderWithContext;
+import io.quarkus.arc.Unremovable;
+import io.quarkus.test.QuarkusUnitTest;
+
+public class ModelAwareSystemMessageProviderModelNameTest {
+
+    @RegisterExtension
+    static QuarkusUnitTest unitTest = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(
+                            Assistant.class,
+                            ModelAwareProvider.class,
+                            DefaultModelSupplier.class,
+                            NamedModelProducer.class,
+                            EchoModel.class));
+
+    @Inject
+    Assistant assistant;
+
+    @Test
+    @ActivateRequestContext
+    public void promptReflectsPerCallModel() {
+        assertThat(assistant.chat("Hello", "claude")).isEqualTo("provider=ANTHROPIC model=claude-opus-4-8");
+        assertThat(assistant.chat("Hello", "gpt")).isEqualTo("provider=OPEN_AI model=gpt-4o");
+    }
+
+    @RegisterAiService(chatLanguageModelSupplier = DefaultModelSupplier.class, chatMemoryProviderSupplier = RegisterAiService.NoChatMemoryProviderSupplier.class, systemMessageProviderSupplier = ModelAwareProvider.class)
+    public interface Assistant {
+        String chat(@UserMessage String userMessage, @ModelName String model);
+    }
+
+    @ApplicationScoped
+    public static class ModelAwareProvider implements SystemMessageProviderWithContext {
+
+        @Override
+        public Optional<String> getSystemMessage(InvocationContext context) {
+            return Optional.of("provider=" + context.modelProvider()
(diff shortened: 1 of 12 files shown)


geoand:
Thanks for this!

The CI failure does look related

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ModelAwareSystemMessageProviderModelNameTest.java
```diff
@@ -0,0 +1,143 @@
...
+        @Override
+        public ChatModel get() {
+            return new EchoModel(ModelProvider.MISTRAL_AI, "default-model");
+        }
+    }
+
+    @Singleton
+    public static class NamedModelProducer {
+
+        // @Unremovable: these beans are only resolved at runtime via @ModelName lookup, so Arc would otherwise remove them
```
geoand, line 84: This does not seem right. The Quarkus LangChain4j extension should make these beans unremovable, not the CDI producer itself

omatheusmesmo: I found this weird too, but without it didn''t work... I copy this solution from:

https://github.com/quarkiverse/quarkus-langchain4j/blob/a9d389e3d2650d1ea149208697d996ea1db66c13/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/guardrails/MultipleCDIBeansMatchGuardrailTest.java#L58

But I can try again

geoand: Sure, but that's a different case. My point is that any `ChatModel` needs to be made unremovable by the extension

omatheusmesmo: > Sure, but that's a different case. My point is that any `ChatModel` needs to be made unremovable by the extension

The `@Unremovable` was masking an edge case: the extension only marks `ChatModel` unremovable when the service injects the default model from CDI (`needsChatModelBean`), so with `chatLanguageModelSupplier` + a `@ModelName` parameter the named beans `effectiveChatModel` resolves at runtime get pruned. 

Should I drop/fix this test variant since the feature is already covered without it or `supplier + @ModelName` should be supported?

geoand: Let's just keep what you've done for now

omatheusmesmo: Are you sure? I can keep working on it.

geoand: Yeah, no problem

core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ModelAwareSystemMessageProviderStreamingTest.java
```diff
@@ -0,0 +1,122 @@
...
+        @Override
+        public Optional<String> getSystemMessage(Object memoryId) {
+            return Optional.of("OLD-PATH should not be used");
+        }
+
+        @Override
+        public Optional<String> getSystemMessage(InvocationContext context) {
+            return Optional.of("provider=" + context.modelProvider()
+                    + " model=" + context.defaultRequestParameters().modelName());
+        }
```
geoand: Requiring implementations to write both messages isn't ideal. That's why in LangChain4j we used a different interface

omatheusmesmo: Hi @geoand , I created a new interface, `SystemMessageProviderWithContext`, to use  `InvocationContext` specifically, please take a look and see if it's what you expected.

geoand:
@alesj PTAL

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/BaseSystemMessageProvider.java
```diff
@@ -0,0 +1,10 @@
+package io.quarkiverse.langchain4j.runtime.aiservice;
+
+/**
+ * Common super-type for the system message providers, so a single {@code @RegisterAiService}
+ * attribute can accept either a {@link SystemMessageProvider} or a {@link SystemMessageProviderWithContext}.
+ * <p>
+ * Implement one of the sub-interfaces, not this one directly.
+ */
+public interface BaseSystemMessageProvider {
```
alesj, line 9: Does it make sense to have an empty interface?

alesj:
> @alesj PTAL

Otherwise lgtm.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/BaseSystemMessageProvider.java
omatheusmesmo: The empty interface only exists so a single `@RegisterAiService` attribute can accept either provider.

`systemMessageProviderSupplier()` has one upper bound (`Class<? extends X>`), and I deliberately did not make `SystemMessageProviderWithContext extend SystemMessageProvider`, since that would force implementations to provide both methods (the thing we moved away from). `BaseSystemMessageProvider` is the shared super-type that lets the attribute accept either, and the concrete provider is selected at runtime via `instanceof` in the recorder.

The alternative would be to drop the marker and add a second attribute (e.g. `systemMessageProviderWithContextSupplier()`), but that adds a second public attribute plus a "both set" case to validate.

fedinskiy:
Hey, wanted to clarify things: this PR allows to change the system message depending on _model provider_ (eg OpenAI), not _the model_ (eg gpt-4.1-nano) in use, doesn't it ?

omatheusmesmo:
> Hey, wanted to clarify things: this PR allows to change the system message depending on _model provider_ (eg OpenAI), not _the model_ (eg gpt-4.1-nano) in use, doesn't it ?

Hi @fedinskiy , I just opened a PR with an automated test for your use case and updating the documentation. See #2743

PR #2565, opened 2026-06-10
Fix/mistral streaming tool call implementation

HakimMohammed:
## Problem
With `MistralAI` as a provider, tool/function calling does not work when the response is streamed. The model returned the tool call, but the tool was never executed.

## Root Cause
`QuarkusMistralAiClient#streamingChatCompletion` only accumulated text content from each delta and used it to build `AiMessage`. The `tool_calls` deltas were ignored, which resulted in `AiMessage` containing no `ToolExecutionRequests`.

## Fix
Gathered streamed `tool_calls` using `ToolCallBuilder`.

## Related
Builds on the streaming null-content NPE fix (#2553 ).

### DIFF
--- a/model-providers/mistral/runtime/src/main/java/io/quarkiverse/langchain4j/mistralai/QuarkusMistralAiClient.java
+++ b/model-providers/mistral/runtime/src/main/java/io/quarkiverse/langchain4j/mistralai/QuarkusMistralAiClient.java
@@ -1,5 +1,6 @@
 package io.quarkiverse.langchain4j.mistralai;
 
+import static dev.langchain4j.internal.Utils.isNotNullOrBlank;
 import static dev.langchain4j.model.mistralai.internal.mapper.MistralAiMapper.finishReasonFrom;
 import static dev.langchain4j.model.mistralai.internal.mapper.MistralAiMapper.tokenUsageFrom;
 import static java.util.stream.Collectors.joining;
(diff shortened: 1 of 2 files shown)


model-providers/mistral/runtime/src/main/java/io/quarkiverse/langchain4j/mistralai/QuarkusMistralAiClient.java
```diff
@@ -81,6 +85,9 @@ public void streamingChatCompletion(MistralAiChatCompletionRequest request,
         AtomicReference<StringBuffer> contentBuilder = new AtomicReference<>(new StringBuffer());
         AtomicReference<TokenUsage> tokenUsage = new AtomicReference<>();
         AtomicReference<FinishReason> finishReason = new AtomicReference<>();
+
+        ToolCallBuilder toolCallBuilder = new ToolCallBuilder(-1);
```
geoand, line 89: Why `-1`?

geoand:
Thanks for this!

I added a small question. 
Moreover, this needs to be rebased onto `main` as we don't allow `merge commits`

model-providers/mistral/runtime/src/main/java/io/quarkiverse/langchain4j/mistralai/QuarkusMistralAiClient.java
HakimMohammed: I followed the same approach as the ToolCallBuilder of Anthropic. I also don't know why

geoand: Okay, let's leave it as is then

geoand:
The test failure is related to the change

geoand:
Can you.please squash the commits?

HakimMohammed:
Commits squashed. Sorry about that

geoand:
NP, thanks!

geoand (approved):

PR #2584, opened 2026-06-11
Skip CI on draft pull requests

### LINKED ISSUE #2564: Run the PR CI conditionally for draft and conflicting PRs
Currently `build-pull-request.yml` triggers on every push to any open PR, including drafts and PRs with merge conflicts. There's no guard on `draft` or on the merge state.

The matrix is large (3 JDKs × all modules + native + in-process embedding + MCP conformance), so:

- Draft PRs run the full matrix on every commit while still being WIP.
- Conflicting PRs are worse: on a `pull_request` event `actions/checkout` checks out the merge ref (`refs/pull/N/merge`), which GitHub does not update when there are conflicts, so the run either uses a stale merge commit (misleading result) or the checkout fails.

`cancel-in-progress: true` already cancels obsolete runs on a new push, so the worst case is mitigated, but the draft/conflict cases still run.

Proposal:

- Draft: add `ready_for_review` to the trigger types and guard the heavy jobs with `if: !github.event.pull_request.draft`. This is reliable since `draft` is always present in the event payload.
- Conflicts: skip when the PR is not mergeable. This is less straightforward because `mergeable`/`mergeable_state` are computed asynchronously and are usually `null` on `synchronize`, so it needs an API check rather than a simple `if:`.

That said, running CI on drafts is intentional in some projects (WIP feedback), so this is a trade-off rather than a clear bug. Would you be open to gating drafts (and optionally conflicting PRs) here, or do you prefer keeping CI on every push?

omatheusmesmo:
* Closes #2564

### DIFF
--- a/.github/workflows/build-pull-request.yml
+++ b/.github/workflows/build-pull-request.yml
@@ -2,6 +2,7 @@ name: Build (on pull request)
 
 on:
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
     paths-ignore:
       - '.gitignore'
       - 'CODEOWNERS'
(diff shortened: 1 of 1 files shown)


geoand (approved):
Thanks!

PR #2583, opened 2026-06-11
Add cache token metrics

matheusandre1:
Closes: #2321

Issue #1899, opened 2025-10-28
Improve Cost Tracking & Observability


cescoffier:
* Track and optimize token usage
* Condense tool result JSON
* Estimate invocation cost during planning
* Support tracing (OpenTelemetry), logging, and metric collection (Micrometer/OTEL bridge) - already quite advanced
* Document techniques to control/minimize token usage, like TokenWindowChatMemory instead of MessageWindowChatMemory,
filter unnecessary MCP tools, select required tools dynamically via AiServices.toolSelector (to be implemented)

(From https://github.com/quarkusio/quarkus/discussions/50582)

PR #2585, opened 2026-06-12
Track prompt cache tokens in metrics and cost estimation

### LINKED ISSUE #2321: MetricsChatModelListener and CostEstimator do not support prompt cache tokens
## Problem

`MetricsChatModelListener` tracks `gen_ai.client.token.usage` for `input` and `output` token types, and `gen_ai.client.estimated_cost` via `CostEstimator`. However, when using Amazon Bedrock with prompt caching enabled, the response contains additional token counts in `BedrockTokenUsage`:

- `cacheWriteInputTokens()` — tokens written to the prompt cache
- `cacheReadInputTokens()` — tokens read from the prompt cache (cache hit)

These are **not tracked** by `MetricsChatModelListener` and **not accessible** through `CostEstimator.CostContext`.

### 1. MetricsChatModelListener — missing cache token metrics

`recordTokenUsage()` only reads `inputTokenCount()` and `outputTokenCount()`:

```java
// MetricsChatModelListener.java
private void recordTokenUsage(ChatModelResponseContext responseContext, Tags tags) {
    TokenUsage tokenUsage = responseContext.chatResponse().tokenUsage();
    // ...
    Integer inputTokenCount = tokenUsage.inputTokenCount();  // ✅
    Integer outputTokenCount = tokenUsage.outputTokenCount(); // ✅
    // cacheWriteInputTokens — ❌ not tracked
    // cacheReadInputTokens  — ❌ not tracked
}
```

Cache tokens have different pricing (cache reads are ~90% cheaper than regular input), so tracking them separately is essential for cost monitoring.

### 2. CostEstimator.CostContext — no cache token fields

The `CostContext` interface only exposes:

```java
interface CostContext extends SupportsContext {
    Integer inputTokens();
    Integer outputTokens();
    // no cacheWriteInputTokens()
    // no cacheReadInputTokens()
}
```

This makes it impossible for `CostEstimator` implementations to calculate cache-specific costs. The `CostResult` record also only has `inputTokensCost` and `outputTokensCost` — no fields for cache costs.

### MetricsChatModelListener should:
1. Check if `TokenUsage` is `instanceof BedrockTokenUsage`
2. Record `gen_ai.client.token.usage` with `gen_ai.token.type=cache_write` and `cache_read`
3. Record `gen_ai.client.estimated_cost` for cache token types

### CostEstimator should:
1. Expose cache token counts in `CostContext` (e.g. `cacheWriteInputTokens()`, `cacheReadInputTokens()`)
2. Add cache cost fields to `CostResult` (e.g. `cacheWriteTokensCost`, `cacheReadTokensCost`)

Or alternatively, make `CostContext` generic enough to support provider-specific token types.

## Current Workaround

We implemented a separate `ChatModelListener` CDI bean that:
1. Casts `TokenUsage` to `BedrockTokenUsage` to access cache fields
2. Records cache token usage counters following GenAI semantic conventions
3. Calculates cache costs using configurable pricing

This works but duplicates tag-building logic from `MetricsChatModelListener` and requires separate cost configuration outside of the `CostEstimator` SPI.

## Environment

- quarkus-langchain4j: 1.8.2
- langchain4j-bedrock: 1.12.2

omatheusmesmo:
## Describe your changes

Adds a `CacheTokenUsageExtractor` SPI in the core so any provider can expose its prompt cache token counts without coupling the core to a provider. Extractors are discovered at build time via `@All`. `MetricsChatModelListener` uses them to record `cache_read`/`cache_creation` Micrometer counters; `CostEstimator` gets the full `ChatModelResponseContext` and can resolve cache tokens through the same SPI to compute cache costs.

Implementations are provided for Bedrock, Anthropic, OpenAI and Gemini. Anthropic reports both cache read and cache creation tokens; OpenAI and Gemini report only cached (read) input tokens, so their cache-creation slot stays null. A native integration test on the OpenAI module resolves an in-process `OpenAiTokenUsage` through the resolver, exercising bean discovery and the base-class reflection in native image without calling the API.

Supersedes #2583 , therefore co-authored by @matheusandre1 too.

I believe this is part of the cost & observability scope of the Agentic working group.

---

- Closes #2321
- Relates to #1899


geoand:
Thanks a lot for this!

Wouldn't it be simpler to have `CostEstimator$SupportsContext` expose the entire `ChatModelResponseContext` and let `CostEstimator` (which is already provider-specific) deal with it?

omatheusmesmo:
> Thanks a lot for this!
> 
> Wouldn't it be simpler to have `CostEstimator$SupportsContext` expose the entire `ChatModelResponseContext` and let `CostEstimator` (which is already provider-specific) deal with it?

It's the alternative the issue mentions too. I went with it: `SupportsContext` now exposes the full `ChatModelResponseContext`, so a (provider-specific) `CostEstimator` reads the cache tokens directly.

I kept the `CacheTokenUsageExtractor` SPI only for `MetricsChatModelListener`, which is built-in and has no user hook: without it the core would be back to `instanceof` per provider (or reflection) just to record the counters. The estimator can inject the same `CacheTokenUsageResolver`, so the extraction stays in one place.

geoand:
> It's the alternative the issue mentions too. I went with it:

Thanks. 

> I kept the CacheTokenUsageExtractor SPI only for MetricsChatModelListener

I still don't understand this. Wouldn't exposing `ChatModelResponseContext` from `CostContext` make ^ redundant?

omatheusmesmo:
> > It's the alternative the issue mentions too. I went with it:
> 
> Thanks.
> 
> > I kept the CacheTokenUsageExtractor SPI only for MetricsChatModelListener
> 
> I still don't understand this. Wouldn't exposing `ChatModelResponseContext` from `CostContext` make ^ redundant?


I think it wouldn't make it redundant, because the listener is a different extension point. The cache counters are emitted by `MetricsChatModelListener` (a `ChatModelListener`, not a `CostEstimator`), which reads the `TokenUsage` directly and never sees a `CostContext`. So exposing the context there doesn't reach it. To read the provider-specific cache fields from the core without `instanceof`/reflection, the listener needs the neutral resolver.

`SpanChatModelListener` is the same case: another core `ChatModelListener` that already reads `tokenUsage` for span attributes, also not a `CostEstimator`. So the SPI serves the observability listeners, and exposing the context on `CostContext` serves estimators. The two are different.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/MetricsChatModelListener.java
```diff
@@ -54,6 +61,16 @@ public MetricsChatModelListener(CostEstimatorService costEstimatorService) {
...
+        this.cacheReadInputTokenUsage = Counter.builder("gen_ai.client.token.usage")
+                .description("Measures number of input tokens read from the prompt cache")
+                .tag("gen_ai.operation.name", "chat")
+                .tag("gen_ai.token.type", "cache_read")
+                .withRegistry(Metrics.globalRegistry);
+        this.cacheWriteInputTokenUsage = Counter.builder("gen_ai.client.token.usage")
+                .description("Measures number of input tokens written to the prompt cache")
+                .tag("gen_ai.operation.name", "chat")
+                .tag("gen_ai.token.type", "cache_write")
+                .withRegistry(Metrics.globalRegistry);
```
geoand, line 73: So this is adding the same metric (i.e. `gen_ai.client.token.usage`) with different tags, right? If so, this is just wrong and will fail with the Prometheus registry IIRC

geoand:
> > > It's the alternative the issue mentions too. I went with it:
> > 
> > 
> > Thanks.
> > > I kept the CacheTokenUsageExtractor SPI only for MetricsChatModelListener
> > 
> > 
> > I still don't understand this. Wouldn't exposing `ChatModelResponseContext` from `CostContext` make ^ redundant?
> 
> I think it wouldn't make it redundant, because the listener is a different extension point. The cache counters are emitted by `MetricsChatModelListener` (a `ChatModelListener`, not a `CostEstimator`), which reads the `TokenUsage` directly and never sees a `CostContext`. So exposing the context there doesn't reach it. To read the provider-specific cache fields from the core without `instanceof`/reflection, the listener needs the neutral resolver.
> 
> `SpanChatModelListener` is the same case: another core `ChatModelListener` that already reads `tokenUsage` for span attributes, also not a `CostEstimator`. So the SPI serves the observability listeners, and exposing the context on `CostContext` serves estimators. The two are different.

As I don't see `CacheTokenUsage` being useful outside the cost context, I really don't want to introduce a specific SPI for it. 
That said, I may be misunderstanding the situation here.

omatheusmesmo:
> As I don't see `CacheTokenUsage` being useful outside the cost context, I really don't want to introduce a specific SPI for it. That said, I may be misunderstanding the situation here.

The metrics are why the SPI exists. The issue asks `MetricsChatModelListener` to record cache tokens, but `core/runtime` has no provider dependency, so the listener can't `instanceof BedrockTokenUsage`. The only ways for the core to read a provider-specific cache field are reflection ( #2583 ) or a build-time SPI. Exposing `ChatModelResponseContext` only fixes the cost side (a `CostEstimator` is provider-specific and casts the `TokenUsage` itself); it never reaches the listener, which has no `CostContext`.

On Prometheus: it won't fail. The constraint is that meters with the same name must share the same tag *keys*, not the same values. These counters use the same key set as the `input`/`output` ones just above (54-63), recorded with the same `tags`; only the `gen_ai.token.type` value differs. If this failed, `input`/`output` would already fail today.

If you'd rather not have the SPI at all, I can drop the metrics part, but then #2321 stays unresolved (cost would still work through the context).

P.S. You're right to question the tag names, though, just not for the Prometheus reason: `cache_read`/`cache_write` isn't the GenAI convention. The spec defines only `input`/`output` for `gen_ai.token.type` and models cache as dedicated attributes (`gen_ai.usage.cache_read.input_tokens` / `cache_creation.input_tokens`, a subset of input), with the metric form still open in https://github.com/open-telemetry/semantic-conventions-genai/issues/23. I took the names from the issue and I'm happy to align them.

geoand:
I took a closer look at this and have two comments:

1) Shouldn't total cost reflect the cached tokens?
2) Wouldn't something like  [this](https://github.com/user-attachments/files/28951271/cache-token-usage-tracking.patch) make things better?

omatheusmesmo:
> I took a closer look at this and have two comments:
> 
> 1. Shouldn't total cost reflect the cached tokens?
> 2. Wouldn't something like  [this](https://github.com/user-attachments/files/28951271/cache-token-usage-tracking.patch) make things better?

Thanks, I applied your patch. I kept it as-is except for two things: renamed `cache_write` to `cache_creation` to align with the GenAI convention (`gen_ai.usage.cache_creation.input_tokens`) while the API is still experimental, and made `resolve()` return `CacheTokenUsage.none()` in every branch so it never returns `null`. 

On your first point: yes, `CostEstimatorService.totalCost()` adds `cacheReadTokensCost` and `cacheCreationTokensCost` on top of input+output, and `CostEstimatorCacheCostTest` asserts the sum (input 100 + output 50 + cacheRead 30 + cacheCreation 12 = 192). Whether the cached portion is discounted from the input price is up to each estimator, since cache is a subset of input.

geoand (commented):
Thanks!

Can we add implementations of `CacheTokenUsageExtractor` for the models we support?

omatheusmesmo:
> Thanks!
> 
> Can we add implementations of `CacheTokenUsageExtractor` for the models we support?

Done. Added `CacheTokenUsageExtractor` implementations for Anthropic, OpenAI and Gemini. Anthropic reports both cache read and cache creation tokens; OpenAI and Gemini report only cached (read) input tokens, so their cache-creation slot stays `null`.

geoand (approved):
Thanks!

omatheusmesmo:
Hi @geoand , did you notice there is something off with the CI on the last commit? AFAICS the CI didn't run completely... I don't know if it is some bug on GitHub UI, but I can't investigate this today either. It should build and test all modules, right?





https://github.com/quarkiverse/quarkus-langchain4j/actions/runs/27688401135

---

> Sent from my smartphone.

geoand:
You're right... Not sure what the issue is, but I'll keep an eye out

PR #2589, opened 2026-06-12
Support @ToolBox annotation also for agents

mariofusco:


agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
```diff
@@ -85,6 +86,8 @@ public class AgenticProcessor {
 
     private static final Logger log = Logger.getLogger(AgenticProcessor.class);
 
+    private static final DotName TOOLBOX = DotName.createSimple("io.quarkiverse.langchain4j.ToolBox");
```
geoand: Don't we have access the actual class here? Or even better,  can't we use the constant from the core module?

geoand (commented):
Thanks!

Just added a small comment

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
mariofusco: Fixed, please check again.

geoand (approved):

PR #2604, opened 2026-06-15
Tool-calling support for GPULlama3

orionpapadakis:
### Description
This PR adds tool-calling support to the GPULlama3 chat models.

The implementation wires LangChain4j tool specifications into the GPULlama3 prompt format, supports model-emitted tool call extraction, and returns `ToolExecutionRequest` responses. It also handles tool execution result messages so follow-up model turns can synthesize final answers from tool output.

### Main Changes
- Add tool-aware prompt construction for GPULlama3.
- Support tool call extraction for non-streaming and streaming chat models.
- Preserve thinking content while separating it from final responses.
- Add support for multiple tool calls in a single response.
- Add tool result encoding back into the conversation.
- Add configurable model options used by the GPULlama3 tool-calling flow:
      - `enable-thinking`
      - `prefill-decode`
      - `prefill-batch-size`
      - `device-memory`

### DIFF
--- a/model-providers/gpu-llama3/runtime/src/main/java/io/quarkiverse/langchain4j/gpullama3/runtime/config/ChatModelConfig.java
+++ b/model-providers/gpu-llama3/runtime/src/main/java/io/quarkiverse/langchain4j/gpullama3/runtime/config/ChatModelConfig.java
@@ -1,5 +1,6 @@
 package io.quarkiverse.langchain4j.gpullama3.runtime.config;
 
+import java.util.Optional;
 import java.util.OptionalDouble;
 import java.util.OptionalInt;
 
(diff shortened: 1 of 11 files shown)


orionpapadakis:
I turned this to draft temporarily as we first need to merge a PR in GPULlama3 side in order to get the CI passing. I will turn it back to ready as soon as possible.

geoand:
Nice!

orionpapadakis:
we released a new GPULlama version (`0.5.0`) that supports tool-calling (from the engine side) and improved performance, especially for Qwen3 models that can handle tool-calling sequences more effectivelly than tiny Llama models (i.e. 1B). @geoand thank you for your patience, now its ready for review

geoand:
Great! Can you please rebase squash the commits and rebase onto `main`?

orionpapadakis:
@geoand sure, I will do it. we also need to publish a new tornado version to mvn central, the team is working on it

geoand:
👍🏽

model-providers/gpu-llama3/runtime/src/main/java/io/quarkiverse/langchain4j/gpullama3/runtime/config/ChatModelConfig.java
```diff
@@ -38,4 +38,45 @@ public interface ChatModelConfig {
...
+    @WithDefault("true")
+    boolean enableThinking();
+
+    /**
+     * Amount of GPU memory TornadoVM is allowed to allocate e.g. {@code "7GB"}, {@code "512MB"}.
+     * <p>
+     * Note: this maps to the JVM-global engine flag {@code tornado.device.memory}, so when
+     * multiple models are configured the first one to initialize wins.
+     */
+    @WithDefault("4GB")
```
geoand: Are you sure we should be setting this by default?

model-providers/gpu-llama3/runtime/src/main/java/io/quarkiverse/langchain4j/gpullama3/runtime/config/ChatModelConfig.java
```diff
@@ -38,4 +38,45 @@ public interface ChatModelConfig {
...
+
+    /**
+     * Whether to enable the model's thinking/reasoning phase. Only models with a thinking mode
+     * (e.g. Qwen3) honor this; for other models it has no effect.
+     * <p>
+     * When {@code true} (the upstream Qwen3 default), the model reasons inside
+     * {@code <think>…</think>} before answering, which tends to improve tool-calling decisions.
+     * Set to {@code false} for faster responses by skipping the reasoning phase.
+     */
+    @WithDefault("true")
```
geoand: I am pretty sure thinking is off by default in the other model providers

model-providers/gpu-llama3/runtime/src/main/java/io/quarkiverse/langchain4j/gpullama3/runtime/config/ChatModelConfig.java
orionpapadakis: yes, this option is crucial to accommodate models of different sizes, the default value in TornadoVM is 4GB, so we aim for the same transparency from quarkus config as well. Regarding the value (4GB) after experimentation we think is reasonable for nowadays GPUs

model-providers/gpu-llama3/runtime/src/main/java/io/quarkiverse/langchain4j/gpullama3/runtime/config/ChatModelConfig.java
orionpapadakis: indeed, this should be false by default

model-providers/gpu-llama3/runtime/src/main/java/io/quarkiverse/langchain4j/gpullama3/runtime/config/ChatModelConfig.java
geoand: In that case, I would make this configuration option `Optional` and use `@ConfigDocDefault` to mention what the default is.

orionpapadakis: done

geoand:
The MCP failures are unrelated

geoand (approved):

PR #2614, opened 2026-06-16
Add a test case demonstrating how guardrails work with agents

mariofusco:


geoand (approved):
Nice!

mariofusco:
It seems that those MCP tests failures are unrelated.

PR #2634, opened 2026-06-26
Avoid requiring a chat model for agentic systems that don't need it

### LINKED ISSUE #2630: A model is required for all use cases when using agentic module
I want to use quarkus agent stuff, but only non-LLM agents and A2A. So I added the quarkus agentic dependency and the a2a langchain4j dependency, but when I start the app, I get

```
Caused by: io.quarkus.runtime.configuration.ConfigurationException: A ChatModel or StreamingChatModel bean was requested, but no langchain4j providers were configured. Consider adding an extension like 'quarkus-langchain4j-openai'
```

cc @mariofusco

mariofusco:
- Closes: https://github.com/quarkiverse/quarkus-langchain4j/issues/2630


geoand (approved):
Thanks!

mariofusco:
At the moment I have no idea why that new `A2AOnlySequenceWithNoModelTest` test is failing. It works running a `mvn clean install` on my machine. I will investigate.

mariofusco:
It should be ok now.

PR #2636, opened 2026-06-26
feat: add LanceDB embedding store extension

### LINKED ISSUE #696: Add LanceDB EmbeddingStore integration
LanceDB is a high-performance, embedded vector database optimized for AI/LLM scenarios, featuring disk persistence, low latency, and seamless cross-language (Java/Python/Node.js) data sharing. It’s widely used in lightweight RAG, edge AI, and knowledge base scenarios, filling the gap for embedded vector storage in Java ecosystem.
Currently, LangChain4j supports mainstream vector databases like Milvus, PGVector, and RedisSearch, but lacks native LanceDB integration. Adding this will:
Expand the embedded vector storage options for Java developers;
Enable seamless RAG/knowledge base development with LanceDB in Spring Boot/Quarkus projects;
Align with Python LangChain’s LanceDB integration for consistent cross-language experience.
I’m willing to contribute to the implementation (including core EmbeddingStore adapter, unit tests, and documentation) if the community approves this feature.
Looking forward to your feedback!

JeffersonSantos29:
### Description

This PR introduces a complete Quarkus extension for integrating [LanceDB](https://lancedb.com/) as an embedding store within the `quarkus-langchain4j` ecosystem. It follows the established project patterns (deployment/runtime split) and supports both LanceDB Cloud and Enterprise.

**Key Features Implemented:**
* **Core Store:** `LanceDbEmbeddingStore` implementation using Apache Arrow IPC for data interchange. 
* **Metadata Filtering:** Full translation of LangChain4j `Filter` objects to LanceDB SQL predicates, covering all comparison and logical operators, applied as pre-filters.
* **Distance Metrics:** Support for `l2` (Euclidean), `cosine`, and `ip` (Inner Product), with accurate distance-to-relevance-score conversion.
* **Named Stores:** Support for multiple independent LanceDB stores via the `@EmbeddingStoreName` qualifier, alongside the default store.

**Structure & Changes:**
* Added `embedding-stores/lancedb` module containing deployment and runtime extensions.
* Added `integration-tests/rag-lancedb` demonstrating multi-tenant RAG with metadata filtering (gated by `enable.lancedb-tests`).
* Fully documented in `rag-lancedb-store.adoc`.
* Updated root, `embedding-stores`, and `integration-tests` POMs to include the new modules and versioning (`0.30.0-final.0`).

**Test Coverage:**
* **Unit Tests:** 28 test cases for `LanceDbFilterMapper` ensuring robust SQL predicate generation and escaping.
* **Deployment Tests:** Coverage for CDI injection, named vs. default stores, and configuration validation (`ConfigValidationException` for missing keys).
* **Integration Tests:** End-to-end RAG tests verifying per-user document isolation.

I have verified that all tests pass locally. Looking forward to your feedback!

- Resolves: langchain4j/langchain4j-community#696


JeffersonSantos29:
The documentation check should be fixed now. I noticed the Native GraalVM tests are failing with exit code 1 during image generation, highly likely due to missing JNI/Resource registrations for Apache Arrow and LanceDB's underlying Rust bindings.

Before I dive deep into fixing the GraalVM native build, I wanted to ask: should the core implementation (LanceDbEmbeddingStore and LanceDbFilterMapper) be ported to the langchain4j-community repository [first](https://github.com/langchain4j/langchain4j-community/pull/710) ? If so, once merged, adjust this PR to only contain the Quarkus wrapper and the proper native image registrations. Let me know how you prefer to proceed!

geoand:
Thanks for this, very neat!

> hould the core implementation (LanceDbEmbeddingStore and LanceDbFilterMapper) be ported to the langchain4j-community repository first?

Yeah, that's usually the way things are done

geoand:
The CI failure seems legitimate

jmartisk:
I'll have a deeper look at this ASAP

integration-tests/rag-lancedb/src/main/resources/application.properties
```diff
@@ -0,0 +1,7 @@
+quarkus.langchain4j.embedding-model.provider=dev.langchain4j.model.embedding.onnx.allminilml6v2.AllMiniLmL6V2EmbeddingModel
+quarkus.langchain4j.openai.api-key=blabla
+
+quarkus.langchain4j.lancedb.api-key=${LANCEDB_API_KEY}
```
jmartisk: So, this means that if we want to have this run in the CI, we would need to get and manage our own API key right? Isn't there a way to do something for free and without depending on external services, like a local deployment via Docker?

JeffersonSantos29: Thanks for looking into this, @jmartisk — that's a fair concern, and I dug into it a bit before replying.

I believe this might not be something we can easily work around on our side: from what I found, the official LanceDB Java SDK currently only supports remote connections to LanceDB Cloud/Enterprise via the REST Namespace API. Local/embedded database support (the mode the Python and JS/TS SDKs seem to already have, where you point at a local directory with no API key) looks like it's still a work in progress upstream — I found it tracked at [lancedb/lancedb#2848](https://github.com/lancedb/lancedb/issues/2848), though I could be missing more recent developments there.

If that's accurate, it would also rule out a Docker-based local deployment, since LanceDB doesn't seem to follow a client/server model like Milvus or Qdrant — there's no "LanceDB server" image to run, as far as I can tell. The local mode appears to be an in-process embedded library, which seems to be exactly the part the Java client doesn't expose yet.

Given that, I'd tentatively suggest:

Keeping integration-tests/rag-lancedb gated behind enable.lancedb-tests (as it already is), so it stays opt-in and doesn't require a key for the default CI build — this seems similar to how Pinecone is handled here, which has a comparable cloud-only constraint, though I haven't checked that in detail.
Possibly revisiting this once #2848 lands upstream, switching the test to a local/embedded connection at that point.

I could be wrong about some of this, so let me know if you're aware of another option I'm missing. Happy to add a note about it in rag-lancedb-store.adoc if this reading is correct.

jmartisk: Ok I kinda dislike these cloud-only services where you depend on an external service and you can't run your own instance, having tests for this in the CI pipeline always creates loads of all kids of problems (when the API changes, when the service is down, and especially when you use a single shared database for all CI runs, meaning that the CI  can't run multiple times in parallel, and that one run could leave behind some mess which causes further runs to fail.... which is a true horror). So, ok, then let's leave it as opt-in for the CI now.

I'll try to have a deeper look at the PR code  in a bit.

embedding-stores/lancedb/runtime/src/main/java/io/quarkiverse/langchain4j/lancedb/LanceDbEmbeddingStore.java
```diff
@@ -0,0 +1,480 @@
...
+        }
+    }
+
+    @Override
+    public void remove(String id) {
+        ensureTableCreated();
+        try {
+            DeleteFromTableRequest request = new DeleteFromTableRequest();
+            request.setId(Arrays.asList(tableName));
+            request.setPredicate(ID_COLUMN + " = '" + id + "'");
```
jmartisk: This smells with SQL injection, we probably should  validate that the `id` is a UUID and not something like `' OR '1'='1`

embedding-stores/lancedb/runtime/src/main/java/io/quarkiverse/langchain4j/lancedb/LanceDbEmbeddingStore.java
```diff
@@ -0,0 +1,480 @@
...
+    private Schema createSchema(int dim) {
+        return new Schema(Arrays.asList(
+                new Field(ID_COLUMN, FieldType.nullable(new ArrowType.Utf8()), null),
+                new Field(VECTOR_COLUMN,
+                        FieldType.nullable(new ArrowType.FixedSizeList(dim)),
+                        Arrays.asList(new Field("item",
+                                FieldType.nullable(new ArrowType.FloatingPoint(FloatingPointPrecision.SINGLE)),
+                                null))),
+                new Field(TEXT_COLUMN, FieldType.nullable(new ArrowType.Utf8()), null),
+                new Field(METADATA_COLUMN, FieldType.nullable(new ArrowType.Binary()), null)));
```
jmartisk: Not that I'm an expert in LanceDB, and I didn't run the integration tests either, so I can't really verify, but Claude says this:

>  Metadata is stored as a raw binary (JSON bytes) in a single Binary Arrow column. However, LanceDbFilterMapper generates SQL predicates like metadata.name = 'John'. LanceDB's SQL filter engine operates on Arrow struct fields, not on JSON blobs stored in a
  Binary column. The generated predicates will either fail at runtime or silently match nothing. To make filters work, metadata must be stored as a proper Arrow Struct type with named sub-fields, or serialized to a top-level column per metadata key.
  This means the primary advertised feature (metadata filtering) is broken.

embedding-stores/lancedb/runtime/src/main/java/io/quarkiverse/langchain4j/lancedb/LanceDbEmbeddingStore.java
```diff
@@ -0,0 +1,480 @@
...
+            FixedSizeListVector vectorVector = (FixedSizeListVector) root.getVector(VECTOR_COLUMN);
+            Float4Vector vectorData = (Float4Vector) vectorVector.getDataVector();
+            VarCharVector textVector = (VarCharVector) root.getVector(TEXT_COLUMN);
+            VarBinaryVector metadataVector = (VarBinaryVector) root.getVector(METADATA_COLUMN);
+
+            int rowCount = ids.size();
+            root.setRowCount(rowCount);
+
+            for (int i = 0; i < rowCount; i++) {
+                idVector.setSafe(i, ids.get(i).getBytes());
```
jmartisk:
```suggestion
                idVector.setSafe(i, ids.get(i).getBytes(StandardCharsets.UTF_8));
```

embedding-stores/lancedb/runtime/src/main/java/io/quarkiverse/langchain4j/lancedb/LanceDbEmbeddingStore.java
```diff
@@ -0,0 +1,480 @@
...
+                VarBinaryVector metadataVector = (VarBinaryVector) root.getVector(METADATA_COLUMN);
+
+                Float4Vector distanceVector = null;
+                try {
+                    distanceVector = (Float4Vector) root.getVector("_distance");
+                } catch (IllegalArgumentException ignored) {
+                }
+
+                for (int row = 0; row < root.getRowCount(); row++) {
+                    String id = new String(idVector.get(row));
```
jmartisk:
```suggestion
                    String id = new String(idVector.get(row), StandardCharsets.UTF_8);
```

embedding-stores/lancedb/runtime/src/main/resources/META-INF/native-image/io.quarkiverse.langchain4j/quarkus-langchain4j-lancedb/native-image.properties
```diff
@@ -0,0 +1 @@
+Args = --add-opens=java.base/java.nio=ALL-UNNAMED --initialize-at-build-time=io.netty.buffer.PooledByteBufAllocatorL
```
jmartisk, line 1: Is this file really necessary? I don't think it is?

jmartisk (commented):
I've posted a few comments to the code. But what is more important - our team doesn't have any LanceDB subscription and we're probably not going to spend money and time on this. We don't have the capacity to maintain a huge amount of embedding store implementations. I guess that means that we can accept this contribution, but it will mostly fall on you to maintain it. I want to make sure you're fully aware of the consequences and you're fine with it.

By the way, is there a specific reason why you're not adding this to langchain4j directly? That is actually the preferred pattern (at least for embedding stores) - langchain4j itself has the implementation, the quarkus extension just provides a thin wrapper around it to integrate it.

JeffersonSantos29:
Hi @jmartisk, thanks a lot for the detailed review!

Quick update: the core implementation (LanceDbEmbeddingStore and LanceDbFilterMapper) has just been merged into langchain4j-community in #710, including full metadata filtering test coverage via EmbeddingStoreWithFilteringIT. So I'll rework this PR to be a thin Quarkus wrapper around that upstream implementation instead of keeping a separate copy here — which directly addresses the metadata filtering inconsistency you flagged, since that logic will no longer live in this module.

I'll go ahead and adjust everything you requested:

Remove the local LanceDbEmbeddingStore/LanceDbFilterMapper code and depend on the langchain4j-community-lancedb artifact instead (this also removes the SQL injection risk on delete and the charset issues you pointed out, since that code won't exist here anymore).

Re-check whether native-image.properties is still needed once the dependency changes, and update the native image registrations accordingly.

Keep the integration test opt-in behind enable.lancedb-tests, as already agreed.

One more thing worth mentioning: the upstream tests actually connect via a local "dir" LanceNamespace (a filesystem-based namespace, no API key involved) rather than the Cloud/Enterprise REST API. What do you think about using that same local mode for our CI integration tests here, instead of keeping them opt-in only due to the API key concern?

jmartisk:
Cool! 

> One more thing worth mentioning: the upstream tests actually connect via a local "dir" LanceNamespace (a filesystem-based namespace, no API key involved) rather than the Cloud/Enterprise REST API. What do you think about using that same local mode for our CI integration tests here, instead of keeping them opt-in only due to the API key concern?

That sounds great :)

JeffersonSantos29:
> That sounds great :)

 As agreed above, this PR was refactored to be a thin wrapper over the upstream LanceDbEmbeddingStore from langchain4j-community-lancedb instead of keeping a local copy of the implementation.

 The catch is that the LanceDB store (langchain4j-community #710) has not been included in any published langchain4j-community release yet — it's currently only available as a -SNAPSHOT. Because of that, the CI Quick Build / Documentation check jobs fail: mvn install can't resolve dev.langchain4j:langchain4j-community-lancedb from a stable repository.

 So I'd suggest we hold this PR until the next langchain4j-community release that ships the LanceDB module is available on Maven Central. Once it's out, I'll pin the dependency to that release version (dropping the SNAPSHOT), and the build should go green.

 I'll keep an eye on the upstream release and ping here as soon as we can move forward.

PR #2639, opened 2026-06-26
Add Workload ModelAuthProvider, update vanilla OpenAI and Anthropic model providers to support ModelAuthProvider

### LINKED ISSUE #2571: Create Workload Identity Federation `ModelAuthProvider` extension
Anthropic, OpenAI support Workload Identity Federation:

https://platform.claude.com/docs/en/manage-claude/workload-identity-federation
https://developers.openai.com/api/docs/guides/workload-identity-federation

The LLM providers are moving away from static API keys, so we should let users use short lived access tokens.

Note we already support it in some flows with Gemini and Azure where a user token is authorized, but WIF offers a new approach.

The extension will pick up a token (SPIFFE JWT SVID or other JWT) and use oidc client to exchange it against the configured model provider's WIF endpoint and that should be all that is needed.

I'll have a look soon

sberyozkin:
This PR does the following:

*  Introduces a `WorkloadModelAuthProvider` (it copies most of the code done by @michalvavrik in Quarkus to support JWT bearer and Spiffe token loading from the disk for the OIDC client authentication)
* Updates vanilla OpenAI provider to recognize `ModelAuthProvider` - my initial demo idea was to demo it with [OpenAI WIF](https://developers.openai.com/api/docs/guides/workload-identity-federation), but unfortunately it just does not allow to complete the configuration related to service accounts - but vanilla OpenAI should still support `ModelAuthProvider`
* Updates Anthropic provider to recognize `ModelAuthProvider`
* And adds two demo that show how Quarkus LangChain4j payloads can access Anthropic models from Openshift with 1) Spiffe (supported by Red Hat Zero Trust WIF Manager, with the work from Andrew Block @sabre1041 proving of great help) and 2) Kubernetes projected tokens for simpler single cluster setups. 
 
I thought of combining 2 demos into one, but it would make it a more complex demo in terms of varying deployment options - IMHO it should make sense to keep both Spiffe and K8s projected tokens demoed separately. If we were to add only a single demo, we should choose Spiffe as it is very topical and works across clusters.

I used Claude to help me with prepare various deployment files in particular.
I also tried pure Spire deployment with Helm charts but it is very difficult, not worth doing it in the demo.

Besides vanilla OpenAi and Anthropic, azure-openai and gemini providers which already support `ModelAuthProvider` can use it.

Once @michalvavrik introduces a Quarkus extension for accessing Spire agent directly with Spiffe API, we can introduce a Spiffe centric `ModelAuthProvider` and avoid storing Spiffe tokens to the disk, with no need for a sidecar.

MCP Client support is to be added later

- Closes: #2571

### DIFF
--- a/model-auth-providers/workload-model-auth-provider/deployment/src/main/java/io/quarkiverse/langchain4j/workload/deployment/WorkloadModelAuthProviderBuildConfig.java
+++ b/model-auth-providers/workload-model-auth-provider/deployment/src/main/java/io/quarkiverse/langchain4j/workload/deployment/WorkloadModelAuthProviderBuildConfig.java
@@ -22,4 +22,11 @@ public interface WorkloadModelAuthProviderBuildConfig {
      * If not set, the default OIDC client is used.
      */
     Optional<String> oidcClientName();
+
+    /**
+     * The model provider name to bind this model authentication provider to.
+     * When set, the provider is qualified with {@link io.quarkiverse.langchain4j.ModelName} and only used
+     * by the matching model provider. When not set, the provider is registered as a global default.
+     */
+    Optional<String> modelName();
 }
--- a/model-auth-providers/workload-model-auth-provider/deployment/src/main/java/io/quarkiverse/langchain4j/workload/deployment/WorkloadModelAuthProviderBuildConfig.java
+++ b/model-auth-providers/workload-model-auth-provider/deployment/src/main/java/io/quarkiverse/langchain4j/workload/deployment/WorkloadModelAuthProviderBuildConfig.java
@@ -1,0 +1,25 @@
+package io.quarkiverse.langchain4j.workload.deployment;
+
+import static io.quarkus.runtime.annotations.ConfigPhase.BUILD_TIME;
+
+import java.util.Optional;
+
+import io.quarkus.runtime.annotations.ConfigDocDefault;
+import io.quarkus.runtime.annotations.ConfigRoot;
+import io.smallrye.config.ConfigMapping;
+
+@ConfigRoot(phase = BUILD_TIME)
+@ConfigMapping(prefix = "quarkus.langchain4j.workload-model-auth-provider")
+public interface WorkloadModelAuthProviderBuildConfig {
+    /**
+     * Whether the Workload Identity Federation ModelAuthProvider should be enabled
+     */
+    @ConfigDocDefault("true")
+    Optional<Boolean> enabled();
+
+    /**
+     * The name of the OIDC client to use for the token exchange.
+     * If not set, the default OIDC client is used.
+     */
+    Optional<String> oidcClientName();
+}
--- a/model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
+++ b/model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
@@ -31,6 +31,7 @@ import io.quarkiverse.langchain4j.anthropic.QuarkusAnthropicClient;
 import io.quarkiverse.langchain4j.anthropic.runtime.config.ChatModelConfig;
 import io.quarkiverse.langchain4j.anthropic.runtime.config.LangChain4jAnthropicConfig;
 import io.quarkiverse.langchain4j.anthropic.runtime.config.ToolSearchType;
+import io.quarkiverse.langchain4j.auth.ModelAuthProvider;
 import io.quarkiverse.langchain4j.runtime.NamedConfigUtil;
 import io.quarkus.arc.SyntheticCreationalContext;
 import io.quarkus.runtime.RuntimeValue;
@@ -46,6 +47,8 @@ public class AnthropicRecorder {
     private static final TypeLiteral<Instance<ModelBuilderCustomizer<AnthropicChatModel.AnthropicChatModelBuilder>>> CHAT_MODEL_CUSTOMIZER_TYPE_LITERAL = new TypeLiteral<>() {
     };
     private static final TypeLiteral<Instance<ModelBuilderCustomizer<AnthropicStreamingChatModel.AnthropicStreamingChatModelBuilder>>> STREAMING_CHAT_MODEL_CUSTOMIZER_TYPE_LITERAL = new TypeLiteral<>() {
+    };
+    private static final TypeLiteral<Instance<ModelAuthProvider>> MODEL_AUTH_PROVIDER_TYPE_LITERAL = new TypeLiteral<>() {
     };
 
     private static final String DUMMY_KEY = "dummy";
@@ -64,10 +67,6 @@ public class AnthropicRecorder {
         if (anthropicConfig.enableIntegration()) {
             var chatModelConfig = anthropicConfig.chatModel();
             var apiKey = anthropicConfig.apiKey();
-
-            if (DUMMY_KEY.equals(apiKey)) {
-                throw new ConfigValidationException(createApiKeyConfigProblem(configName));
-            }
 
             if (chatModelConfig.maxRetries() < 1) {
                 throw new ConfigValidationException(createMaxRetriesConfigProblem(configName));
@@ -192,10 +191,12 @@ public class AnthropicRecorder {
             return new Function<>() {
                 @Override
                 public ChatModel apply(SyntheticCreationalContext<ChatModel> context) {
+                    throwIfApiKeyNotConfigured(apiKey, context, configName);
                     builder.listeners(context.getInjectedReference(CHAT_MODEL_LISTENER_TYPE_LITERAL).stream()
                             .collect(Collectors.toList()));
                     QuarkusAnthropicClient.setLogCurlHint(logCurl);
                     QuarkusAnthropicClient.setDisableBetaHint(disableBeta);
+                    QuarkusAnthropicClient.setConfigNameHint(configName);
                     ModelBuilderCustomizer.applyCustomizers(
                             context.getInjectedReference(CHAT_MODEL_CUSTOMIZER_TYPE_LITERAL, Any.Literal.INSTANCE),
                             builder, configName);
@@ -219,10 +220,6 @@ public class AnthropicRecorder {
         if (anthropicConfig.enableIntegration()) {
             var chatModelConfig = anthropicConfig.chatModel();
             var apiKey = anthropicConfig.apiKey();
-
-            if (DUMMY_KEY.equals(apiKey)) {
-                throw new ConfigValidationException(createApiKeyConfigProblem(configName));
-            }
 
             var builder = AnthropicStreamingChatModel.builder()
                     .baseUrl(anthropicConfig.baseUrl())
@@ -341,10 +338,12 @@ public class AnthropicRecorder {
             return new Function<>() {
                 @Override
                 public StreamingChatModel apply(SyntheticCreationalContext<StreamingChatModel> context) {
+                    throwIfApiKeyNotConfigured(apiKey, context, configName);
                     builder.listeners(context.getInjectedReference(CHAT_MODEL_LISTENER_TYPE_LITERAL).stream()
                             .collect(Collectors.toList()));
                     QuarkusAnthropicClient.setLogCurlHint(logCurl);
                     QuarkusAnthropicClient.setDisableBetaHint(disableBeta);
+                    QuarkusAnthropicClient.setConfigNameHint(configName);
                     ModelBuilderCustomizer.applyCustomizers(
                             context.getInjectedReference(STREAMING_CHAT_MODEL_CUSTOMIZER_TYPE_LITERAL,
                                     Any.Literal.INSTANCE),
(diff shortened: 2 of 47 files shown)


model-auth-providers/workload-model-auth-provider/deployment/src/main/java/io/quarkiverse/langchain4j/workload/deployment/WorkloadModelAuthProviderBuildConfig.java
```diff
@@ -0,0 +1,25 @@
...
+     * Whether the Workload Identity Federation ModelAuthProvider should be enabled
+     */
+    @ConfigDocDefault("true")
+    Optional<Boolean> enabled();
+
+    /**
+     * The name of the OIDC client to use for the token exchange.
+     * If not set, the default OIDC client is used.
+     */
+    Optional<String> oidcClientName();
```
sberyozkin, line 24: Note for myself - we don't need this property anymore.
Update - let's keep it for now, as a build time property

sberyozkin:
@jmartisk Hi Jan, I've updated the security docs. I had to restore some content related to API keys and `ModelAuthProvider` that was located earlier in the old `llms.adoc` and was lost with #1549.
Now, the security doc lists 3 options, with the best practices there being primarily about the secure API keys management becoming a subsection of the API keys section.
The old link to the LLM section from Best practices was not working anymore, and it is no longer needed as 2 other sections cover `ModelAuthProvider` and Workload Identity Federation.

Let me also CC @myfear as Markus prepared Best Practices, they remain in this PR only slightly modified to align them with the new proposed structure

geoand:
The failures are odd. We definitely don't have those in other PRs

sberyozkin:
Hi @geoand 

>The failures are odd. We definitely don't have those in other PRs

Confirmed, will have a look

sberyozkin:
Hi @geoand Failing on `main` too, I've been rebasing all the time so maybe this PR collected something, can you please try  `mvn clean install -Dtest=MissingChatModelWhenRequiredShouldFailTest` on main

geoand:
Oh, that's interesting. I'll check

geoand:
@sberyozkin it passed for me. Are you sure you're using the latest `main`?

sberyozkin:
I think so, let me retry

sberyozkin:
Oops, was still on the PR branch :-), looking further into it

model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
```diff
@@ -403,6 +402,17 @@ public StreamingChatModel apply(SyntheticCreationalContext<StreamingChatModel> c
...
 
+    private static <T> void throwIfApiKeyNotConfigured(String apiKey,
+            SyntheticCreationalContext<T> context, String configName) {
+        if (DUMMY_KEY.equals(apiKey) && !isAuthProviderAvailable(context, configName)) {
+            throw new ConfigValidationException(createApiKeyConfigProblem(configName));
+        }
+    }
+
+    private static <T> boolean isAuthProviderAvailable(SyntheticCreationalContext<T> context, String configName) {
+        return context.getInjectedReference(MODEL_AUTH_PROVIDER_TYPE_LITERAL).isResolvable();
```
jmartisk, line 430: This ignores the `configName` parameter and only tries to inject an unqualified provider

model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/OpenAiRecorder.java
```diff
@@ -808,26 +799,37 @@ private LangChain4jOpenAiConfig.OpenAiConfig correspondingOpenAiConfig(LangChain
...
 
+    private static <T> void throwIfApiKeyNotConfigured(String apiKey, String baseUrl,
+            SyntheticCreationalContext<T> context, String configName) {
+        if (DUMMY_KEY.equals(apiKey) && OPENAI_BASE_URL.equals(baseUrl) && !isAuthProviderAvailable(context, configName)) {
+            throw new ConfigValidationException(createApiKeyConfigProblems(configName));
+        }
+    }
+
+    private static <T> boolean isAuthProviderAvailable(SyntheticCreationalContext<T> context, String configName) {
+        return context.getInjectedReference(MODEL_AUTH_PROVIDER_TYPE_LITERAL).isResolvable();
```
jmartisk, line 853: This ignores the `configName` parameter and only tries to inject an unqualified provider

model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
```diff
@@ -65,10 +68,6 @@ public Function<SyntheticCreationalContext<ChatModel>, ChatModel> chatModel(Stri
             var chatModelConfig = anthropicConfig.chatModel();
             var apiKey = anthropicConfig.apiKey();
 
-            if (DUMMY_KEY.equals(apiKey)) {
```
jmartisk, line 68: `responsesChatModel` and `responsesStreamingChatModel` methods still use this old hardcoded DUMMY_KEY build-time check and need to be switched to `throwIfApiKeyNotConfigured` too (lines 260 and 375)

jmartisk (changes requested):

model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
jmartisk: Ah sorry, this applies to `OpenAiRecorder`, not `AnthropicRecorder`. But the rest of the comment stays the same :)

sberyozkin:
Thanks @jmartisk I addressed your comments.

Still dealing with the failing test as discussed with @geoand.
The problem is, this test expects an immediate failure if no api-key is configured, while this PR also checks `ModelAuthProvider` that can be only be checked with the CDI container is available. So this test can't fail at the startup but only when the lazy init completes.

This is the same code we did for Azure OpenAI, (Vertex) Gemini, we discussed it with Georgios back then, checking `Arc` was not 100% guaranteed to produce 100% results.

What I'd like to do is to do 1) early best effort try that should work in most cases where we check Arc if the container is not null, and finally, we also check it at the injection point

sberyozkin:
@geoand @jmartisk 

Agentic test will be fine now. So both OpenAi and Anthropic recorders will now fail early if no api key is available and it can be reliably checked that no ModelAuthProvider is available, when the Arc container is not null and running. 
I guess that it will be fine in most cases, so the 2nd check at the injection point is probably going to be redundant, but it is still rather cheap and with both checks it is guaranteed the failure before the request to LLM is made when neither api key not ModelAuthProvider is available.

If you are fine with this approach, I'll follow up in another PR with similar updates for Azure OpenAI and (Vertx) Gemini

model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
```diff
@@ -403,6 +407,37 @@ public StreamingChatModel apply(SyntheticCreationalContext<StreamingChatModel> c
         }
     }
 
+    private static void throwIfApiKeyNotConfigured(String apiKey, String configName) {
+        ArcContainer container = Arc.container();
```
geoand, line 411: Instead of doing this, we should check for the existence of the `ModelAuthProvider` instance via `SyntheticCreationalContext`. The same thing should be done in the other providers.

geoand (commented):
Thanks @sberyozkin!

I added a comment on the Anthropic recorder that should also be applied to the other providers

model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java
sberyozkin: Hi @geoand 

We already do it from the start of the PR, see [here](https://github.com/quarkiverse/quarkus-langchain4j/pull/2639/changes#diff-db530f5631b961d0c53e1502562b07d8e23c020fd369010df98c835527e60548R418-R431),  I only brought this one back to get that failing Agentic `MissingChatModelWhenRequiredShouldFailTest` test passing, see my comment above.

The test expects an immediate failure at the start up when no api key is configured, when `SyntheticCreationalContext` is not available yet for lazily initialized Agentic beans.

So this code that checks `Arc` is the best effort only to try to fail immediately, before the requests start arriving, if neither api-key nor `ModelAuthProvider` is available. 

If, due to the Arc unavailability, this check gives false, the later `SyntheticCreationalContext` check will definitely fire if neither api-key nor `ModelAuthProvider` is available. 

If we don't do it and retain the `SyntheticCreationalContext` check only then we'd need to change test expectations to fail at the bean creation time instead, I'm fine with it for sure, let me know what you prefer

Thanks

geoand: 👋🏽 

The way I envision it, the behavior should be that a Quarkus application without an API key and without a `ModelAuthProvider` bean configured should fail at startup. Isn't that what we do? If so, there is no need to fall back to checking `Arc.containerI()`...

sberyozkin: @geoand 

Sorry, got distracted a bit, 

>The way I envision it, the behavior should be that a Quarkus application without an API key and without a ModelAuthProvider bean configured should fail at startup.

Right, I thought so too, but the Agentic `MissingChatModelWhenRequiredShouldFailTest` test fails because, from what I understand, that test setup failure check comes too early before the synthetic creational context is checked.

I'll have a look at it again to remind myself why I decided to introduce an Arc fallback, and update you and Jan a bit later

Thanks, enjoy the weekend

geoand: 👍🏽 

You too!

sberyozkin: Hi @geoand @jmartisk 

So, to make the test passing, with OpenAiRecorder relying on `SyntheticCreationalContext` only to fail when neither api key nor `ModelAuthProvider` is available, the test must be updated like this:
```
--- a/agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/validation/MissingChatM
  odelWhenRequiredShouldFailTest.java
  +++ b/agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/validation/MissingChatM
  odelWhenRequiredShouldFailTest.java
  @@
  -import jakarta.inject.Inject;
  -import jakarta.inject.Singleton;
  +import static org.assertj.core.api.Assertions.assertThat;
  +import static org.assertj.core.api.Assertions.catchThrowable;

  -import org.assertj.core.api.Assertions;
  +import jakarta.inject.Inject;
  +
   import org.jboss.shrinkwrap.api.ShrinkWrap;
  @@
       @RegisterExtension
       static final QuarkusUnitTest unitTest = new QuarkusUnitTest()
               .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
  -                    .addClasses(DummyChatModel.class, StoryCreator.class, CreativeWriter.class, 
  AudienceEditor.class))
  -            .assertException(throwable -> Assertions.assertThat(throwable)
  -                    .hasMessageContaining("quarkus.langchain4j.openai.api-key"));
  +                    .addClasses(DummyChatModel.class, StoryCreator.class, CreativeWriter.class, 
  AudienceEditor.class));
  +
  +    @Inject
  +    StoryCreator storyCreator;

       @Test
       void test() {
  +        assertThat(catchThrowable(() -> storyCreator.write("dragons and wizards", "kids")))
  +                .hasStackTraceContaining("quarkus.langchain4j.openai.api-key");
       }

  -    @Singleton
  -    public static class StoryCreatorConsumer {
  -
  -        @Inject
  -        StoryCreator storyCreator;
  -    }
  -

```

otherwise, 

```
 .assertException(throwable -> Assertions.assertThat(throwable)
  -                    .hasMessageContaining("quarkus.langchain4j.openai.api-key"));
```

does not fire at all, since the synthetic creational context for the OpenAI chat model is only becoming available when `StoryCreator` is accessed.

So, if we keep `SyntheticCreationalContext` check only then, at least with this Agentic original test setup, we need to fix it as shown in the diff above.

This is why the PR proposes 1) eager early check with Arc that should catch immediately both the api key and `ModelAuthProvider` absence, and 2) final check with the synthetic creational context if Arc was not yet ready at the time of the 1st check.

IMHO it is a reasonable compromise, let me know please what you think

jmartisk: Yeah, this double-check proposal makes sense to me

geoand: I still don't think this is the right thing to do, but it's minor, so let's move ahead like this for now

sberyozkin: Thanks @geoand @jmartisk, yeah, I'm not 100% happy with having to do both checks either but at the moment at appears to be right combination if we want to fail immediately at the startup. 
I'll be happy to follow up with implementing a more optimal strategy to fail early across all providers that support `ModelAuthProvider` once we work it out. One idea I had was to consider making key beans that require a model eagerly, no lazily, initialized - but it will probably  impact some performance stats... I guess it would all deserve a dedicated issue/discussion/PR.

jmartisk (approved):

Issue #1952, opened 2025-11-21
[Enhancement] Add support for OpenAi Responses API


ramesh-spendflo:
It appears that the OpenAI `v1/responses` API is not currently supported: (https://platform.openai.com/docs/api-reference/responses/create)
Can we please support this API?

geoand:
When https://github.com/langchain4j/langchain4j/issues/3252 lands, we can support it too

hariprasadkc:
@geoand langchain4j support for v1/responses has landed. Can we pick this up ?

nabeel001:
Hey @geoand 
All checks have passed and the build is successful. Kindly review and merge the PR.
https://github.com/quarkiverse/quarkus-langchain4j/pull/2649#issuecomment-4864491578

geoand:
I'm traveling,  so that will have to wait

hariprasadkc:
Thanks for the contribution @nabeel001 and @geoand for the quick turnaround. Do you have a rough idea on when it might make into a release ?

geoand:
Hopefully I'll do a new `1.12` release this week

PR #2649, opened 2026-07-02
feat(openai): Add support for the OpenAI Responses API

nabeel001:
## 🚀 Summary

Adds opt-in support for OpenAI's **Responses API** (`POST /v1/responses`) to the `quarkus-langchain4j-openai` extension, alongside the existing Chat Completions support. This is the first step toward resolving #1952.

Why it matters: some models (like `gpt-5-pro` and `o1-pro`) are only available through the Responses API, and it unlocks Responses-specific capabilities. By default nothing changes - you opt in with a single config property.

## 🧩 How it works

Upstream langchain4j exposes the Responses API as separate `OpenAiResponsesChatModel` / `OpenAiResponsesStreamingChatModel` classes. Unlike the classic `OpenAiChatModel`, these have no builder-factory SPI and use their own internal client, so the existing `QuarkusOpenAiClient` path can't be reused. Instead we feed the upstream models a Quarkus-native `HttpClientBuilder` (`JaxRsHttpClientBuilder` — the same transport Ollama and Gemini use). This keeps traffic on the Quarkus REST Client stack, so native-image, TLS registry, and HTTP-level observability all keep working.

## ✨ What's included

- **Opt-in switch:** `quarkus.langchain4j.openai.chat-model.mode` = `chat-completion` (default) or `responses`. Fully non-breaking.
- **Responses settings** under `quarkus.langchain4j.openai.chat-model.responses.*` — `store`, `previous-response-id`, `reasoning-summary`, `max-output-tokens`, `max-tool-calls`, `parallel-tool-calls`, `include`, `truncation`, `text-verbosity`, `prompt-cache-key`, `prompt-cache-retention`, `safety-identifier`, `top-logprobs`, `strict-tools`, `stream-include-obfuscation`. Shared options (model name, temperature, top-p, response format, service tier, reasoning effort, logging) are reused.
- **Same beans you already use** — `ChatModel` and `StreamingChatModel` inject exactly as before; no new API to learn.
- Blocking + streaming, tool/function calling, and structured output all supported, plus TLS registry, listeners, `ModelBuilderCustomizer`, and native reflection registration.
- Docs added to `openai-chat-model.adoc`.

## 📌 Scope & notes

- Targets `openai-vanilla` only. Azure OpenAI Responses support is a possible follow-up (Azure uses custom model wrappers).
- In `responses` mode, `log-requests-curl` and proxy configuration aren't honored (documented). Standard `log-requests` / `log-responses` still work.

## ✅ Test plan

- [x] Blocking chat hits `/v1/responses` with the correct payload and auth header
- [x] SSE streaming (partial tokens + final response)
- [x] Tool/function call output parsed into a tool execution request
- [x] Existing chat-completion path still passes (no regression)

### DIFF
--- a/docs/modules/ROOT/pages/openai-chat-model.adoc
+++ b/docs/modules/ROOT/pages/openai-chat-model.adoc
@@ -111,6 +111,42 @@ import jakarta.inject.Inject;
 
 If no model name is specified, the default configuration is used.
 
+== Responses API
+
+By default, the chat model targets OpenAI's `/v1/chat/completions` endpoint. OpenAI also offers the newer
+https://platform.openai.com/docs/api-reference/responses[Responses API] (`/v1/responses`), which is required for models
+that are only exposed through it (for example `gpt-5-pro`, `o1-pro`) and which unlocks Responses-specific settings.
+
+To switch the chat model to the Responses API, set the build-time `mode` property:
+
+[source,properties,subs=attributes+]
+----
+quarkus.langchain4j.openai.chat-model.mode=responses
+quarkus.langchain4j.openai.api-key=sk-...
+quarkus.langchain4j.openai.chat-model.model-name=gpt-5-pro
+----
+
+Both the blocking `ChatModel` and the `StreamingChatModel` are supported in this mode, including tool/function calling and
+structured outputs. The injection points and AI service usage are unchanged — only the underlying API and endpoint differ.
+
+Responses-specific options are configured under `quarkus.langchain4j.openai.chat-model.responses.*`, for example:
+
+[source,properties,subs=attributes+]
+----
+quarkus.langchain4j.openai.chat-model.responses.store=false
+quarkus.langchain4j.openai.chat-model.responses.reasoning-summary=auto
+quarkus.langchain4j.openai.chat-model.responses.max-output-tokens=1024
+----
+
+[NOTE]
+====
+In `responses` mode the model communicates over a Quarkus REST Client-based HTTP client (the same engine used elsewhere,
+so native image and TLS registry configuration continue to work). A few features of the chat-completions path do not apply:
+`quarkus.langchain4j.openai.log-requests-curl` and proxy configuration (`quarkus.langchain4j.openai.proxy-configuration-name`)
+are not honored for Responses traffic. Request/response logging still works via the standard
+`log-requests`/`log-responses` settings.
+====
+
 == Advanced usage
 
 The OpenAI extension internally uses a Quarkus REST Client to make API calls. This client can also be used directly:
--- a/model-providers/openai/openai-vanilla/deployment/src/main/java/io/quarkiverse/langchain4j/openai/deployment/OpenAiProcessor.java
+++ b/model-providers/openai/openai-vanilla/deployment/src/main/java/io/quarkiverse/langchain4j/openai/deployment/OpenAiProcessor.java
@@ -24,6 +24,8 @@ import dev.langchain4j.model.openai.OpenAiAudioTranscriptionModel;
 import dev.langchain4j.model.openai.OpenAiChatModel;
 import dev.langchain4j.model.openai.OpenAiEmbeddingModel;
 import dev.langchain4j.model.openai.OpenAiModerationModel;
+import dev.langchain4j.model.openai.OpenAiResponsesChatModel;
+import dev.langchain4j.model.openai.OpenAiResponsesStreamingChatModel;
 import dev.langchain4j.model.openai.OpenAiStreamingChatModel;
 import dev.langchain4j.model.openai.spi.OpenAiAudioTranscriptionModelBuilderFactory;
 import dev.langchain4j.model.openai.spi.OpenAiChatModelBuilderFactory;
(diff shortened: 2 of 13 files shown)


model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/OpenAiRecorder.java
geoand, line 1: Please remove all lambdas and method references that were added here. In Quarkus we generally try to not use them because they affect (even minimally) the startup time

geoand (commented):
Very cool! I added a small comment

model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/OpenAiRecorder.java
nabeel001: Removed all lambdas and method references

nabeel001:
Hey @geoand 
I have addressed your comment. Kindly re-review it.

docs/modules/ROOT/pages/openai-chat-model.adoc
```diff
@@ -111,6 +111,42 @@ import jakarta.inject.Inject;
 
 If no model name is specified, the default configuration is used.
 
+== Responses API
+
+By default, the chat model targets OpenAI's `/v1/chat/completions` endpoint. OpenAI also offers the newer
+https://platform.openai.com/docs/api-reference/responses[Responses API] (`/v1/responses`), which is required for models
+that are only exposed through it (for example `gpt-5-pro`, `o1-pro`) and which unlocks Responses-specific settings.
+
+To switch the chat model to the Responses API, set the build-time `mode` property:
```
geoand, line 120: In the futrure, I wonder if this switch should be done automatically... Something to think about for a follow up

geoand (commented):
Please squash the commits and this should be good to go

nabeel001:
Hey @geoand 
I have squashed my commits to a single one. Kindly verify the same and trigger the git workflows.

model-providers/openai/openai-vanilla/deployment/src/main/java/io/quarkiverse/langchain4j/openai/deployment/OpenAiProcessor.java
```diff
@@ -121,42 +127,79 @@ void generateBeans(OpenAiRecorder recorder,
...
-                        .addInjectionPoint(ParameterizedType.create(DotNames.CDI_INSTANCE,
-                                new Type[] { ClassType.create(DotNames.CHAT_MODEL_LISTENER) }, null))
-                        .addInjectionPoint(ParameterizedType.create(DotNames.CDI_INSTANCE,
-                                new Type[] { ParameterizedType.create(DotNames.MODEL_BUILDER_CUSTOMIZER,
-                                        new Type[] { ClassType.create(OPENAI_STREAMING_CHAT_MODEL_BUILDER) }, null) },
-                                null), ANY)
-                        .createWith(recorder.streamingChatModel(configName));
-                addQualifierIfNecessary(streamingBuilder, configName);
-                beanProducer.produce(streamingBuilder.done());
+                if (responsesMode) {
```
geoand: Minor nitpick: please rename to `useResponseMode`

nabeel001: I have addressed the comment

model-providers/openai/openai-vanilla/deployment/src/main/java/io/quarkiverse/langchain4j/openai/deployment/OpenAiProcessor.java
```diff
@@ -121,42 +127,79 @@ void generateBeans(OpenAiRecorder recorder,
...
+                            .addInjectionPoint(Type.create(ProxyConfigurationRegistry.class))
+                            .addInjectionPoint(ParameterizedType.create(DotNames.CDI_INSTANCE,
+                                    new Type[] { ClassType.create(DotNames.CHAT_MODEL_LISTENER) }, null))
+                            .addInjectionPoint(ParameterizedType.create(DotNames.CDI_INSTANCE,
+                                    new Type[] { ParameterizedType.create(DotNames.MODEL_BUILDER_CUSTOMIZER,
+                                            new Type[] { ClassType.create(OPENAI_STREAMING_CHAT_MODEL_BUILDER) }, null) },
+                                    null), ANY)
+                            .createWith(recorder.streamingChatModel(configName));
+                    addQualifierIfNecessary(streamingBuilder, configName);
+                    beanProducer.produce(streamingBuilder.done());
```
geoand: Now that I see this again, do we really need to duplicate everything? Can't we extract the common stuff used to configure `SyntheticBeanBuildItem` and then only change  `createWith` according to whether `useResponseMode` is set?

nabeel001: Addressed the comment

geoand (approved):
Thanks!

PR #2251, opened 2026-03-19
Introduce Skills

jmartisk:
Quarkus-side integration for skills. See https://github.com/langchain4j/langchain4j/pull/4646 for the upstream implementation. 
For now, only tool mode is supported. We can add shell mode later if there is demand.

PR #2654, opened 2026-07-03
Introduce @Skills annotation

### LINKED ISSUE #2653: Implement new @Skills annotation
Add a new `@Skills` annotation to be used with agents. When present it enables the agent to use all the available skills. Optionally it can be passed with an array of `Strings` in order to select and put at disposition of the agent only a subset of all the possible skills, like in:

`@Skills("adventure-trip", "business-trip", "family-trip")`

It should fail fast if a required skill doesn't exist.

mariofusco:
- Closes: https://github.com/quarkiverse/quarkus-langchain4j/issues/2653

/cc @kdubois

### DIFF
--- a/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticLangChain4jDotNames.java
+++ b/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticLangChain4jDotNames.java
@@ -87,6 +87,8 @@ public final class AgenticLangChain4jDotNames {
     public static final DotName TOOL_SUPPLIER = DotName
             .createSimple(ToolsSupplier.class.getName());
 
+    public static final DotName SKILLS = DotName.createSimple("io.quarkiverse.langchain4j.skills.Skills");
+
     public static final DotName CDI_BEAN = DotName.createSimple("io.quarkiverse.langchain4j.agentic.runtime.CdiBean");
 
     public static final DotName MEMORY_ID = DotName.createSimple(MemoryId.class.getName());
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -32,6 +32,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -133,11 +134,13 @@ import io.quarkus.arc.DefaultBean;
 import io.quarkus.arc.InstanceHandle;
 import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
 import io.quarkus.arc.deployment.AnnotationsTransformerBuildItem;
+import io.quarkus.arc.deployment.BeanContainerBuildItem;
 import io.quarkus.arc.deployment.CustomScopeAnnotationsBuildItem;
 import io.quarkus.arc.deployment.GeneratedBeanBuildItem;
 import io.quarkus.arc.deployment.GeneratedBeanGizmoAdaptor;
 import io.quarkus.arc.deployment.SynthesisFinishedBuildItem;
 import io.quarkus.arc.deployment.SyntheticBeanBuildItem;
+import io.quarkus.arc.deployment.SyntheticBeansRuntimeInitBuildItem;
 import io.quarkus.arc.deployment.UnremovableBeanBuildItem;
 import io.quarkus.arc.deployment.ValidationPhaseBuildItem;
 import io.quarkus.arc.processor.BuiltinScope;
@@ -146,6 +149,7 @@ import io.quarkus.deployment.Capability;
 import io.quarkus.deployment.GeneratedClassGizmoAdaptor;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.annotations.Consume;
 import io.quarkus.deployment.annotations.ExecutionTime;
 import io.quarkus.deployment.annotations.Record;
 import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
@@ -598,39 +602,52 @@ public class AiServicesProcessor {
                 }
             }
 
-            declarativeAiServiceProducer.produce(
-                    new DeclarativeAiServiceBuildItem(
-                            declarativeAiServiceClassInfo,
-                            chatLanguageModelSupplierClassDotName,
-                            streamingChatLanguageModelSupplierClassDotName,
-                            tools,
-                            chatMemoryProviderSupplierClassDotName,
-                            retrievalAugmentorSupplierClassName,
-                            customRetrievalAugmentorSupplierClassIsABean,
-                            moderationModelSupplierClassName,
-                            imageModelSupplierClassName(reflectiveClassProducer, unremovableBeanProducer, instance, index),
-                            determineChatMemorySeeder(declarativeAiServiceClassInfo, generatedClassOutput),
-                            determineThinkingHandler(declarativeAiServiceClassInfo, generatedClassOutput),
-                            systemMessageProviderClassDotName,
-                            cdiScope(customScopes, declarativeAiServiceClassInfo),
-                            chatModelName,
-                            moderationModelName,
-                            imageModelName,
-                            toolProviderClassName,
-                            toolSearchStrategyClassName,
-                            beanName(declarativeAiServiceClassInfo),
-                            toolHallucinationStrategy(instance),
-                            classInputGuardrails(declarativeAiServiceClassInfo, index),
-                            classOutputGuardrails(declarativeAiServiceClassInfo, index),
-                            toolArgumentsErrorHandlerDotName(declarativeAiServiceClassInfo, generatedBeanProducer),
-                            toolExecutionErrorHandlerDotName(declarativeAiServiceClassInfo, generatedBeanProducer),
-                            maxToolCallingRoundTrips,
-                            maxToolCallsPerResponse,
-                            allowContinuousForcedToolCalling,
-                            // we need to make these @DefaultBean because there could be other CDI beans of the same type that need to take precedence
-                            impliedRegisterAiServiceTarget.contains(declarativeAiServiceClassInfo.name()),
-                            shouldThrowExceptionOnEventError,
-                            chatMemoryFlushStrategySupplierClassDotName));
+            // @Skills annotation support
+            List<String> skillNames = null;
+            AnnotationInstance skillsAnnotation = declarativeAiServiceClassInfo
+                    .declaredAnnotation(LangChain4jDotNames.SKILLS);
+            if (skillsAnnotation != null) {
+                AnnotationValue skillsValue = skillsAnnotation.value();
+                if (skillsValue != null && skillsValue.asStringArray().length > 0) {
+                    skillNames = List.of(skillsValue.asStringArray());
+                } else {
+                    skillNames = List.of();
+                }
+            }
+
+            declarativeAiServiceProducer.produce(new DeclarativeAiServiceBuildItem(
+                    declarativeAiServiceClassInfo,
+                    chatLanguageModelSupplierClassDotName,
+                    streamingChatLanguageModelSupplierClassDotName,
+                    tools,
+                    chatMemoryProviderSupplierClassDotName,
+                    retrievalAugmentorSupplierClassName,
+                    customRetrievalAugmentorSupplierClassIsABean,
+                    moderationModelSupplierClassName,
+                    imageModelSupplierClassName(reflectiveClassProducer, unremovableBeanProducer, instance, index),
+                    determineChatMemorySeeder(declarativeAiServiceClassInfo, generatedClassOutput),
+                    determineThinkingHandler(declarativeAiServiceClassInfo, generatedClassOutput),
+                    systemMessageProviderClassDotName,
+                    cdiScope(customScopes, declarativeAiServiceClassInfo),
+                    chatModelName,
+                    moderationModelName,
+                    imageModelName,
+                    toolProviderClassName,
+                    toolSearchStrategyClassName,
+                    beanName(declarativeAiServiceClassInfo),
+                    toolHallucinationStrategy(instance),
+                    classInputGuardrails(declarativeAiServiceClassInfo, index),
+                    classOutputGuardrails(declarativeAiServiceClassInfo, index),
+                    toolArgumentsErrorHandlerDotName(declarativeAiServiceClassInfo, generatedBeanProducer),
+                    toolExecutionErrorHandlerDotName(declarativeAiServiceClassInfo, generatedBeanProducer),
+                    maxToolCallingRoundTrips,
+                    maxToolCallsPerResponse,
+                    allowContinuousForcedToolCalling,
+                    // we need to make these @DefaultBean because there could be other CDI beans of the same type that need to take precedence
+                    impliedRegisterAiServiceTarget.contains(declarativeAiServiceClassInfo.name()),
+                    shouldThrowExceptionOnEventError,
+                    chatMemoryFlushStrategySupplierClassDotName,
+                    skillNames));
 
         }
         toolProviderProducer.produce(new ToolProviderMetaBuildItem(toolProviderInfos));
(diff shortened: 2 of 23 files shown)


kdubois:
You'll want to update the docs as well right?

mariofusco:
> You'll want to update the docs as well right?

Yes, I will.

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticLangChain4jDotNames.java
```diff
@@ -87,6 +87,8 @@ public final class AgenticLangChain4jDotNames {
     public static final DotName TOOL_SUPPLIER = DotName
             .createSimple(ToolsSupplier.class.getName());
 
+    public static final DotName SKILLS = DotName.createSimple("io.quarkiverse.langchain4j.skills.Skills");
```
geoand, line 90: This could be a reference to the actual class, right?

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticLangChain4jDotNames.java
```diff
@@ -87,6 +87,8 @@ public final class AgenticLangChain4jDotNames {
...
+
     public static final DotName CDI_BEAN = DotName.createSimple("io.quarkiverse.langchain4j.agentic.runtime.CdiBean");
```
geoand, line 92: Not your fault obviously (as I made the mistake), but this is in wrong package 😢

skills/runtime/src/main/java/io/quarkiverse/langchain4j/skills/Skills.java
geoand, line 1: I don't understand why we have a new `skills` module. Care to elaborate?

mariofusco:
> You'll want to update the docs as well right?

I updated documentation in `skills.adoc`, please check it. As a side note, I have the impression that now registering skills through the `systemMessageProviderSupplier` field of the `@RegisterAiService` annotation seems to be a less clear and configurable alternative. Should we remove it?

geoand:
>  Should we remove it?

Remove what exactly?

skills/runtime/src/main/java/io/quarkiverse/langchain4j/skills/Skills.java
mariofusco: To be honest this is not new, it has been introduced with this pull request https://github.com/quarkiverse/quarkus-langchain4j/pull/2251 I only added my stuff to it.

geoand: You're right; somehow I missed that...

mariofusco:
> Remove what exactly?

Remove the `SkillsSystemMessageProvider` as used with:

`@RegisterAiService(systemMessageProviderSupplier = SkillsSystemMessageProvider.class)`

Having the new `@Skills` annotation, it seems an useless alternative and could confuse users?

geoand:
We should probably deprecate it. WDYT @jmartisk ?

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticLangChain4jDotNames.java
mariofusco: I don't think so, because it is in the skills module.

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticLangChain4jDotNames.java
mariofusco: Sorry, don't understanding this. The package looks right?

geoand: It's wrong because of `runtime`

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticLangChain4jDotNames.java
geoand: Right, I asked before realizing we had a separate `skills` module

agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticLangChain4jDotNames.java
mariofusco: Ah you mean, it should be refactored and moved to a different package? I guess this change doesn't belong here and could be addressed with a different pull request?

geoand: Yeah, I'll do it separately

mariofusco:
I reviewed and improved what I had at the light of this comment https://github.com/quarkusio/quarkus-workshop-langchain4j/pull/364#discussion_r3528522920 also adding a few more tests to double check that the generated system message is correct.

Please let me know if you have any other remarks or this is ready to be merged.

geoand:
I'll try and take a look tomorrow

geoand (approved):
LGTM, but I want @jmartisk to take a look as well

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -598,39 +601,52 @@ public void findDeclarativeServices(CombinedIndexBuildItem indexBuildItem,
...
+            AnnotationInstance skillsAnnotation = declarativeAiServiceClassInfo
+                    .declaredAnnotation(LangChain4jDotNames.SKILLS);
+            if (skillsAnnotation != null) {
+                AnnotationValue skillsValue = skillsAnnotation.value();
+                if (skillsValue != null && skillsValue.asStringArray().length > 0) {
+                    buildItem.setSkillNames(List.of(skillsValue.asStringArray()));
+                } else {
+                    buildItem.setSkillNames(List.of());
+                }
+            }
```
geoand: Actually, I don't like using a setter just for this. The same pattern should be used throughout. Ideally (likely in a next PR), we move off the constructor and into a builder

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
```diff
@@ -1332,6 +1349,22 @@ public void handleDeclarativeServices(AiServicesRecorder recorder,
...
+    void validateSkillNames(List<DeclarativeAiServiceBuildItem> items, AiServicesRecorder recorder) {
+        Set<String> allSkillNames = new LinkedHashSet<>();
+        for (DeclarativeAiServiceBuildItem bi : items) {
+            List<String> names = bi.getSkillNames();
+            if (names != null && !names.isEmpty()) {
+                allSkillNames.addAll(names);
+            }
+        }
+        if (!allSkillNames.isEmpty()) {
+            recorder.validateSkillNames(List.copyOf(allSkillNames));
```
geoand: Why are we validating at runtime? Can't it be done at build time?

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
mariofusco: Fixed.

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
mariofusco: Unfortunately this is not possible because `quarkus.langchain4j.skills.directories` is a runtime config property and skills are loaded from the filesystem at startup, so the list of available skill names doesn't exist at build time. The validation runs as early as possible, at `RUNTIME_INIT` right after the skills beans are created, so it still fails fast at startup rather than lazily on first use.

geoand: Got it, thanks

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/AiServicesRecorder.java
```diff
@@ -400,6 +405,21 @@ public QuarkusAiServiceContext apply(SyntheticCreationalContext<QuarkusAiService
                     aiServiceContext.eventListenerRegistrar
                             .shouldThrowExceptionOnEventError(info.shouldThrowExceptionOnEventError());
 
+                    // @Skills annotation support
+                    if (info.skillNames() != null) {
+                        Instance<SkillsConfigurator> skillsInstance = Arc.container().select(SkillsConfigurator.class);
```
geoand: We shouldn't do this. We should use `creationalContext.getInjectedReference()`

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/AiServicesRecorder.java
```diff
@@ -416,6 +436,15 @@ private static Class<?> loadClass(String info) throws ClassNotFoundException {
         };
     }
 
+    public void validateSkillNames(List<String> skillNames) {
+        Instance<SkillsConfigurator> instance = Arc.container().select(SkillsConfigurator.class);
```
geoand: This should not be done as per the previous comment

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/AiServicesRecorder.java
mariofusco: I reworked this, please give it a second look.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/AiServicesRecorder.java
mariofusco: Same.

geoand (approved):
Thanks!

cescoffier:

jmartisk:
> > Remove what exactly?
> 
> Remove the `SkillsSystemMessageProvider` as used with:
> 
> `@RegisterAiService(systemMessageProviderSupplier = SkillsSystemMessageProvider.class)`
> 
> Having the new `@Skills` annotation, it seems an useless alternative and could confuse users?

Hey, I'm back from a long PTO now, so sorry for the late reply and that I might not have my brain booted up properly yet. I don't understand, how are users supposed to provide the system message provider with this new `@Skills` annotation? That would be providing a custom `SkillsConfigurator`, maybe? But I'm missing in the docs/javadocs how to do override the built-in one.

jmartisk:
Also, how does it work when multiple AI service methods are invoked in a single session and they have different skills declared? You can't retroactively change the system message for each invocation separately when they share the same chat memory, no?

skills/runtime/src/main/java/io/quarkiverse/langchain4j/skills/runtime/SkillsToolProvider.java
```diff
@@ -14,10 +17,12 @@ public class SkillsToolProvider implements ToolProvider {
 
     private final ToolProvider delegate;
     private final Skills skills;
+    private final List<Skill> allSkills;
```
jmartisk, line 20: A comment explaining the difference between `skills` and `allSkills` would've been nice

mariofusco:
> I don't understand, how are users supposed to provide the system message provider with this new `@Skills` annotation? That would be providing a custom `SkillsConfigurator`, maybe? But I'm missing in the docs/javadocs how to do override the built-in one.

I don't think that this is possible at the moment. In theory users could create an alternative implementation of the `SkillsConfigurator` interface, but I'm afraid that if you do this in practice (I didn't try), it will simply go in conflict with the `DefaultSkillsConfigurator` one, provided out-of-the-box, and CDI will complain that it cannot disambiguate between the two.

That said do you think that users could actually need to do something like that? What the current implementation does is simply enriching the system message with the list of available skills and a quick instruction on how to use them.

```
    default String buildSkillsSystemMessage(List<String> skillNames) {
        return "You have access to the following skills:\n"
                + formatAvailableSkills(skillNames)
                + "\nWhen the user's request relates to one of these skills, "
                + "activate it first using the `activate_skill` tool before proceeding.";
    }
```

In which practical scenario you may want to change this behavior and plug your own system message?

mariofusco:
> Also, how does it work when multiple AI service methods are invoked in a single session and they have different skills declared? You can't retroactively change the system message for each invocation separately when they share the same chat memory, no?

Good question. For agents you don't have this problem because the `@Skills` annotation has to be put at the method level and anyway an agent interface can have only and only one agentic method. For what regards AiServices you're supposed to put this annotation at the type level and this implies that all methods in that AiService will be forced to use the same set of skills. Maybe we could allow to place the `@Skills` annotation at method level also for AiService thus allowing each of them do declare its on set of skills?

kdubois:
> Maybe we could allow to place the `@Skills` annotation at method level also for AiService thus allowing each of them do declare its on set of skills?

That would make sense to me ...

jmartisk:
> In which practical scenario you may want to change this behavior and plug your own system message?

I guess that at minimum, it should be possible to have it in a different language than English.

jmartisk:
> For agents you don't have this problem because the @Skills annotation has to be put at the method level and anyway an agent interface can have only and only one agentic method.

But can't a single memory be shared between different agents in one session? 

> Maybe we could allow to place the `@Skills` annotation at method level also for AiService thus allowing each of them do declare its on set of skills?

Well that makes sense, but then we might need to address the original concern about changing the system message depending on which method is being executed :) if the set of skills changes inside a conversation, I would say that it may have to be moved to the user message...

jmartisk:
... or instead of listing the skills in the system/user message, we would expose a tool that lists available skills so that the LLM can retrieve the list only when it needs to

mariofusco:
> > Maybe we could allow to place the `@Skills` annotation at method level also for AiService thus allowing each of them do declare its on set of skills?
> 
> Well that makes sense, but then we might need to address the original concern about changing the system message depending on which method is being executed :) if the set of skills changes inside a conversation, I would say that it may have to be moved to the user message...

Sorry, I don't understand the problem. AFAIK also the system message is set at method level, so each method in the AiService has its own system message, isn't it?

jmartisk:
> Sorry, I don't understand the problem. AFAIK also the system message is set at method level, so each method in the AiService has its own system message, isn't it?

But multiple methods may be called in a single conversation (in a single chat memory), so you would need to change the system message within a conversation, which is illegal (well the LLM will accept it because it's stateless, but it could confuse it a lot)

mariofusco:
> > Sorry, I don't understand the problem. AFAIK also the system message is set at method level, so each method in the AiService has its own system message, isn't it?
> 
> But multiple methods may be called in a single conversation (in a single chat memory), so you would need to change the system message within a conversation, which is illegal (well the LLM will accept it because it's stateless, but it could confuse it a lot)

Are you sure? Can you please point me to a test or any other example doing this?

jmartisk:
Obviously we don't have an example because it's not supported yet :) but you could imagine that this will affect pretty much any AI service that uses a chat memory AND declares multiple methods AND you have the `@Skills` annotations on its methods.

jmartisk:
Actually we can already reproduce this with multiple AI services. Let me experiment with it when I get a bit of time and I'll come back with the findings.

mariofusco:
> Actually we can already reproduce this with multiple AI services. Let me experiment with it when I get a bit of time and I'll come back with the findings.

Ok, but I believe that sharing the same ChatMemory across multiple AiServices, while technically feasible, is considered an anti-pattern (@dliubarskyi please confirm). Moreover if you do so you will have the same problem of swapping the system message when moving from an AiService to the other regardless of the `@Skills` annotation. And actually you would have the same problem also for 2 methods declared in the same AiService. Am I missing something?

dliubarskyi:
> Moreover if you do so you will have the same problem of swapping the system message when moving from an AiService to the other

`ChatMemory` can handle changed `SystemMessage` without problems

mariofusco:
> `ChatMemory` can handle changed `SystemMessage` without problems

Ok, so in this case this will work also for skills.

jmartisk:
> `ChatMemory` can handle changed `SystemMessage` without problems

Of course the memory can, but it's a gigantic anti pattern, no? You basically show the LLM a history that is not "true", you retroactively changed something in it. Imagine asking an LLM how much is 2+2, it answers 4, but then you retroactively add an instruction that 2+2 is 5, but keep its original response where it said 4, and then you continue the discussion from there. Then you basically "prove" that the LLM lied to you.

In the context of skills, when you change the skill list retroactively by removing one skill from it (after it was used), then basically the history will say that the LLM invoked a skill that wasn't meant to be available.
I would say this is very dangerous and can lead to very weird results and maybe security issues.

mariofusco:
See https://github.com/quarkiverse/quarkus-langchain4j/pull/2688

dliubarskyi:
> Of course the memory can, but it's a gigantic anti pattern, no? You basically show the LLM a history that is not "true", you retroactively changed something in it. Imagine asking an LLM how much is 2+2, it answers 4, but then you retroactively add an instruction that 2+2 is 5, but keep its original response where it said 4, and then you continue the discussion from there. Then you basically "prove" that the LLM lied to you.

I was talking only about technical possibility. But I agree it is not ideal. Although, by default, when system message changes, it is put at the end of the chat memory, so it is "applied" to the following messages, not to the existing ones

PR #2655, opened 2026-07-04
Distinguish RAG-augmented context in Dev UI chat

### LINKED ISSUE #2595: Dev UI chat: distinguish RAG-augmented messages from the user message
While trying the [`chatbot-easy-rag` ](https://github.com/quarkiverse/quarkus-langchain4j/tree/main/samples/chatbot-easy-rag)sample, I noticed that in the Dev UI chat the retrieved RAG context is shown merged into the user's own message bubble, with no visual distinction. The System message, for example, already gets its own avatar.

It would be nice to render the augmented message distinctly (its own avatar/style), similar to how System and AI messages already look.

One nuance from a quick look at the code: with RAG enabled the Dev UI uses the non-streaming path (`ChatJsonRPCService.chat`), where the augmented message is stored in `ChatMemory` as a plain `UserMessage` and serialized as `MessageType.USER`, so the frontend currently has no way to tell it apart. The streaming path already emits it as a separate `augmentedMessage` field, but the UI forces non-streaming when RAG is on.

Leaving this open as a question: is this something the project would want to change?

Image

omatheusmesmo:
## Describe your changes

When RAG is enabled in the Dev UI chat, the retrieved context was merged into the user's message with no visual distinction.

The augmented part is now detected by common prefix + suffix against the original message and wrapped in a dedicated `lc4j-rag-context` box, kept in the same bubble but styled apart. If the augmentor rewrote the message instead of only inserting into it, the message is left unchanged.

### Screenshots

screenshot-2026-07-04_13-33-07

screenshot-2026-07-04_13-31-38

---

- Closes #2595

### DIFF
--- a/core/deployment/src/main/resources/dev-ui/qwc-chat.js
+++ b/core/deployment/src/main/resources/dev-ui/qwc-chat.js
@@ -15,6 +15,20 @@ import 'qui-alert';
 import { JsonRpc } from 'jsonrpc';
 import { systemMessages } from 'build-time-data';
 import { dialogHeaderRenderer, dialogRenderer } from '@vaadin/dialog/lit.js';
+import { registerStyles, css as vaadinCss } from '@vaadin/vaadin-themable-mixin/register-styles.js';
+
+// Styles the RAG context box (see ChatMessagePojo#formatRagAugmentedText) as a distinct box in the same bubble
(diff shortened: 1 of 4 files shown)


jmartisk (approved):
Ok, I think it may break if the augmentor changes something instead of just adding stuff, but hopefully that shouldn't be common. Thanks

PR #2662, opened 2026-07-06
fix: simplify CDI instance selection in SampleLoaderResolver

### LINKED ISSUE #2661: CDI lookup of `SampleLoader<SomeType>` doesn't work
Registering a custom sample loader via CDI (see https://docs.quarkiverse.io/quarkus-langchain4j/dev/testing.html#_registering_via_cdi) only seems to work if my sample loader implements `SampleLoader` directly, without any generic types.

`SampleLoader` does have a generic type, so if I do something like this:

```java
@ApplicationScoped
public class MySampleLoader implements SampleLoader<String> {

}
```

The loading mechanism does not find it. I'm forced to use the service loader SPI.

matheusandre1:
The class does not have a parameterized type; the bean is registered as the concrete class `SampleLoader`. When using `TypeLiteral<SampleLoader<?>>`, CDI looks for `SampleLoader<?>` or a wildcard, whereas in Quarkus Arc, beans are registered with the exact type `SampleLoader<String>`, not a wildcard.

- Closes: #2661

### DIFF
--- a/testing/evaluation/evaluation-junit5/src/test/java/io/quarkiverse/langchain4j/evaluation/junit5/it/QuarkusCustomCdiLoaderIT.java
+++ b/testing/evaluation/evaluation-junit5/src/test/java/io/quarkiverse/langchain4j/evaluation/junit5/it/QuarkusCustomCdiLoaderIT.java
@@ -7,7 +7,9 @@ import org.junit.jupiter.api.extension.ExtendWith;
 
 import io.quarkiverse.langchain4j.evaluation.junit5.EvaluationExtension;
 import io.quarkiverse.langchain4j.evaluation.junit5.SampleLocation;
+import io.quarkiverse.langchain4j.testing.evaluation.SampleLoader;
 import io.quarkiverse.langchain4j.testing.evaluation.Samples;
+import io.quarkus.arc.Arc;
 import io.quarkus.test.junit.QuarkusTest;
 
 /**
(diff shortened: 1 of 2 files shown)


geoand (commented):
Makes sense, but it would be great to have a test as well

cescoffier:
@edeandrea That's the trick I was talking about :-)

testing/evaluation/evaluation-junit5/src/test/java/io/quarkiverse/langchain4j/evaluation/junit5/it/QuarkusCustomCdiLoaderIT.java
```diff
@@ -87,6 +88,15 @@ void shouldLoadMultipleSampleSourcesInSameTest(
...
     }
 
+    @Test
+    void shouldDiscoverGenericCdiSampleLoaderViaResolver() {
+        Samples<String> samples = SampleLoaderResolver.load("custom:greetings", String.class);
+
+        assertThat(samples).isNotNull().hasSize(3);
+        assertThat(samples.get(0).name()).isEqualTo("Greeting1");
+        assertThat(samples.get(0).parameters().<String> get(0)).isEqualTo("Hello");
+    }
```
geoand, line 101: How does this test the original issue?

matheusandre1: That makes sense; I forgot to call the CDI.

PR #2666, opened 2026-07-07
Add ability to contribute additional info to tool spans

edeandrea:
This was modeled after `ChatModelSpanContributor`

- Introduced `ToolExecutionRequestContext`, `ToolExecutionResponseContext`, and `ToolExecutionErrorContext` to encapsulate tool execution lifecycle data.
- Added `ToolSpanContributor` interface for customizing span behavior during tool execution.
- Updated `ToolSpanWrapper` to enhance tracing capabilities by notifying contributors on tool request, response, and error events.
- Created unit tests to validate span creation, including successful and failure cases.

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolExecutionErrorContext.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolExecutionErrorContext.java
@@ -1,0 +1,100 @@
+package io.quarkiverse.langchain4j.runtime.tool;
+
+import java.util.Map;
+
+/**
+ * Represents the response context of a tool execution, encapsulating the
+ * associated request context and the resulting execution outcome.
+ */
+public sealed interface ToolExecutionErrorContext permits DefaultToolExecutionErrorContext {
+    /**
+     * Creates and returns a new instance of the {@code Builder} class for constructing
+     * an instance of {@code ToolExecutionErrorContext}.
+     */
+    static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * Retrieves the context of the tool execution request, which provides details
+     * about the request and invocation.
+     */
+    ToolExecutionRequestContext requestContext();
+
+    /**
+     * Retrieves the error associated with the tool execution context.
+     * This method provides information about any error that occurred during the tool's execution.
+     *
+     * @return a {@code Throwable} representing the error, or {@code null} if no error occurred.
+     */
+    Throwable error();
+
+    /**
+     * Retrieves a map of attributes associated with the current tool execution request context.
+     * These attributes provide additional metadata or details relevant to the execution context.
+     *
+     * @return an unmodifiable map of attributes for the current request context.
+     */
+    default Map<Object, Object> attributes() {
+        return requestContext().attributes();
+    }
+
+    /**
+     * Creates a new {@code Builder} instance initialized with the current state of this object.
+     * This allows for modifying and regenerating the object while preserving the existing values.
+     */
+    default Builder toBuilder() {
+        return new Builder(this);
+    }
+
+    /**
+     * A builder class for constructing instances of {@code ToolExecutionErrorContext}.
+     * This builder allows for step-by-step construction of a {@code ToolExecutionErrorContext} object
+     * by providing methods to set its component properties.
+     */
+    class Builder {
+        ToolExecutionRequestContext requestContext;
+        Throwable error;
+
+        private Builder() {
+        }
+
+        private Builder(ToolExecutionErrorContext context) {
+            this.requestContext = context.requestContext();
+            this.error = context.error();
+        }
+
+        /**
+         * Sets the {@link DefaultToolExecutionRequestContext} for the builder.
+         *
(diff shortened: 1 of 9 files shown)


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolExecutionErrorContext.java
```diff
@@ -0,0 +1,112 @@
+package io.quarkiverse.langchain4j.runtime.tool;
+
+import java.util.Map;
+
+/**
+ * Represents the response context of a tool execution, encapsulating the
+ * associated request context and the resulting execution outcome.
+ */
+public class ToolExecutionErrorContext {
```
geoand: In general, context objects like this should be interfaces to allow for their evolution without breaking user code in any way.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolSpanWrapper.java
```diff
@@ -13,27 +15,37 @@
 import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.context.Scope;
+import io.quarkus.arc.All;
 
 public class ToolSpanWrapper implements QuarkusToolExecutor.Wrapper {
+    private static final Logger LOG = Logger.getLogger(ToolSpanWrapper.class);
+    private static final String OTEL_SCOPE_KEY_NAME = "OTelScope";
+    private static final String OTEL_SPAN_KEY_NAME = "OTelSpan";
+    public static final String OTEL_PARENT_SPAN_KEY_NAME = "OTelParentSpan";
```
geoand, line 24: Where do these names come from?

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolSpanWrapper.java
```diff
@@ -42,21 +54,85 @@ public ToolExecutionResult wrap(ToolExecutionRequest toolExecutionRequest, Invoc
...
+
+    private void notifyContributorsOnError(ToolExecutionErrorContext errorContext, Span span) {
+        this.toolSpanContributors.forEach(contributor -> {
+            try {
+                contributor.onError(errorContext, span);
+            } catch (Exception ex) {
+                recordLogAndSwallow(span, ex);
+            }
+        });
+    }
```
geoand, line 132: No lambdas please

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolSpanWrapper.java
edeandrea: It's consistent with what's used in https://github.com/quarkiverse/quarkus-langchain4j/blob/main/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/ChatModelSpanContributor.java

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolExecutionErrorContext.java
edeandrea: It's consistent with how `ChatModelRequestContext`/`ChatModelResponseContext`/`ChatModelErrorContext` are created, which themselves are not interfaces.

Because they are builders they can evolve without breaking backwards compatibility.

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolSpanWrapper.java
edeandrea: Done

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/tool/ToolExecutionErrorContext.java
edeandrea: If you'd really prefer them to be interfaces I can do that. I was just following existing patterns.

geoand: > It's consistent with how 

It's consistently bad. 

> Because they are builders they can evolve without breaking backwards compatibility.

Not true.

geoand: > If you'd really prefer them to be interfaces I can do that. I was just following existing patterns.

Despite the fact that this is not the best API design, we can let it slide as the same (subpar) pattern is used in LangChain4j

geoand (approved):

edeandrea:
@geoand please don't merge yet. I found 1 thing i need to fix

geoand:
NP. I'll wait for your signal

edeandrea:
All set. I've switched it to interfaces as well.

geoand:
Great, thanks!

geoand (approved):

Issue #2668, opened 2026-07-08
Agentic Dev UI: richer Testing page and AgenticScope state in execution traces


omatheusmesmo:
Two related gaps in the agentic Dev UI that make workflow debugging harder. Raising them together to check whether they fit your plans before proposing a contribution.

1. Testing page - complex parameters and multi-turn.

   The Testing page only handles primitive parameters; a POJO parameter falls back to `getString` and breaks. It also creates a fresh `memoryId` per invocation, so a multi-turn conversation can't be exercised. Proposal: accept a JSON object for non-primitive parameters (deserialized to the parameter type) and add an option to reuse the `memoryId`.

2. Execution traces - AgenticScope state.

   Traces show output, tokens and tool calls, but not the AgenticScope key/value state each sub-agent writes via `outputKey` - the state that drives the next agent. `MonitoredExecution.agenticScope().state()` already exposes it, so this is a serialization + rendering change: show the scope per invocation in an expandable panel.

Both need the Dev UI Executions page to be client-rendered from JSON (today it renders server-side HTML via the upstream HtmlReportGenerator), which ties into the service issue above.

Does this direction make sense, or is something similar already planned? Happy to contribute.

---

Relates to #2667

geoand:
cc @mariofusco

omatheusmesmo:
@mariofusco , do you mind if I start to work on it and you review later? Thanks!

mariofusco:
@omatheusmesmo I don't have time for this now, so I would really appreciate it if you could move this forward, thanks a lot.

Issue #1898, opened 2025-10-28
Agentic - Developer Experience & Tooling


cescoffier:
* Quarkus Dev UI integration:

   - Visualize workflows
   - Trace agent invocations and tool calls

* Build-time validation for LangChain4j constructs (missing tools, bad metadata, etc.)
* Evaluation harnesses and support for repeatable agent test scenarios

From https://github.com/quarkusio/quarkus/discussions/50582.

PR #2674, opened 2026-07-10
Make the agentic Dev UI monitoring populate Topology and Executions

### LINKED ISSUE #2667: Agentic Dev UI: invokeAgent allow-list is dead code (a duplicate service shadows the security fix)
Commit e74850f36 added an `invokeAgent` allow-list to `AgenticJsonRpcService` and moved service to a new `quarkus-langchain4j-agentic-dev` module wired as a `conditionalDevDependency`. The previous copy stayed in `quarkus-langchain4j-agentic` under the same FQN, and
`agentic/deployment` still compiles against it.

In dev mode both classes are on the classpath and the old copy shadows the new one, so the live `invokeAgent` runs without the allow-list: the guard is inert. The Dev UI pages still render (they call the old HTML methods), so nothing looks broken.

Reproduce (fails on current main). Add to agentic/deployment/pom.xml:

```xml
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-devui-test-spi</artifactId>
      <scope>test</scope>
    </dependency>
```

New test at agentic/deployment/src/test/java/io/quarkiverse/langchain4j/agentic/deployment/devui/AgenticDevUiAllowListTest.java
(standard DevUIJsonRPCTest / QuarkusDevModeTest / langchain4j imports):

```java
    public class AgenticDevUiAllowListTest extends DevUIJsonRPCTest {

        @RegisterExtension
        static final QuarkusDevModeTest devMode = new QuarkusDevModeTest()
                .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
                        .addClasses(Consumer.class, EchoAgent.class, Wrapper.class, EchoModel.class)
                        .addAsResource(new StringAsset(
                                "quarkus.langchain4j.openai.echo.api-key=echo\n"
                                + "quarkus.langchain4j.openai.base-url=http://localhost:8089/v1"),
                                "application.properties"));

        public AgenticDevUiAllowListTest() {
            super("quarkus-langchain4j-agentic", "http://localhost:8080");
        }

        @Test
        void invokeAgentRejectsNonAgentClass() throws Exception {
            JsonNode r = executeJsonRPCMethod("invokeAgent", Map.of(
                    "agentClassName", "java.lang.String", "methodName", "toString", "inputJson", "{}"));
            assertThat(r.get("error").asText()).contains("Unknown agent class");
        }

        @Singleton static class Consumer { @Inject Wrapper wrapper; }

        interface EchoAgent {
            @UserMessage("{{text}}") @Agent(description = "echo", outputKey = "echo") String echo(String text);
            @ChatModelSupplier static ChatModel model() { return new EchoModel(); }
        }
        interface Wrapper {
            @SequenceAgent(outputKey = "echo", subAgents = EchoAgent.class) String run(String text);
        }
        static class EchoModel implements ChatModel {
            public ChatResponse doChat(ChatRequest req) {
                return ChatResponse.builder().aiMessage(new AiMessage("echo")).build();
            }
        }
    }
```

On main the call returns an UnsatisfiedResolutionException (it tried to instantiate the arbitrary class) instead of the "Unknown agent class" rejection, showing the un-guarded copy is live.

The JSON rewrite in runtime-dev looks like work in progress (the JS was never migrated to the JSON methods). Two ways to resolve, happy to send a PR either way:

- Minimal, no visual change: keep the working HTML rendering but as a single dev-only service (move it to runtime-dev with the allow-list, delete the runtime copy, make deployment depend on runtime-dev, gate the provider on dev mode). Restores the guard, UI unchanged.
- Or finish the JSON migration started (client-render the Executions/Topology pages).

Which direction do you prefer?

omatheusmesmo:
## Describe your changes

Consolidates the agentic Dev UI to the single dev-only JSON-RPC service (restoring the `invokeAgent` allow-list) and makes the monitoring actually populate, so Topology and Executions render live data instead of nothing.

`DevAgentMonitorHolder` was never filled: monitoring was gated on a filter a declarative agent interface can't satisfy, and no `AgentMonitor` was ever attached. A dev-only `DevModeAgentMonitor` CDI bean is now wired into every agent through the additive `AgentListener` mechanism and injected into the RPC service, the same store-via-injection shape quarkus-flow uses.

Both pages move to Lit over JSON-RPC. Executions is a per-run waterfall grouped by `memoryId`, with the workflow as the parent row and its sub-agents nested; the parent is synthesized on the client from the shared `memoryId`, since the generated workflow agent does not emit its own invocation event. Topology is an ECharts graph laid out by workflow type (sequence chains vertically, parallel fans out, loops draw their repeat back-edge), rendered client-side from coordinates the service computes. Agents are labelled by class name and annotation type across all three pages, reusing the annotation detection the Agents page already does.

A project with more than one agentic system keeps them separate: each root's report is scoped to that root, matched by its subtree, and when two systems share a sub-agent type they are told apart by the `memoryId` tagged at invocation.

The `DevUIJsonRPCTest`s invoke an agent through the allow-listed path and assert the report is scoped to the selected root (including when a sub-agent is shared) and that the allow-list rejects a non-agent class, offline against a local echo model.

---

#### Topology, sequence and loop workflows:

devui-topology-loop
devui-topology-parallel
devui-topology-sequence

#### Executions, per-run waterfall (the loop shows the score-driven iterations):

devui-executions
devui-executions-loop

#### Agents:

devui-agents

---

- Closes #2667
- Relates to #1898
- Relates to #2668


agentic/deployment/src/main/resources/dev-ui/qwc-agents-topology.js
```diff
@@ -1,9 +1,12 @@
 import { LitElement, html, css } from 'lit';
 import { JsonRpc } from 'jsonrpc';
+import { themeState } from 'theme-state';
 import '@vaadin/button';
 import '@vaadin/select';
 import '@vaadin/progress-bar';
 
+const MERMAID_CDN = 'https://cdn.jsdelivr.net/npm/mermaid@11.12.0/dist/mermaid.min.js';
```
phillip-kruger: Don't use a CDN. Use mvnpm. Also for diagram we already have apache echart available.

phillip-kruger (changes requested):

omatheusmesmo:
Quick update: I have the fix on my personal machine, but I will send by the middle of the next week, I'm attending an in person event.

omatheusmesmo:
I'm facing a issue on my main machine with the right fixes and I was not able to fix yet, I will try to finish this by the end of the next week.

agentic/deployment/src/main/resources/dev-ui/qwc-agents-topology.js
omatheusmesmo: Hi @phillip-kruger , I so sorry for taking so long, I had a serious problem with my main machine's HDD where I had the fixes, I did not want to start again from scratch and took some time to get fixed.

Now it's ready for review and using echart which gaves a great UI.

phillip-kruger (approved):
LGTM from a Dev UI p.o.v

omatheusmesmo:
Hi @jmartisk @quarkiverse/quarkiverse-langchain4j , shall we merge this one given the last review? :smile:

geoand (approved):
Thanks!

mariofusco:
This change went completely out of my radar (I think I was on PTO when it has been merged) and to be honest I'm not understanding the reasons behind it. @omatheusmesmo can you please clarify why you implemented it in this way?

My main concern is that before the report was generated as a simply HTML page on the LangChain4j side and the dev-ui just rendered exactly the same HTML.

image

Now the 2 visualizations, the one generated by LangChain4j and the one in the dev-ui, are totally different and quite inconsistent and I'm afraid this could confuse users, other than adding the burden of maintaining 2 different implementations instead of reusing a single one. If we can convey more dynamic features and information in the dev-ui than in the static page generated by LangChain4j I'm totally fine and happy of this, but the 2 visualizations should look as similar and consistent as possible, and possibly also reuse the same code. 

In essence, if we don't have any evidence of the advantages and necessity of this second UI, I would suggest to revert this pull request, or in alternative to generate its static part from LangChain4j as it did before.

/cc @geoand @phillip-kruger

agentic/runtime-dev/src/main/java/io/quarkiverse/langchain4j/agentic/runtime/devui/DevModeAgentMonitor.java
```diff
@@ -0,0 +1,46 @@
...
+ * Dev-only CDI {@link AgentMonitor} wired into every agent through the additive AgentListener
+ * mechanism, capturing invocations so the Dev UI Executions page can render them.
+ * <p>
+ * A single monitor is shared across every agentic system, so it also records which root agent
+ * started each run: when the Dev UI invokes a root, {@link #markPendingRoot(String)} tags the
+ * run's memoryId (captured from the first invocation on the calling thread) with that root.
+ * This attributes executions to the correct root even when two systems share a sub-agent type.
+ */
+@Singleton
+public class DevModeAgentMonitor extends AgentMonitor {
```
mariofusco, line 21: Also the introduction of this new `AgentMonitor` implementation is breaking the internal LangChain4j monitoring system because  both implementations are instanced for the same request randomly registering the agent invocations on one or another, thus occasionally causing error like the following:

```
2026-09-10 11:40:17,197 ERROR [dev.langchain4j.agentic.observability.ListenerNotifierUtil] () beforeAgentInvocation listener for agent planItinerary failed: No ongoing parent invocation found for agent ID: research$0: java.lang.IllegalStateException: No ongoing parent invocation found for agent ID: research$0
	at dev.langchain4j.agentic.observability.MonitoredExecution.beforeAgentInvocation(MonitoredExecution.java:30)
	at dev.langchain4j.agentic.observability.AgentMonitor.beforeAgentInvocation(AgentMonitor.java:109)
	at io.quarkiverse.langchain4j.agentic.runtime.devui.DevModeAgentMonitor.beforeAgentInvocation(DevModeAgentMonitor.java:44)
	at dev.langchain4j.agentic.observability.ComposedAgentListener.beforeAgentInvocation(ComposedAgentListener.java:56)
	at dev.langchain4j.agentic.observability.ListenerNotifierUtil.beforeAgentInvocation(ListenerNotifierUtil.java:20)
	at dev.langchain4j.agentic.internal.AgentInvoker.invoke(AgentInvoker.java:34)
	at dev.langchain4j.agentic.internal.AbstractAgentInvoker.invoke(AbstractAgentInvoker.java:88)
	at dev.langchain4j.agentic.internal.AgentExecutor.agentResponse(AgentExecutor.java:130)
	at dev.langchain4j.agentic.internal.AgentExecutor.internalExecute(AgentExecutor.java:82)
	at dev.langchain4j.agentic.internal.AgentExecutor.execute(AgentExecutor.java:40)
	at dev.langchain4j.agentic.internal.AgentExecutor.execute(AgentExecutor.java:28)
	at dev.langchain4j.agentic.internal.PlannerBasedInvocationHandler$PlannerLoop.lambda$parallelExecution$1(PlannerBasedInvocationHandler.java:488)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1789)
	at java.base/java.util.concurrent.ThreadPerTaskExecutor$TaskRunner.run(ThreadPerTaskExecutor.java:291)
	at java.base/java.lang.VirtualThread.run(VirtualThread.java:456)
```

I'm sorry, but at this point I strongly suggest to revert this commit.

geoand: @mariofusco go ahead and propose a revert and I will approve it

omatheusmesmo:
Thanks for taking the time to look at this @mariofusco , and sorry it landed while you were away.

I opened #2667 because the invokeAgent allow-list added in e74850f3 was dead code, shadowed by a second copy of AgenticJsonRpcService under the same FQN. I laid out both options there, a minimal fix keeping the server-rendered HTML or finishing the JSON migration, and asked which one you preferred. @phillip-kruger chose the migration, on the grounds that #2668 (POJO parameters, multi-turn testing, AgenticScope state in traces) needs client-rendered pages as a prerequisite, and @geoand agreed.

The technical case is mostly testability. The DevUIJsonRPCTests in this PR assert real behaviour, per-root scoping and allow-list rejection, which is not possible against an opaque HTML string, and that is exactly how the allow-list bug stayed invisible for weeks: the page rendered correctly while the service behind it was dead. The report is also a full HTML document with its own :root palette and no prefers-color-scheme rule, so embedded in the Dev UI it cannot follow the Lumo theme.

Your criticism of the Topology page is fair, and that part is on me. I changed the visual metaphor from cards to a node-link graph and dropped what the report shows: description, return type, planner, loop bounds and exit condition, router conditions, and the input/output key badges. I did reuse your palette and topology labels, and the Executions page kept the report's columns, badges and layout, but I did not hold Topology to that same standard.

If you are open to it, I would redo that page as cards mirroring the report's structure and restore those fields, so both views read the same, and keep a way to open the LangChain4j report itself from the Dev UI.

agentic/runtime-dev/src/main/java/io/quarkiverse/langchain4j/agentic/runtime/devui/DevModeAgentMonitor.java
omatheusmesmo: @mariofusco @geoand I looked into the crash. It reproduces in plain LangChain4j with no Quarkus involved: an AgentMonitor attached to the sub-agents of a @ParallelAgent instead of its root cannot resolve the parent invocation of the second overlapping sub-agent, and MonitoredExecution.beforeAgentInvocation throws. Filed as langchain4j/langchain4j#6360 with a self-contained reproducer that failed on 5 out of 5 runs. I did not claim it is the same trigger as yours: yours is a planner with research$0 as the missing parent, and a parent that already finished would produce the same error, so there is more than one route to it.

On this side, CDI AgentListener beans are only attached to leaf AI agents, so the dev monitor ended up below the root. Reverting was the right call either way. Once upstream decides how the monitor should behave, I can bring it back attached at the root, together with the Topology changes.

mariofusco: Thanks for reproducer on parallel agents, I will give a look immediately, but yes this is a completely different problem of the one caused by the new `AgentMonitor` implementation introduced by this pull request.

mariofusco:
> If you are open to it, I would redo that page as cards mirroring the report's structure and restore those fields, so both views read the same, and keep a way to open the LangChain4j report itself from the Dev UI.

Sure, that would be great, please feel free to go forward with this.

omatheusmesmo:
> > If you are open to it, I would redo that page as cards mirroring the report's structure and restore those fields, so both views read the same, and keep a way to open the LangChain4j report itself from the Dev UI.
> 
> Sure, that would be great, please feel free to go forward with this.

Thank you!

PR #2675, opened 2026-07-10
Fix missing thinking events for Ollama streaming

### LINKED ISSUE #2660: Missing `PartialThinkingEvent` in Streaming Response
Experienced with Quarkus 3.37.1 and the Ollama adapter.

Returning a `Multi<ChatEvent>` from an `AiService` method does not include any `PartialThinkingEvent` items.

When starting the example application and sending a message to a local model capable of thinking via the SSE endpoint, the application logs a thinking part in the console. However, the SSE connection does not return a `PartialThinkingEvent`; it only returns `PartialResponseEvent` items.

Sample application with Ollama and an SSE endpoint:
[code-with-quarkus.zip](https://github.com/user-attachments/files/29712642/code-with-quarkus.zip)

arnabnandy7:
## Description

Fixes missing thinking events when an AI service backed by Ollama returns a `Multi<ChatEvent>`.

Ollama returns reasoning from thinking-capable models in the streamed `message.thinking` field. The custom Ollama streaming implementation did not model this field and only forwarded `message.content`, so thinking chunks were discarded before they could reach the AI service streaming layer.

This change:

- Adds the `thinking` field to the Ollama `Message` response DTO.
- Forwards non-empty thinking chunks through `StreamingChatResponseHandler.onPartialThinking(...)`.
- Accumulates streamed thinking and includes it in the completed `AiMessage`.
- Preserves accumulated thinking when the response contains tool execution requests.
- Adds a WireMock-based regression test covering:
  - ordered `PartialThinkingEvent` emissions;
  - regular `PartialResponseEvent` emissions;
  - accumulated response text in the completed event;
  - accumulated thinking in the completed `AiMessage`.

This makes Ollama consistent with the existing streaming behaviour used by providers such as Anthropic.

- Fixes: #2660

### DIFF
--- a/model-providers/ollama/deployment/src/test/java/io/quarkiverse/langchain4j/ollama/deployment/OllamaStreamingChatLanguageModelSmokeTest.java
+++ b/model-providers/ollama/deployment/src/test/java/io/quarkiverse/langchain4j/ollama/deployment/OllamaStreamingChatLanguageModelSmokeTest.java
@@ -26,6 +26,7 @@ import dev.langchain4j.data.message.ToolExecutionResultMessage;
 import dev.langchain4j.service.MemoryId;
 import dev.langchain4j.store.memory.chat.ChatMemoryStore;
 import io.quarkiverse.langchain4j.RegisterAiService;
+import io.quarkiverse.langchain4j.runtime.aiservice.ChatEvent;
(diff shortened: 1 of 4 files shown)


geoand (approved):
Thanks!

PR #2676, opened 2026-07-10
Fix native compatibility with Quarkus main #2645

### LINKED ISSUE #2645: Native image build fails against the Quarkus `main` snapshot (`999-SNAPSHOT`)
Native image build fails against the Quarkus `main` snapshot (`999-SNAPSHOT`) because three `io.quarkiverse.langchain4j.*` classes call methods that no longer exist on `main`:

- `io.quarkus.tls.TlsConfiguration.getSSLOptions()` — from `JaxRsHttpClient` (`JaxRsHttpClient.java:69`)
- `io.vertx.core.Context.getLocal(Object)` — from `OllamaRestApi$OllamaRestApiReaderInterceptor`
- `io.vertx.core.Context.putLocal(Object, Object)` — from `ContextLocals`

The TLS registry API changed and Vert.x 5 removed the `Context.getLocal/putLocal` method. Since these classes are linked at build time, native image analysis fails with `Unsupported features in 3 methods`.

```
Error: Unsupported features in 3 methods
Error: Discovered unresolved method during parsing: io.quarkus.tls.TlsConfiguration.getSSLOptions()
Error: Discovered unresolved method during parsing: io.vertx.core.Context.getLocal(java.lang.Object)
Error: Discovered unresolved method during parsing: io.vertx.core.Context.putLocal(java.lang.Object, java.lang.Object)
```

Released Quarkus versions (e.g. LTS 3.33) are unaffected; only the build against `main` fails.

- quarkus-langchain4j: 1.11.2
- Quarkus: main / 999-SNAPSHOT
- CI run (quarkus-flow): https://github.com/quarkiverse/quarkus-flow/actions/runs/28343711568/job/83962964886

arnabnandy7:
Fixes #2645.

Quarkus `main` uses Vert.x 5, which removed the object-keyed context-local methods and changed the TLS configuration API. These binary incompatibilities caused native image analysis to fail when using an extension built against Quarkus 3.33.

This change:

- uses SmallRye's context-local compatibility API instead of calling Vert.x directly;
- resolves client SSL options across both TLS configuration APIs;
- registers the TLS configuration methods for reflection in native images;
- updates the affected streaming tests to use the same context-local API.

The public configuration and behaviour remain unchanged.

Tests:

- JAX-RS client tests
- Ollama runtime tests
- `BlockingMemoryStoreOnStreamedResponseTest`
- Core deployment test compilation

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/streaming/BlockingMemoryStoreOnStreamedResponseTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/streaming/BlockingMemoryStoreOnStreamedResponseTest.java
@@ -22,6 +22,7 @@ import dev.langchain4j.service.UserMessage;
 import io.quarkiverse.langchain4j.RegisterAiService;
 import io.quarkus.arc.Arc;
 import io.quarkus.test.QuarkusUnitTest;
+import io.smallrye.common.vertx.ContextLocals;
 import io.smallrye.common.vertx.VertxContext;
 import io.smallrye.mutiny.Multi;
 import io.vertx.core.Context;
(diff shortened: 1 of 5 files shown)


geoand:
Thanks for this, but I do plan to handle this myself once the Jackson 3 stuff lands in Quarkus main (which will probably be a while)

PR #2680, opened 2026-07-16
Add direct AI service configuration attributes

### LINKED ISSUE #2577: feat: @MemoryConfig and @ToolConfig composition annotations
## Parent epic
Part of #2572

## Summary
Group memory-related and tool-related attributes currently scattered across `@RegisterAiService` into focused composition annotations.

**`@MemoryConfig`** — groups chat memory provider, memory window, memory ID attributes.

**`@ToolConfig`** — groups tool provider, tool classes, and tool-related attributes.

## Rationale
These are the simpler annotations from #2572. They reduce the attribute count on `@RegisterAiService` and make each concern independently composable.

## Acceptance criteria
- [ ] `@MemoryConfig` annotation defined, grouping all memory-related attributes
- [ ] `@ToolConfig` annotation defined, grouping all tool-related attributes
- [ ] Both use bean-class references (not supplier-class)
- [ ] Both processed by the same build step as `@RegisterAiService`

arnabnandy7:
## Summary

Adds direct CDI bean configuration to the existing `@RegisterAiService` annotation, without introducing additional
top-level annotations.

Applications can now select concrete memory and tool infrastructure beans directly instead of creating supplier wrappers.
Existing supplier attributes remain supported for compatibility.

### Memory

- `chatMemoryProvider`
- `chatMemoryStore`
- `chatMemoryMaxMessages`
- `defaultMemoryIdProvider`
- `chatMemoryFlushStrategy`

When a store or message limit is configured, the extension creates a service-specific `MessageWindowChatMemory` backed
by the selected `ChatMemoryStore`. A configured default memory-ID provider is used only when an invocation has no explicit
memory ID.

### Tools

- `toolProvider`
- `toolSearchStrategy`

Existing attributes continue to configure static tools, hallucination handling, calling limits, and continuous forced
calling:

- `tools`
- `toolHallucinationStrategy`
- `maxToolCallingRoundTrips`
- `maxToolCallsPerResponse`
- `allowContinuousForcedToolCalling`

## CDI integration

Direct class references are resolved as CDI beans through the synthetic AI-service context. This preserves CDI scopes,
lifecycle, interception, and build-time validation. Missing referenced beans fail during augmentation.

## Compatibility

The supplier-based attributes remain supported:

- `chatMemoryProviderSupplier`
- `chatMemoryFlushStrategySupplier`
- `toolProviderSupplier`
- `toolSearchStrategySupplier`

A direct bean attribute and its corresponding supplier attribute cannot both be configured.

Additional validation rejects:

- combining `chatMemoryProvider` with `chatMemoryStore` or `chatMemoryMaxMessages`
- non-positive `chatMemoryMaxMessages`
- negative tool-call limits
- direct references that do not resolve to CDI beans

## Examples

```java
@RegisterAiService(
    chatMemoryStore = RedisChatMemoryStore.class,
    chatMemoryMaxMessages = 20,
    defaultMemoryIdProvider = TenantMemoryIdProvider.class,
    chatMemoryFlushStrategy = ChatMemoryFlushStrategy.IMMEDIATE
)
interface Assistant {
}
```

```java
@RegisterAiService(
    tools = { BookingTool.class, WeatherTool.class },
    toolProvider = DynamicToolProvider.class,
    toolSearchStrategy = SemanticToolSearchStrategy.class,
    maxToolCallingRoundTrips = 5
)
interface Assistant {
    String chat(String message);
}
```

- Closes: #2577

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -446,6 +446,10 @@ public class AiServicesProcessor {
             }
             alreadyHandled.add(declarativeAiServiceClassInfo.name());
 
+            boolean chatMemoryProviderClassIsBean = false;
+            boolean toolProviderClassIsBean = false;
+            boolean toolSearchStrategyClassIsBean = false;
+
             DotName chatLanguageModelSupplierClassDotName = getSupplierDotName(instance.value("chatLanguageModelSupplier"),
                     LangChain4jDotNames.BEAN_CHAT_MODEL_SUPPLIER,
                     supplierDotName -> validateSupplierAndRegister(
@@ -494,7 +498,15 @@ public class AiServicesProcessor {
 
             DotName toolProviderClassName = LangChain4jDotNames.BEAN_IF_EXISTS_TOOL_PROVIDER_SUPPLIER;
             AnnotationValue toolProviderValue = instance.value("toolProviderSupplier");
-            if (toolProviderValue != null) {
+            AnnotationValue configuredToolProvider = instance.value("toolProvider");
+            rejectConflictingConfiguration(declarativeAiServiceClassInfo, toolProviderValue, configuredToolProvider,
+                    "toolProviderSupplier", "toolProvider");
+            if (configuredToolProvider != null) {
+                toolProviderClassName = configuredToolProvider.asClass().name();
+                toolProviderClassIsBean = true;
+                toolProviderInfos.add(new ToolProviderInfo(toolProviderClassName.toString(),
+                        declarativeAiServiceClassInfo.simpleName()));
+            } else if (toolProviderValue != null) {
                 if (LangChain4jDotNames.NO_TOOL_PROVIDER_SUPPLIER.equals(toolProviderValue.asClass().name())) {
                     toolProviderClassName = null;
                 } else {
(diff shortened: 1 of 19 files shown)


arnabnandy7:
All failures came from `McpRegistryClientTest` timing out after 10 seconds while contacting `registry.modelcontextprotocol.io`. Different registry requests failed across the Java versions, which strongly indicates an external registry/network outage or flakiness rather than my changes.

geoand:
Thanks for this, but I really want to avoid annotation proliferation. That is the reason why currently (almost) everything is configured via `@RegisterAiService`

arnabnandy7:
> Thanks for this, but I really want to avoid annotation proliferation. That is the reason why currently (almost) everything is configured via `@RegisterAiService`

Let me check on this, thanks for pointing out

arnabnandy7:
@geoand it seems issue #2577 explicitly proposes `@MemoryConfig` and `@ToolConfig`, so I implemented those composition annotations while preserving all existing `@RegisterAiService` attributes. 

Would you prefer that the issue be reconsidered and the configuration remain consolidated in `@RegisterAiService`, or is there another composition mechanism you would accept without adding top-level annotations? Please share your thoughts, I'll refactor or redesign based on the same if plausible.

geoand:
The issue does propose that, but I never said I want it done that way :).

In general, I don't want a ton of annotations that make it hard for users to discover what is available.
Attributes on `@RegisterAiService` solve that problem.

arnabnandy7:
Understood, @geoand thanks for clarifying. I’ll rework the proposal around `@RegisterAiService` and retain the direct CDI bean-class approach where it improves the existing supplier-based configuration. I’ll first map the requested memory and tool settings onto the existing attributes to avoid unnecessary duplication.

arnabnandy7:
@geoand I've refactored the code accordingly as suggested.

geoand:
I like it. WDYT @cescoffier @jmartisk ?

One thing we could also think of doing is moving all the related settings using nested annotations...

cescoffier:
I fear that when we will need to add more to AIServices we will have an explosion of attributes. 

I like that fact that it simplifies discovery and conflict detection.

We can merge this, but soon (post summer) we need to redesign how AI Service configuration works and is expressed, especially memory and rag.

cescoffier:
BTW +1 for the nested annotation, I think it's the right "move"

geoand:
Actually, I'm probably going to leave this open and use it as a basis to overhaul the configuration - after the summer.

PR #2681, opened 2026-07-17
Add AI service conversation resume support

### LINKED ISSUE #2677: Resume a conversation without a user prompt
Hi all,

I've put together a blocking human-in-the-loop implementation, where we store all ChatMessages in a DB, then, when ou rHITL tool is called it makes a snapshot of context data it required to resume later on, and returns with ReturnBehavior.IMMEDIATE.

I store this AI ToolRequest message in the DB and wait for user response eg fill in a form. Once the user submits the form, I store the user input in a ToolResponse ChatMessage which I create manually.

There is a custom ChatModelStore, to reconstruct the context so all messages are ready to be submitted to LLM.

However, AIService only expose a mechanism to perform an agent turn by passing a UserMessage as an argument.

Is it possible for QuarkusLangchian4j to provide a mechanism to resume an agent loop using existing ChatMessages?

In vanilla LC4J, one would use the low-level API -> chatModel.chat(request).

To get the right abstraction might be tricky to implement though?

Kind regards

Bob

arnabnandy7:
## Summary

This PR adds `@ResumeConversation` for declarative AI Services, allowing an existing conversation to continue from persisted chat memory without adding a synthetic `UserMessage`.

```java
@RegisterAiService
interface Assistant {

    String chat(@MemoryId String conversationId, @UserMessage String message);

    @ResumeConversation
    String resume(@MemoryId String conversationId);
}
```

Calling `resume(conversationId)` loads the messages already associated with that memory ID and submits them to the model through the normal AI Service execution path.

## Why

Applications can receive or persist conversation messages outside a regular user-prompt invocation. A human-in-the-loop tool response is one example, but the requirement is more general: continue an existing conversation without recording or displaying a user message that was never entered by the user.

Using the low-level `ChatModel` API can submit the messages directly, but it bypasses the declarative AI Service abstraction and requires the application to reproduce its orchestration behavior.

This is separate from LangChain4j's agentic suspension and resume support. That feature operates on agentic workflows and `AgenticScope`; this PR supports applications using plain declarative AI Services and `ChatMemory`, without requiring the agentic module.

## Behavior

A method annotated with `@ResumeConversation`:

- loads the existing messages from the configured chat memory;
- does not create, append, or send a synthetic user message;
- supports both tool-result histories and other model-compatible message sequences;
- continues through the normal model and tool-calling loop;
- preserves output parsing, output guardrails, observability, response augmentation, and memory commits;
- skips retrieval augmentation and input guardrails because there is no new user input.

The conversation must contain at least one message. The caller remains responsible for providing a message sequence accepted by the configured model/provider.

At build time, a resume method is rejected if it also defines a user message.

- Closes: #2677.

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -2076,6 +2076,7 @@ public class AiServicesProcessor {
         validateReturnType(method);
 
         boolean requiresModeration = method.hasAnnotation(LangChain4jDotNames.MODERATE);
+        boolean resumeConversation = method.hasAnnotation(LangChain4jDotNames.RESUME_CONVERSATION);
         java.lang.reflect.Type returnType = javaLangReturnType(method);
 
         List<MethodParameterInfo> params = method.parameters();
@@ -2096,6 +2097,11 @@ public class AiServicesProcessor {
         Optional<AiServiceMethodCreateInfo.TemplateInfo> systemMessageInfo = gatherSystemMessageInfo(method, templateParams);
         AiServiceMethodCreateInfo.UserMessageInfo userMessageInfo = gatherUserMessageInfo(method, templateParams,
                 systemMessageInfo, fallbackToDummyUserMessagePredicate, chatRequestParametersParamPosition);
+        if (resumeConversation
+                && (userMessageInfo.template().isPresent() || userMessageInfo.paramPosition().isPresent())) {
+            throw illegalConfigurationForMethod(
+                    "A method annotated with @ResumeConversation must not define a user message", method);
+        }
 
         AiServiceMethodCreateInfo.ResponseSchemaInfo responseSchemaInfo = ResponseSchemaInfo.of(generateResponseSchema,
                 systemMessageInfo,
@@ -2146,7 +2152,7 @@ public class AiServicesProcessor {
 
         return new AiServiceMethodCreateInfo(method.declaringClass().name().toString(), method.name(), parameterInfoList,
                 systemMessageInfo,
-                userMessageInfo, memoryIdParamPosition, requiresModeration, methodReturnTypeSignature,
+                userMessageInfo, memoryIdParamPosition, requiresModeration, resumeConversation, methodReturnTypeSignature,
                 overrideChatModelParamPosition, chatRequestParametersParamPosition,
                 metricsTimedInfo, metricsCountedInfo, spanInfo, responseSchemaInfo,
                 methodToolClassInfo, methodMcpClientNames, switchToWorkerThreadForToolExecution,
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/LangChain4jDotNames.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/LangChain4jDotNames.java
@@ -48,6 +48,7 @@ import io.quarkiverse.langchain4j.ModelName;
 import io.quarkiverse.langchain4j.OnThinking;
 import io.quarkiverse.langchain4j.PdfUrl;
 import io.quarkiverse.langchain4j.RegisterAiService;
+import io.quarkiverse.langchain4j.ResumeConversation;
 import io.quarkiverse.langchain4j.SeedMemory;
 import io.quarkiverse.langchain4j.VideoUrl;
 import io.quarkiverse.langchain4j.guardrails.ToolInputGuardrail;
@@ -96,6 +97,7 @@ public class LangChain4jDotNames {
 
     public static final DotName MODEL_NAME = DotName.createSimple(ModelName.class);
     public static final DotName REGISTER_AI_SERVICES = DotName.createSimple(RegisterAiService.class);
+    static final DotName RESUME_CONVERSATION = DotName.createSimple(ResumeConversation.class);
 
     static final DotName BEAN_CHAT_MODEL_SUPPLIER = DotName.createSimple(
             RegisterAiService.BeanChatLanguageModelSupplier.class);
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ResumeConversationTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ResumeConversationTest.java
@@ -66,7 +66,7 @@ class ResumeConversationTest {
                 ToolExecutionResultMessage.from(toolRequest, "approved"));
         TestChatMemoryStore.MESSAGES.put(memoryId, new ArrayList<>(suspendedConversation));
 
-        assertThat(assistant.resume(memoryId)).isEqualTo("Continuing after approval");
+        assertThat(assistant.resume(memoryId)).isEqualTo("Conversation continued");
 
         assertThat(TestChatModel.lastRequest.messages()).containsExactlyElementsOf(suspendedConversation);
         assertThat(TestChatModel.lastRequest.messages()).filteredOn(UserMessage.class::isInstance).hasSize(1);
@@ -75,17 +75,27 @@ class ResumeConversationTest {
                         suspendedConversation.get(0),
                         suspendedConversation.get(1),
                         suspendedConversation.get(2),
-                        AiMessage.from("Continuing after approval")));
+                        AiMessage.from("Conversation continued")));
     }
 
     @Test
     @ActivateRequestContext
-    void rejectsConversationThatIsNotWaitingForAToolResult() {
-        String memoryId = "active-conversation";
-        TestChatMemoryStore.MESSAGES.put(memoryId, new ArrayList<>(List.of(UserMessage.from("Hello"))));
+    void resumesFromGeneralConversationHistory() {
+        String memoryId = "general-conversation";
+        List<ChatMessage> conversation = List.of(
+                UserMessage.from("Summarize our discussion"),
+                AiMessage.from("We discussed deployment options."));
+        TestChatMemoryStore.MESSAGES.put(memoryId, new ArrayList<>(conversation));
 
-        assertThatThrownBy(() -> assistant.resume(memoryId))
-                .hasMessageContaining("last chat memory message is a ToolExecutionResultMessage");
+        assertThat(assistant.resume(memoryId)).isEqualTo("Conversation continued");
+        assertThat(TestChatModel.lastRequest.messages()).containsExactlyElementsOf(conversation);
+    }
+
+    @Test
+    @ActivateRequestContext
+    void rejectsEmptyConversation() {
+        assertThatThrownBy(() -> assistant.resume("empty-conversation"))
+                .hasMessageContaining("Cannot resume a conversation with empty chat memory");
     }
 
     @RegisterAiService(chatLanguageModelSupplier = TestChatModelSupplier.class, chatMemoryProviderSupplier = TestChatMemoryProviderSupplier.class)
@@ -143,7 +153,7 @@ class ResumeConversationTest {
         public ChatResponse doChat(ChatRequest chatRequest) {
             lastRequest = chatRequest;
             return ChatResponse.builder()
-                    .aiMessage(AiMessage.from("Continuing after approval"))
+                    .aiMessage(AiMessage.from("Conversation continued"))
                     .build();
         }
     }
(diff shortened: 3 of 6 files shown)


arnabnandy7:
@geoand please verify the changes as discussed.

geoand:
cc @mariofusco @sabob

mariofusco:
I need to give a deeper look to this, but it seems that this pull request has quite a few overlaps with this new feature that we just merged in vanilla langchain4j https://github.com/langchain4j/langchain4j/pull/5767 did you see it? Probably you can review what you've done here reusing part of what has been done in langchain4j?

arnabnandy7:
> I need to give a deeper look to this, but it seems that this pull request has quite a few overlaps with this new feature that we just merged in vanilla langchain4j https://github.com/langchain4j/langchain4j/pull/5767 did you see it? Probably you can review what you've done here reusing part of what has been done in langchain4j?

Sure I'll take a look, I've missed it. Thanks for sharing.

arnabnandy7:
@mariofusco I reviewed [langchain4j/langchain4j#5767](https://github.com/langchain4j/langchain4j/pull/5767).

It provides durable suspension/resume for agentic workflows through AgenticScope, whereas this PR targets declarative AI Services using persisted `ChatMemory` and `ToolExecutionResultMessages`. So there is conceptual overlap, but the upstream API does not appear to directly cover this non-agentic AI Service use case.

Would you prefer that we retain this smaller AI Service-specific API, or rework the solution around the upstream agentic suspension mechanism?

cc @geoand @sabob

sabob:
This PR looks like what I'm after. 
Didn't know about the new persistent agentic support added by @mariofusco. That looks great as well.
But we not using the agentic module, just plain langchain4j.
We also didn't intend to build a HITL functionality, it just sort of came up on the radar and we realized we already have most of the plumbing in place for other requirements we had.
The Resume functionality is simply a missing piece, and it does seem useful feature in general.
One note on the PR:
``` java
if (!(messages.get(messages.size() - 1) instanceof ToolExecutionResultMessage)) {
```
why the guard to have a ToolExecturionResultMessage? Technically correct I suppose, but I would just send the messages as provided and let the LLM sort it out :)

Kind regards

Bob

arnabnandy7:
> This PR looks like what I'm after. 

Thanks for confirming. Your original use case is exactly what this PR targets: continuing a persisted AI Service conversation without inserting a synthetic user message.

> But we not using the agentic module, just plain langchain4j. 

That is the key distinction. The new upstream persistence support is designed around agentic workflows and `AgenticScope`, while this PR addresses declarative AI Services backed by `ChatMemory`, without requiring the agentic module.

> We also didn't intend to build a HITL functionality, it just sort of came up on the radar and we realized we already have most of the plumbing in place for other requirements we had. 

Gotya. Although HITL motivated the issue, the underlying requirement is more general: an application may already have persisted messages and need the AI Service to continue processing them without creating a user prompt that never existed.

>The Resume functionality is simply a missing piece, and it does seem useful feature in general. One note on the PR:
> 
> ```java
> if (!(messages.get(messages.size() - 1) instanceof ToolExecutionResultMessage)) {
> ```

I agree with you on this, so I’ll keep the API framed as general conversation continuation rather than HITL-specific functionality. HITL is one use case, but it should not define or constrain the abstraction.

> why the guard to have a ToolExecturionResultMessage? Technically correct I suppose, but I would just send the messages as provided and let the LLM sort it out :)

The guard was initially intended to detect invalid HITL continuation state early, but it unnecessarily restricts the more general resume API. I’ll remove the message-type requirement and retain only the empty-memory validation. The caller will be responsible for supplying a conversation history accepted by the selected model/provider.

geoand (commented):
I like this!

What is missing is some examples in the docs and also making the Javadoc of the new annotation more extensive as currently it's hard to tell what use case it targets

arnabnandy7:
> I like this!
> 
> What is missing is some examples in the docs and also making the Javadoc of the new annotation more extensive as currently it's hard to tell what use case it targets

I'll do it shortly and notify you here.

mariofusco:
This is my very personal point of view and I would prefer to also know @dliubarskyi's opinion on this.

Correct me if I'm wrong, but the only situation in which this could be useful is to implement, in a quite convoluted way as described [here](https://github.com/quarkiverse/quarkus-langchain4j/issues/2677), a HITL feature at the level of the single AiService instead of configuring a whole agentic system.

If so, I totally recognize this need, but implementing it through a tool configured with `ReturnBehavior.IMMEDIATE`, that has been introduced for a completely different use case, seems to be more an ugly workaround for a missing feature rather than the way in which we want to support HITL in LangChain4j's AiServices.

If possible I'd take a step back and try to figure out if we want to provide HITL support in AiServices and how to properly design this feature. Finally, even if we realize that we cannot do much better to implement a HITL (and I'm skeptical of this), and want to bring in this `@ResumeConversation` annotation as it is, I don't see why we should keep it in the quarkus-langchain4j extension alone, instead of also offering it in upstream LangChain4j.

sabob:
> This is my very personal point of view and I would prefer to also know @dliubarskyi's opinion on this.
> 
> Correct me if I'm wrong, but the only situation in which this could be useful is to implement, in a quite convoluted way as described [here](https://github.com/quarkiverse/quarkus-langchain4j/issues/2677), a HITL feature at the level of the single AiService instead of configuring a whole agentic system.

Yes to convoluted and yes, it's a finer granularity than the agentic module. Basically, at the tool level.

We have HITL tools eg "provide me with your location" or "you need to be authenticated before I can run this tool, so go login, then come back".

In another thread it was recommended to use Langraph4j instead, but that is a bit extreme.

> If so, I totally recognize this need, but implementing it through a tool configured with `ReturnBehavior.IMMEDIATE`, that has been introduced for a completely different use case, seems to be more an ugly workaround for a missing feature rather than the way in which we want to support HITL in LangChain4j's AiServices.

Yep, dont need to be IMMEDIATE, could be any tool, as long as it ends the agent loop.

> 
> If possible I'd take a step back and try to figure out if we want to provide HITL support in AiServices and how to properly design this feature. Finally, even if we realize that we cannot do much better to implement a HITL (and I'm skeptical of this), and want to bring in this `@ResumeConversation` annotation as it is, I don't see why we should keep it in the quarkus-langchain4j extension alone, instead of also offering it in upstream LangChain4j.

Anything that improves HITL is great, its trickier than I imagined :)

Another use case for this @ResumeConversation would be crash recovery. In this case it's more of a "Retry" mechanism though. 
Or asynchronous use case, where an external system wants to trigger an agent loop without adding a user msg.

mariofusco:
> > If so, I totally recognize this need, but implementing it through a tool configured with `ReturnBehavior.IMMEDIATE`, that has been introduced for a completely different use case, seems to be more an ugly workaround for a missing feature rather than the way in which we want to support HITL in LangChain4j's AiServices.
> 
> Yep, dont need to be IMMEDIATE, could be any tool, as long as it ends the agent loop.

To my knowledge the only tools that can interrupt the agent loop are the `IMMEDIATE` ones. The only other alternative that I see is using a tool guardrail, but I guess that it would be even more complicated, and for no good reason.

sabob:
> > Yep, dont need to be IMMEDIATE, could be any tool, as long as it ends the agent loop.
> 
> To my knowledge the only tools that can interrupt the agent loop are the `IMMEDIATE` ones. The only other alternative that I see is using a tool guardrail, but I guess that it would be even more complicated, and for no good reason.

We also use the SystemPrompt to stop the loop, although, the agent can ignore it and just keep going ;-)

sabob:
Curious though, is a "persistent" HILT implementation on the cards for langchian4j?  Similar to the work @mariofusco did for the agentic module?

arnabnandy7:
Thanks @mariofusco, I agree that it is worth taking a step back and separating two concerns:

1. a generic way to continue an existing AI Service invocation from persisted messages; and
2. a complete HITL suspension/checkpoint protocol.

The first looks like a useful upstream AI Service primitive, while the second needs a broader design and should not be defined around `ReturnBehavior.IMMEDIATE`.

As a starting point, I would like to propose the following upstream API (names are open for discussion):

```java
interface Assistant {

    String chat(@MemoryId String conversationId,
                @UserMessage String message);

    @ContinueConversation
    Result<String> continueConversation(@MemoryId String conversationId);
}
```

Proposed contract for `@ContinueConversation`:

- require configured, non-empty `ChatMemory`;
- load and submit the existing message sequence without creating or appending a new `UserMessage`;
- accept any model/provider-compatible history rather than requiring a trailing `ToolExecutionResultMessage`;
- skip stages that specifically require fresh user input, such as input guardrails and retrieval augmentation;
- retain the normal AI Service model call, tool loop, output parsing/guardrails, observability and memory commits;
- allow the usual invocation-control parameters, such as `@MemoryId`, `InvocationParameters` and `ChatRequestParameters`, but reject parameters that define new user content.

Internally, this could be implemented as a reusable continuation path in the upstream AI Service execution code, allowing Quarkus and other integrations to use the same semantics instead of implementing their own orchestration.

I would deliberately keep suspension IDs, checkpoints, pending human responses, duplicate-resume protection and workflow recovery outside this annotation. Those belong to a proper AI Service HITL design, which could later build on the continuation primitive.

This also covers non-HITL cases raised by @sabob, such as an external asynchronous event continuing an existing conversation. We should be more careful about describing it as crash recovery, since retries can repeat model calls or tool side effects unless the application provides idempotency and concurrency control.

If this direction sounds reasonable, I am happy to open an issue in `langchain4j/langchain4j` to refine the contract and contribute the upstream implementation. We can pause or rework this Quarkus PR to consume the upstream API rather than introducing a Quarkus-only annotation.

Would you prefer an explicit method annotation like the above, or a programmatic continuation API on `AiServices`?

dliubarskyi:
> I would just send the messages as provided and let the LLM sort it out :)

LLMs are very picky about message order, so I would be careful here

> an application may already have persisted messages and need the AI Service to continue processing them without creating a user prompt that never existed.

Could you please provide a few realistic examples of this?

> I don't see why we should keep it in the quarkus-langchain4j extension alone, instead of also offering it in upstream LangChain4j.

Agreed

---

To proceed with this discussion, if would be very helpful to have a few examples of when resuming AI Service is useful.
For each example, please specify how AI Service was suspended, what is the state of ChatMemory (list of messages with their types and content)  after suspension and right before AI Service is resumed.

Thanks!

sabob:
> To proceed with this discussion, if would be very helpful to have a few examples of when resuming AI Service is useful. For each example, please specify how AI Service was suspended, what is the state of ChatMemory (list of messages with their types and content) after suspension and right before AI Service is resumed.
> 
> Thanks!


We persist every ChatMessage, UserMessage, AI, ToolExecutionRequestMessage, ToolExecutionResultMessage to our own database for audit purposes. Since we have the complete message history, we can also rebuild ChatMemory from those persisted messages whenever we need to resume a conversation.

Our primary use case is HITL

So, a tool interrupts the conversation (e.g. approval, login, or location). 
Once the human responds, we persist the user data as a ToolExecutionResultMessage. We contemplates storing it as a UserMessage as well.

ChatMemory could contain:

UserMessage: Book me a flight to London.
AI: Call requestApproval(), id: 123
ToolExecutionResultMessage: Approved, id: 123

At this point, we want the AI Service to continue processing the existing conversation.

Crash recovery
We haven't run into a crash yet, but I'd use the same resume functionality. (Technically, its retry ;-) )

After a restart, we would rebuild ChatMemory from the messages that were successfully persisted and continue from there. Before rebuilding the memory, we remove any unmatched ToolExecutionRequestMessages (requests without a corresponding ToolExecutionResultMessage), ensuring the model only receives a valid message sequence.

In both cases, the application already has the complete conversation history. 
We just need kick off another agent run and we will provide the messages from our Database MEmoryStore.

Kind regards

Bob

sabob:
Another real-world use case is when an application wants to trigger an agent run based on new messages added to its own persistence by external clients. The goal isn't to continue a user conversation, but to let the agent react.
For example by generating a summary and sending it to another API.

Example:

```text
UserMessage:
Book me a flight to London.

AiMessage:
Call requestApproval(), id: 123

ToolExecutionResultMessage:
Approved, id: 123
```

An external API then inserts a new message into the application's database:

```text
UserMessage:
<event type="ui-click">
User clicked the "Accept" button.
</event>
```

The application then resumes the conversation from the persisted history, allowing the agent to generate a summary of the newly added messages and invoke another API with that summary. This can be useful for integrating UI interactions or external system events into an agentic workflow without requiring a synthetic "continue" message.

The general pattern for a resume requirement becomes applicable whenever an application persist messages because the conversation can evolve independently of a new user prompt.

Kind regards,
Bob

sbtznb:
This is great and exactly what we are also looking for. At the moment AIService is very limited in only being able to execute on new user message input.

The general idea is agent could be resumed from other signals besides user input. HITL and asynchronous long running tool calls are good examples.

LC4J would hand back control to our runtime using IMMEDIATE return behaviour. We would essentially keep agent run hanging with unanswered tool call. Then we would _eventually_ persist tool call result and use "resume".

PR #2682, opened 2026-07-20
Bump to LangChain4j 1.18.0

mariofusco:


model-providers/jaxrs-client/src/main/java/io/quarkiverse/langchain4j/jaxrsclient/JaxRsHttpClient.java
```diff
@@ -170,7 +170,6 @@ private void executeBlocking(HttpRequest request, ServerSentEventParser parser,
                 listener.onOpen(SuccessfulHttpResponse.builder()
                         .statusCode(response.getStatus())
                         .headers(response.getStringHeaders())
-                        .body(null)
```
geoand, line 173: Why this change?

mariofusco: With the latest release it has been added a second `body()` method taking as argument a `String` instead of a `byte[]` so this statement didn't compile anymore because the call `body(null)` was ambiguous. Rather than replacing it with `body((byte[])null)`, that anyway is useless and has no effect since `null` is the default, I decided to simply remove the statement.

mariofusco:
I also added support for the new `@BeforeCall` and `@A2AServerUrlSupplier` annotations, now it should be ready to be reviewed.

model-providers/jaxrs-client/src/main/java/io/quarkiverse/langchain4j/jaxrsclient/JaxRsHttpClient.java
mariofusco: Mmm, maybe thinking twice the purpose of that statement was indeed to nullify the body regardless of the original content? If so I will add it back, please let me know.

geoand: Yeah, I prefer we do

mariofusco: Done.

geoand (approved):

mariofusco:
The build now fails with two tests timed out after 10s while calling the public MCP registry, which I believe is something totally unrelated with this version upgrade.

PR #2688, opened 2026-07-22
Allow to use @Skills annotation also on AiServices methods

mariofusco:
This is a follow up of the work done in this pull request https://github.com/quarkiverse/quarkus-langchain4j/pull/2654

It allows to use `@Skills` annottaion also on `AiServices` methods and also demonstrates how users can provide their own `SkillsConfigurator` implementation without CDI ambiguity.

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
@@ -114,6 +114,7 @@ import io.quarkiverse.langchain4j.runtime.QuarkusServiceOutputParser;
 import io.quarkiverse.langchain4j.runtime.ResponseSchemaUtil;
 import io.quarkiverse.langchain4j.runtime.ToolCallsLimitExceededException;
 import io.quarkiverse.langchain4j.runtime.aiservice.GuardrailsSupport.OutputGuardrailStreamingMapper;
+import io.quarkiverse.langchain4j.runtime.skills.SkillsConfigurator;
 import io.quarkiverse.langchain4j.runtime.tool.QuarkusToolExecutor;
 import io.quarkiverse.langchain4j.runtime.types.TypeSignatureParser;
 import io.quarkiverse.langchain4j.runtime.types.TypeUtil;
@@ -287,6 +288,25 @@ public class AiServiceMethodImplementationSupport {
             }
         }
 
+        // @Skills annotation support — resolve per-method or class-level skills
+        List<String> effectiveSkillNames = methodCreateInfo.getSkillNames();
+        if (effectiveSkillNames == null && context instanceof QuarkusAiServiceContext) {
+            effectiveSkillNames = ((QuarkusAiServiceContext) context).classLevelSkillNames;
+        }
+        String skillsSystemMessage = null;
+        if (effectiveSkillNames != null) {
+            SkillsConfigurator configurator = Arc.container().select(SkillsConfigurator.class).get();
+            ToolProviderRequest skillsRequest = new QuarkusToolProviderRequest(invocationContext, userMessage,
+                    methodCreateInfo.getMcpClientNames());
+            ToolProviderResult skillsResult = configurator.createToolProvider(effectiveSkillNames)
+                    .provideTools(skillsRequest);
+            for (ToolSpecification spec : skillsResult.tools().keySet()) {
+                toolSpecifications.add(spec);
+                toolExecutors.put(spec.name(), skillsResult.tools().get(spec));
+            }
+            skillsSystemMessage = configurator.buildSkillsSystemMessage(effectiveSkillNames);
+        }
+
         AugmentationResult augmentationResult = null;
         if (context.retrievalAugmentor != null) {
             Metadata metadata = Metadata.builder()
@@ -322,6 +342,21 @@ public class AiServiceMethodImplementationSupport {
         } else {
             messagesToSend = createMessagesToSendForNoMemory(systemMessage, userMessage, needsMemorySeed, context,
                     methodCreateInfo);
+        }
+
+        if (skillsSystemMessage != null) {
+            boolean appended = false;
+            for (int i = 0; i < messagesToSend.size(); i++) {
+                if (messagesToSend.get(i) instanceof dev.langchain4j.data.message.SystemMessage existing) {
+                    messagesToSend.set(i,
+                            dev.langchain4j.data.message.SystemMessage.from(existing.text() + "\n\n" + skillsSystemMessage));
+                    appended = true;
+                    break;
+                }
+            }
+            if (!appended) {
+                messagesToSend.add(0, dev.langchain4j.data.message.SystemMessage.from(skillsSystemMessage));
+            }
         }
 
         if (TypeUtil.isTokenStream(returnType)) {
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -614,6 +614,8 @@ public class AiServicesProcessor {
                     skillNames = List.of();
                 }
             }
+            boolean hasMethodLevelSkills = declarativeAiServiceClassInfo.methods().stream()
+                    .anyMatch(m -> m.hasAnnotation(LangChain4jDotNames.SKILLS));
 
             declarativeAiServiceProducer.produce(new DeclarativeAiServiceBuildItem(
                     declarativeAiServiceClassInfo,
@@ -647,7 +649,8 @@ public class AiServicesProcessor {
                     impliedRegisterAiServiceTarget.contains(declarativeAiServiceClassInfo.name()),
                     shouldThrowExceptionOnEventError,
                     chatMemoryFlushStrategySupplierClassDotName,
-                    skillNames));
+                    skillNames,
+                    hasMethodLevelSkills));
 
         }
         toolProviderProducer.produce(new ToolProviderMetaBuildItem(toolProviderInfos));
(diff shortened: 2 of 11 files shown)


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -324,6 +344,21 @@ private static Object doImplement0(AiServiceMethodCreateInfo methodCreateInfo, I
                     methodCreateInfo);
         }
 
+        if (skillsSystemMessage != null) {
+            boolean appended = false;
+            for (int i = 0; i < messagesToSend.size(); i++) {
+                if (messagesToSend.get(i) instanceof dev.langchain4j.data.message.SystemMessage existing) {
+                    messagesToSend.set(i,
+                            dev.langchain4j.data.message.SystemMessage.from(existing.text() + "\n\n" + skillsSystemMessage));
```
jmartisk, line 352: As discussed previously, I'm very concerned about retroactively changing the system message.
Furthermore, if this system message already contains the skills from a previous call in the same conversation, then the declaration will be duplicated, no?

jmartisk: For sure we should have some test cases that invoke two different AI service methods (with different sets of skills) in a single conversation

jmartisk: Well I asked AI about the possible consequences of this...

> 
> So when method 2 runs with a different skill set:
> 
>   - The system message might list different available skills, but the full content of previously activated skills is already sitting in the conversation as prior tool results
>   - The LLM doesn't need the skill to be "available" anymore — it already has the entire markdown content in context
> 

Generally, once a skill is activated, you can't deactivate it unless you explicitly remove the related parts of history.

mariofusco: I added the new `SkillsChatMemoryTest` demonstrating this.

mariofusco: > Well I asked AI about the possible consequences of this...
> 
> > So when method 2 runs with a different skill set:
> > 
> > * The system message might list different available skills, but the full content of previously activated skills is already sitting in the conversation as prior tool results
> > * The LLM doesn't need the skill to be "available" anymore — it already has the entire markdown content in context
> 
> Generally, once a skill is activated, you can't deactivate it unless you explicitly remove the related parts of history.

I have seen, but I'm not sure what to do about it. Maybe we could simply document this behavior? Any better suggestion?

jmartisk: This general concept of retroactively changing the system message may also be problematic from a traceability/auditability PoV, not just because it can confuse the LLM. I would say a conversation should be append-only, like a blockchain, and what you add to it should become immutable, otherwise it becomes hard to audit and reason about it. For example, if some sort of auditing system submits the conversation parts somewhere over time, and then the application changes something in it after it was submitted, then you have an inconsistency that's hard to reconcile if you don't keep some sort of transactional records to trace the changes. But this is more like a general concern about industry practices then a concrete technical problem. It's hard to say if it indeed is a problem.

But..... let's say that I accept that we're going to do this, I see a problem in this implementation here - you don't actually update the chat memory with the (new or updated) system message - just changing it in the `messagesToSend` list doesn't submit it to the `ChatMemory`. You would need to also call `committableChatMemory.set(messagesToSend)` or something like that after  this `for` loop, to actually update the memory.
This is why your test is passing and isn't seeing the duplicated declarations on the second method call. But it basically means that the chat memory will not contain the updated system message at all and will be different from what is actually sent to the LLM, which is again a problem.

However, if we call `committableChatMemory.set(...)` here, we will break what is written in the javadoc of that method:

> NOTE: This method is never called automatically by LangChain4j.

jmartisk: I don't know, sorry I rather only present problems and no solutions.. :) I think the only proper solution here would be to move the declaration from the system message to the user message, in a RAG-augmenting kind of way.

mariofusco: I see your point, and I don't have an answer as well. What I'm trying to say is:

1. I'm not sure if this is more a theoretical rather than a practical problem: if I understand correctly this should happen when 2 or more methods of an `AiService` share the same `ChatMemory`, but we don't have any test case or example covering this scenario at all, so I guess it is not common and I believe it should also be an anti-pattern.
2. In the scenario at the former point, the issues related with the need to change a system message, hasn't been introduced by skills or the new `@Skills` annotation, but it eventually existed exactly in the same way even before. 

To clarify 2. suppose that you have something like:

```
@RegisterAiService
public interface MyService {
    @SystemMessage("Message #1")
    String method1(@MemoryId String memoryId, @UserMessage String msg);

    @SystemMessage("Message #2")
    String method2(@MemoryId String memoryId, @UserMessage String msg);
}
```

wouldn't you have the same problem of changing system message when calling a method and then the other one? Do we deal with this situation at the moment and if so how?

To recap I admit that we have a (potential) problem here, but it is more general than the one introduced by skills and ultimately orthogonal to it, so my suggestion is to merge this pull request as it is (or let me know if there is something else that I could improve) and investigate the issue of the changing system messages in a separated work.

jmartisk: Your point 2 is a very good point, I didn't think of that. So we do have this problem already (in slightly different variations). I would say it's a bit of a design issue, but if nobody has complained yet, maybe I'm overestimating the impact. Or maybe you're right that nobody uses multiple AI service methods within one conversation.

Ok, let's go with this in that case, but I guess we should at least find out how to properly update the system message in the memory, otherwise the memory will be inconsistent with what is actually sent to the LLM.

And maybe a test where the affected methods have `@Skills` as well as a `@SystemMessage` on them, to see how these things will be merged together.

mariofusco: I added the test that you suggested.

For what regards updating the system message in the chat memory, thinking twice I'm not sure that this is the right thing to do. Skills are per-method, not per-conversation. Persisting skills text to shared conversation memory would cause incorrect behavior when different methods with different skills share the same memory. Specifically, if method A stores its skills text in memory and method B (without `@SystemMessage`) inherits it, method B's skills injection would append to the old skills text, causing accumulation. 

In essence I'm not sure of what is the right thing to do here (and I would appreciate @geoand and @dliubarskyi feedback on this), but I believe that the current transient-only injection approach is not only the simplest thing to do but also the most correct one.

jmartisk: IMO the most correct approach would be to move it to the user message (in a RAG augmentation style) instead of system message. Then you could freely adjust it for each method separately. But yes, that would be a rather intrusive change.

Ok, I'm fine with this as it is, even though I have concerns about the correctness, I guess it's good enough for the time being.

geoand: I have to say I've lost the plot completely here...

mariofusco: Moving skills from the system to the user message (I honestly don't know if there is a standard on where they should be placed) would have its pros and cons. In a long conversation you will end up with a ChatMemory with multiple contradictory user messages about the available skills and I'm afraid this would confuse the LLM even more.

Regardless if it is a good idea or not, this change is much more invasive, will also heavily involve the upstream langchain4j implementation and is definitively out of scope for this pull request. @jmartisk If you agree please approve this PR.

jmartisk (approved):

core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
mariofusco: > I have to say I've lost the plot completely here...

@geoand Please let me know if you need any clarification.

geoand: I trust that you and @jmartisk have worked it out :)

jmartisk: Use AI to summarize it :) The gist is that I'm concerned about retroactively changing the contents of the system message and then continuing the conversation while showing the LLM a history that isn't the same as it actually happened. But as it turns out, we already have this kind of problem in a different form anyway.

dliubarskyi: > For what regards updating the system message in the chat memory, thinking twice I'm not sure that this is the right thing to do. Skills are per-method, not per-conversation. Persisting skills text to shared conversation memory would cause incorrect behavior when different methods with different skills share the same memory. Specifically, if method A stores its skills text in memory and method B (without @SystemMessage) inherits it, method B's skills injection would append to the old skills text, causing accumulation.

Isn't it an implementation problem? Can't we make sure method B does not reuse the system message from other method?

> In essence I'm not sure of what is the right thing to do here (and I would appreciate @geoand and @dliubarskyi feedback on this), but I believe that the current transient-only injection approach is not only the simplest thing to do but also the most correct one.

I am not sure either, but skills are not transient in their nature, they are activated and live in scope of a specific conversation/memory

mariofusco:
Can we merge this?

jmartisk:
:)

PR #2442, opened 2026-05-09
Support named OIDC clients in OidcClientMcpAuthProvider

omatheusmesmo:
## Description

Adds `quarkus.langchain4j.mcp.<client>.oidc-client-name` to bind each MCP client to a distinct named OIDC client configuration. Previously, `OidcClientMcpAuthProvider` only used the default OIDC client, making it impossible to connect to multiple MCP servers with unrelated OIDC issuers in the same application.

The implementation uses `SyntheticBeanBuildItem` instead of `AdditionalBeanBuildItem` so that a specific `OidcClient` can be selected at runtime per MCP client name. A default bean (no qualifier) is still registered for backward compatibility. 

### Example configuration:
  
  ```properties
  quarkus.langchain4j.mcp.client1.transport-type=http
  quarkus.langchain4j.mcp.client1.url=https://server1.example.com/mcp/sse
  quarkus.langchain4j.mcp.client1.oidc-client-name=provider1

  quarkus.langchain4j.mcp.client2.transport-type=http
  quarkus.langchain4j.mcp.client2.url=https://server2.example.com/mcp/sse
  quarkus.langchain4j.mcp.client2.oidc-client-name=provider2

  quarkus.oidc-client.provider1.auth-server-url=https://sso1.example.com
  quarkus.oidc-client.provider1.client-id=app-client-1
  quarkus.oidc-client.provider1.credentials.secret=...
  quarkus.oidc-client.provider1.grant.type=client_credentials

  quarkus.oidc-client.provider2.auth-server-url=https://sso2.example.com
  quarkus.oidc-client.provider2.client-id=app-client-2
  quarkus.oidc-client.provider2.credentials.secret=...
  quarkus.oidc-client.provider2.grant.type=client_credentials
```
---

- Fixes: #2090

PR #2689, opened 2026-07-22
Deprecate oidc-client-name in MCP client config, introduce extra-params

sberyozkin:
#2442 that added support for named OIDC clients to MCP Client authentication provider was good, but I missed that it introduced a rather visible reference to OIDC client in the MCP Client configuration. IMHO it should be avoided if possible.

This PR proposes a looser option for associating extension specific properties with a given MCP client configuration via an `extra-params` map property instead.


sberyozkin:
Sorry, a bit more thinking is needed

sberyozkin:
It occurred to me it should not really be a build time property either, I'll try to get a similar pattern done that is used with Azure OpenAI for checking ModelAuthProvider

sberyozkin:
Took me a while, butthese extra-params are now runtime, also optimized the recorder to have a single function - where `OidcClients.getClient()` returns a default client if no oidc-client-name is configured and a named client otherwise.

PR #2687, opened 2026-07-21
Replace langchain4j-qdrant gRPC client with quarkus-qdrant REST client

zbendhiba:
- Closes: #451

PR #2691, opened 2026-07-22
feat: Add create-collections flag to Qdrant Dev Services

edeandrea:
## Summary

- Adds a `quarkus.langchain4j.qdrant.devservices.create-collections` config property (default: `false`) that automatically creates Qdrant collections for all configured named stores during Dev Services startup
- When enabled, uses the same vector-params (distance, size) configured for the default collection to create each named store's collection
- Propagates the resolved collection name into each named store's runtime config, so users don't need to separately configure both `collection-name` (build-time) and `collection.name` (runtime)

### Design note: `Map<String, String>` vs `Set<String>` in `QdrantDevServiceCfg`

The previous `Set<String>` of named store keys (passed as a method parameter) has been replaced with a `Map<String, String>` of store name → resolved collection name stored in the `QdrantDevServiceCfg` record. This is intentional: the record's `equals()` is used for hot-reload change detection, and a `Set<String>` of keys would not detect collection-name renames — leaving stale collections in the running container. A reviewer may reasonably prefer the simpler `Set<String>` approach if hot-reload detection of collection-name renames is considered unnecessary.

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-all-config.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-all-config.adoc
@@ -25371,6 +25371,27 @@ endif::add-copy-button-to-env-var[]
 |string
 |`+++qdrant+++`
 
+a|icon:lock[title=Fixed at build time] [[quarkus-langchain4j-qdrant_quarkus-langchain4j-qdrant-devservices-create-collections]] [.property-path]##link:#quarkus-langchain4j-qdrant_quarkus-langchain4j-qdrant-devservices-create-collections[`+++quarkus.langchain4j.qdrant.devservices.create-collections+++`]##
+ifdef::add-copy-button-to-config-props[]
+config_property_copy_button:+++quarkus.langchain4j.qdrant.devservices.create-collections+++[]
+endif::add-copy-button-to-config-props[]
+
+
+[.description]
+--
+Whether to automatically create Qdrant collections for all configured named stores when Dev Services starts. Requires the collection vector-params to be configured.
+
+
+ifdef::add-copy-button-to-env-var[]
+Environment variable: env_var_with_copy_button:+++QUARKUS_LANGCHAIN4J_QDRANT_DEVSERVICES_CREATE_COLLECTIONS+++[]
+endif::add-copy-button-to-env-var[]
+ifndef::add-copy-button-to-env-var[]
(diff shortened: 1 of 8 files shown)


geoand:
Thanks!

> This is intentional: the record's equals() is used for hot-reload change detection, and a Set<String> of keys would not detect collection-name renames — leaving stale collections in the running container. A reviewer may reasonably prefer the simpler Set<String> approach if hot-reload detection of collection-name renames is considered unnecessary.

I don't understand this at all

edeandrea:
> Thanks!
> 
> > This is intentional: the record's equals() is used for hot-reload change detection, and a Set of keys would not detect collection-name renames — leaving stale collections in the running container. A reviewer may reasonably prefer the simpler Set approach if hot-reload detection of collection-name renames is considered unnecessary.
> 
> I don't understand this at all

It means that if you just use the initial Set of collection names, if the user changes a collection name in the config, dev services will not remove the previous collection that was there before.

Say in your config you define "collection-a". Dev services runs and with this PR will auto-create "collection-a". If you change your config and change "collection-a" to "collection_a", dev services will kick in and create "collection_a", but doesn't know enough to remove "collection-a" because it doesn't exist anymore in the config. The Map approach allows the collection names to "persist" per-se across dev mode restarts. Once the collection is destroyed it is removed

geoand:
Please squash the commits

edeandrea:
> Please squash the commits

Done

edeandrea:
No idea about these tests - all i did was pull in changes from `main`

geoand:
Don't worry about the test failures - those are due to the external MCP registry having spotty availability

edeandrea:
@geoand am I ok to merge this?

geoand:
I would like @jmartisk to approve

edeandrea:
@jmartisk ?

jmartisk:
Sorry I'd somehow missed this in the influx of notifications :(
LGTM, but needs rebase now...

edeandrea:
Closing this PR — the upstream refactor (PR #2687) replaced the custom Qdrant dev services with the `quarkus-qdrant` extension, which already supports collection creation via `quarkus.qdrant.devservices.collections.<name>.vector-size` and `.distance`. Since the vector dimension depends on the embedding model and can't be auto-detected, the remaining benefit of auto-bridging collection names is marginal. Users should configure collections directly through `quarkus.qdrant.devservices.collections.*`.

PR #2722, opened 2026-07-31
Fixes #2721: Add ExcludeFromImpliedAiServiceBuildItem

### LINKED ISSUE #2721: Add ExcludeFromImpliedAiServiceBuildItem to allow excluding interfaces from implied AI service registration
When QL4J is on the classpath with a default `ChatModel` configured, the implied AI service registration discovers any interface with `@UserMessage`/`@SystemMessage`/`@Moderate` and auto-registers it. This causes problems for other extensions whose library interfaces use these annotations but should not be managed as AI services.

Add a build item that allows other extensions to mark specific interfaces as excluded from implied registration.

This follows the pattern of `IgnoreJaxbAnnotatedClassesBuildItem` in `quarkus-jaxb`, which solves the same problem for JAXB auto-discovery.

Reference: https://github.com/apache/camel-quarkus/issues/8836

JiriOndrusek:
Add a build item that allows other extensions to exclude specific interfaces from implied AI service registration.

Fixes: #2721
Reference: https://github.com/apache/camel-quarkus/issues/8836

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -406,6 +406,7 @@ public class AiServicesProcessor {
     public void findDeclarativeServices(CombinedIndexBuildItem indexBuildItem,
             CustomScopeAnnotationsBuildItem customScopes,
             List<AnnotationsImpliesAiServiceBuildItem> annotationsImpliesAiServiceItems,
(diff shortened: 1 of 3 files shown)


geoand (approved):
Looks good, thanks!

geoand:
Actually, the docs change needs to be reverted (and will cause CI to fail anyway)

JiriOndrusek:
> Actually, the docs change needs to be reverted (and will cause CI to fail anyway)

It was generated for me during the build, so I attached it. Now it is removed. Thanks!

geoand (approved):

PR #2723, opened 2026-07-31
LangChain4j 1.19.0 + MCP client according to 2026-07-28 + watsonx changes

jmartisk:
Quarkus side of https://github.com/langchain4j/langchain4j/pull/5881

Note: tests with `Legacy` in the name means that it uses the old version of the protocol (2025-11-25).

Obviously, this will need langchain4j 1.19 released before it can be merged.


mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/config/McpClientRuntimeConfig.java
```diff
@@ -158,9 +158,38 @@ public interface McpClientRuntimeConfig {
...
+     * Only used with the legacy MCP protocol (versions up to 2025-11-25).
      * Only applies to MCP clients using the streamable HTTP transport.
      */
     @WithDefault("false")
     boolean subsidiaryChannel();
 
+    /**
+     * The MCP protocol version to use. If not set, the client will auto-detect
+     * the server's protocol version, preferring 2026-07-28 over 2025-11-25.
+     * Set to "2026-07-28" to force modern protocol, or "2025-11-25" to force legacy.
```
geoand: Can we add links to the versions? Also, maybe note what the basic differences are between modern and legacy?

geoand:
As you mentioned to me, indeed this change is not as big is much smaller than the 91 files changed would indicate

mcp/runtime/src/main/java/io/quarkiverse/langchain4j/mcp/runtime/config/McpClientRuntimeConfig.java
jmartisk: Well they have a bit of a history of changing the specification links, but ok I've added links to the old and new one. Also a note that the old one is stateful and the new one stateless.

jmartisk:
Can somebody rebuild the docs and push please? I still need to figure out why on my machine, it produces something vastly different from that the CI expects.

andreadimaio:
> Can somebody rebuild the docs and push please? I still need to figure out why on my machine, it produces something vastly different from that the CI expects.

Done.

jmartisk:
Thanks, it's still failing though :( This is super weird. I wonder if that's some sort of bug in the docs generator.. but given that it always produces the same result on a single machine, it's hard to diagnose.

jmartisk:
@andreadimaio the docs build passed this time, did you do something differently? :)

andreadimaio:
> @andreadimaio the docs build passed this time, did you do something differently? :)

I did a `mvn clean package` (skipping tests) instead of `mvn compile` :/

jmartisk:
Ok, thanks, I'll need to check that :D anyway let's get this merged now. Thanks a lot

jmartisk:
@andreadimaio for reference, I have finally figured out the problem with documentation. See https://github.com/quarkusio/quarkus/issues/56104 and https://github.com/quarkusio/quarkus/pull/56105
Until we depend on a Quarkus version that fixes this, we must avoid building the docs with git worktrees nested inside the project directory.

PR #2725, opened 2026-07-31
Replace NoopChatMemory with LocalCommittableChatMemory

### LINKED ISSUE #2724: Replace NoopChatMemory with a list-backed CommittableChatMemory for the no-memory case
## Problem

`NoopChatMemory` silently breaks tool execution loops and output guardrail retries when no `ChatMemoryProvider` is configured.

This was not detected by existing QL4J tests because **all QL4J test AI services have a `ChatMemoryProvider` configured** — it is the default setup via `ChatMemoryProcessor.setupBeans()`. The `NoopChatMemory` path (no memory) is never exercised with tools or guardrails in the test suite.

The issue becomes visible with the introduction of `ExcludeFromImpliedAiServiceBuildItem` (used by Apache Camel Quarkus in [apache/camel-quarkus#8836](https://github.com/apache/camel-quarkus/issues/8836)). When an interface is excluded from implied AI service registration, it no longer gets the default `ChatMemoryProvider` injected — so `determineCommittableChatMemory()` returns `NoopChatMemory`, and tools/guardrails break.

### Why NoopChatMemory is not sufficient

`NoopChatMemory` has three methods that cause failures:

1. **`add()` is a no-op** — messages added during the tool execution loop (AI messages with tool requests, tool execution results) are silently discarded. The conversation context is lost.

2. **`messages()` returns `Collections.emptyList()`** — when the tool loop builds the retry LLM request via `committableChatMemory.messages()`, it gets an empty list. The LLM receives no context (no system message, no user message, no tool results), causing `IllegalArgumentException: messages cannot be null or empty`.

3. **`Collections.emptyList()` is unmodifiable** — `OutputGuardrailExecutor.execute()` calls `chatMemory.messages()` and then `add()` on the returned list to append the reprompt message. This throws `UnsupportedOperationException`.

### Why `DefaultCommittableChatMemory` is not suitable

`DefaultCommittableChatMemory` requires a `ChatMemory` delegate — its constructor seeds from `delegate.messages()` and its `commit()` persists back via `delegate.set(newMessages)`. In the no-memory case there is no backing `ChatMemory` to delegate to. Using a dummy delegate would conflate "deferred persistence" with "no persistence at all."

## Proposed fix

Replace `NoopChatMemory` with a `LocalCommittableChatMemory` in `determineCommittableChatMemory()`. This class:

- Tracks messages in a local `ArrayList` (like `DefaultCommittableChatMemory.newMessages`)
- Returns a defensive copy from `messages()` (like `DefaultCommittableChatMemory`)
- Handles `SystemMessage` ordering and `replaceLastAiMessage()` (same logic as `DefaultCommittableChatMemory`)
- Has a no-op `commit()` — nothing to persist
- Gets garbage-collected after the invocation — no cross-call state

The change is a single-line swap in `determineCommittableChatMemory()` plus the new class. All existing behavior for the memory case is unchanged.

In `doImplement0()`, the no-memory branch that creates `messagesToSend` via `createMessagesToSendForNoMemory()` also needs to populate the `LocalCommittableChatMemory` with those initial messages, so that `committableChatMemory.messages()` returns the full conversation during tool loops and guardrail retries.

## Related

- Previously reported as #2344 / PR #2345 (closed — the root cause is now confirmed)
- Apache Camel Quarkus issue: [apache/camel-quarkus#8836](https://github.com/apache/camel-quarkus/issues/8836)

JiriOndrusek:
## Summary
- Replace `NoopChatMemory` with `LocalCommittableChatMemory` in `determineCommittableChatMemory()` so that tool-loop retries and guardrail reprompts work when no `ChatMemoryProvider` is configured
- `LocalCommittableChatMemory` tracks messages within a single invocation without persistence
- Seed the local memory with initial messages in the no-memory branch so the tool-execution loop can read back conversation history
- Add `ToolWithoutMemoryProviderTest` exercising the programmatic `AiServices.builder()` + `toolProvider()` path without a `ChatMemoryProvider`

---

- Fixes: https://github.com/quarkiverse/quarkus-langchain4j/issues/2724

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
+++ b/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
@@ -342,6 +342,10 @@ public class AiServiceMethodImplementationSupport {
         } else {
             messagesToSend = createMessagesToSendForNoMemory(systemMessage, userMessage, needsMemorySeed, context,
                     methodCreateInfo);
+            // Seed the local memory so the tool-execution loop can read back the conversation history
+            for (ChatMessage msg : messagesToSend) {
+                committableChatMemory.add(msg);
+            }
         }
 
         if (skillsSystemMessage != null) {
@@ -1385,7 +1389,7 @@ public class AiServiceMethodImplementationSupport {
             ChatMemoryFlushStrategy chatMemoryFlushStrategy) {
         CommittableChatMemory committableChatMemory;
         if (chatMemory == null) {
-            committableChatMemory = new NoopChatMemory();
+            committableChatMemory = new LocalCommittableChatMemory();
         } else {
             if (chatMemoryFlushStrategy == ChatMemoryFlushStrategy.IMMEDIATE) {
                 committableChatMemory = new ImmediateFlushChatMemory(chatMemory);
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolWithoutMemoryProviderTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/tools/ToolWithoutMemoryProviderTest.java
@@ -1,0 +1,113 @@
+package io.quarkiverse.langchain4j.test.tools;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+import java.util.function.Supplier;
+
+import jakarta.enterprise.context.control.ActivateRequestContext;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.data.message.ChatMessage;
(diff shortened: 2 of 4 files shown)


core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/AiServiceMethodImplementationSupport.java
```diff
@@ -1385,7 +1389,7 @@ private static CommittableChatMemory determineCommittableChatMemory(ChatMemory c
             ChatMemoryFlushStrategy chatMemoryFlushStrategy) {
         CommittableChatMemory committableChatMemory;
         if (chatMemory == null) {
-            committableChatMemory = new NoopChatMemory();
+            committableChatMemory = new LocalCommittableChatMemory();
```
geoand, line 1392: After this change, is `NoopChatMemory` used anywhere?

JiriOndrusek: > After this change, is `NoopChatMemory` used anywhere?

yes, really lot of tests are using it. Perhaps I can try to move it into `src/test`, WDYT?

geoand: Sure yeah

JiriOndrusek: > Sure yeah

Unfortunately it is not possible, as the test using the noopChatMemory are located in other modules.
So I kept the file and added a  comment explaining it.

geoand (approved):

PR #2726, opened 2026-07-31
Add a ConversationContext API generalizing the definition of the boundaries of a conversation

### LINKED ISSUE #2694: [FEATURE] - Need a way to generically define conversation boundaries
@mariofusco and I have been chatting about this offline a bit. This feature request is to formally document what I'm looking for.

We need a way to generically define conversational boundaries. In the simplest form, think of a human interacting with a chat bot. Each time the user sends a request and receives a response, thats a single interaction. Then the user may ask more questions until they are done. At that point the user either leaves, closes the window, or starts a new chat (essentially clearing the context). That series of interactions is a conversation. In technical-speak, if using WebSockets, the websocket session itself could be considered a conversation.

Conversational boundaries are part of business logic. The application decides when a conversation starts & ends. If one application calls another application, that conversation needs to be propagated over the wire to a downstream service via the [`gen_ai.conversation.id` span attribute from the OpenTelemetry gen ai semantic convensions](https://github.com/open-telemetry/semantic-conventions-genai/blob/712c94fe6f6a165ee0b93d9687286357eb335c7e/model/gen-ai/spans.yaml#L112-L131).

This needs to work regardless of the method of wire transport, whether its over a websocket, stateless http request, async message, or part of an agentic workflow. Anything invoking an AI interaction could also potentially be the beginning of a conversation, or it could be part of an existing conversation.

I started prototyping it a bit integrated with the [Quarkus LangChain4j Chat Scopes Framework](https://docs.quarkiverse.io/quarkus-langchain4j/dev/chat-scopes.html) here: https://github.com/edeandrea/non-deterministic-no-problem/blob/main/src/main/java/ai/scoring/conversation/ConversationalBaggageHandler.java. I used the [CDI lifecycle events](https://docs.quarkiverse.io/quarkus-langchain4j/dev/chat-scopes.html#_chat_scope_cdi_lifecyle_events). My prototype gets the identifier and adds it as OTel baggage. I then have a [custom OTel `SpanProcessor`](https://github.com/quarkiverse/quarkus-langfuse/blob/main/runtime/src/main/java/io/quarkiverse/langfuse/runtime/otel/LangfuseSpanProcessor.java) which [propagates it](https://github.com/quarkiverse/quarkus-langfuse/blob/978f8c34c90c911b023fb8314c1d2b33c2a52d53/runtime/src/main/java/io/quarkiverse/langfuse/runtime/otel/LangfuseSpanProcessor.java#L115-L122) on all spans.

This may not be the correct or final implementation, but it gives you an idea of what needs to happen.

We will also need some kind of lifecycle events that can be listened to when conversations start & end so that we can take some action. This will eventually be used by the [Quarkus Langfuse extension](https://docs.quarkiverse.io/quarkus-langfuse/dev/) to be able to gather interactions into conversations within Langfuse, and then when a conversation ends, trigger a separate evaluation/scoring of the conversation as a whole.

mariofusco:
This is a first draft implementation to provide a generic API to define conversation boundaries (as requested in https://github.com/quarkiverse/quarkus-langchain4j/issues/2694 ) and to add `@ChatScoped` support for agentic systems.

**Motivation**

There are two independent but complementary gaps this PR addresses:

1. No standard way to define conversation boundaries for observability. Chat Scopes define a scope for CDI bean lifecycle and memory isolation, but nothing ties them to OpenTelemetry's `gen_ai.conversation.id` semantic convention. Users working outside Chat Scopes (REST endpoints, message listeners, custom transports) have no API at all to mark where a conversation begins and ends.
2. Agentic systems ignore @ChatScoped. The `AgenticProcessor` hardcodes `@ApplicationScoped` for all agent beans, regardless of what scope annotation is on the interface. Additionally, the `defaultMemoryIdProvider`, used by `@ChatScoped` to  eliminate the need for explicit `@MemoryId` on leaf AI services, is never wired for agents because the agent build path creates a fresh `QuarkusAiServiceContext` that bypasses the metadata where the provider class name is stored.

**What this PR introduces**

1. ConversationContext API (core/runtime)

A lightweight, transport-agnostic API for marking conversation boundaries. It stores the conversation ID in Vert.x duplicated context locals (safe for event loops) and fires `ConversationStarted` / `ConversationEnded` CDI events at boundary transitions. A `ConversationThreadContextProvider` propagates the ID across async boundaries via MicroProfile Context Propagation.

This API is independent of Chat Scopes — it works in any context where a Vert.x duplicated context is available or fallback to a `ThreadLocal` when it isn't.

2. ChatScopeConversationBridge (chat-scopes/core/runtime)

An `@ApplicationScoped` CDI observer that bridges Chat Scope lifecycle events (ChatScopeStarted / ChatScopeEnded) to ConversationContext. Only top-level scopes trigger conversation events, while push/pop transitions within a conversation do not affect the conversation identity. This means Chat Scope users get conversation tracking automatically, with zero configuration.

3. ConversationIdSpanContributor (core/runtime, OpenTelemetry)

A `ChatModelSpanContributor` that sets `gen_ai.conversation.id` on LLM call spans when a conversation is active. Falls back to reading from OTel Baggage for cross-service propagation. The `ListenersProcessor` registers it unconditionally when OpenTelemetry is on the classpath, alongside a matching `ConversationIdToolSpanContributor` for tool execution spans.

4. `@ChatScoped` support for agentic agents

Two gaps prevented `@ChatScoped` from working on agent beans:

- Scope was hardcoded. `AgenticProcessor.cdiSupport()` always set `scope(ApplicationScoped.class)`. Now it reads the scope annotation from the interface (via `CustomScopeAnnotationsBuildItem`) and respects it. If no custom scope is present, it defaults to `@ApplicationScoped` as before.
- `defaultMemoryIdProvider` was lost for agents. The implied-AI-service build path correctly stores the provider class name on the `DeclarativeAiServiceBuildItem`, and `ChatScopeProcessor` correctly sets it when the scope is `@ChatScoped`. But when `QuarkusAiServices.build()` runs for agent-created AI services, it looks up `AiServiceClassCreateInfo` from metadata and that record didn't carry the provider class name. The fix adds `defaultMemoryIdProviderClassName` to `AiServiceClassCreateInfo`, populates it at build time from the `DeclarativeAiServiceBuildItem`, and wires it in `QuarkusAiServices.build()` as a fallback when the context's provider is still `null`. For regular AI services, the provider is already set by `AiServicesRecorder.createDeclarativeAiService()` before build() runs, so the null check prevents double-setting.

5. Known limitation: composite agent root still requires `@MemoryId`

The `@ChatScoped` support for automatic memory ID resolution works fully for leaf agents (`@Agent`), which go through `AiServiceMethodImplementationSupport.memoryId()` and its fallback chain that consults `defaultMemoryIdProvider`.

Composite agents use the `PlannerBasedInvocationHandler` from the `langchain4j-agentic` module, which has its own `memoryId()` method that only checks for `@MemoryId` parameter annotations, it has no fallback to a provider. When no `@MemoryId` is present, it creates an ephemeral scope that is not keyed and not retrievable via `getAgenticScope(id)`.

This means the root of a composite agentic system still needs `@MemoryId` on its method signature for the agentic scope key, even when `@ChatScoped`. The sub-agents (leaf agents) work without `@MemoryId` because their memory ID is automatically derived from `ChatScope.id()`. Fixing this requires adding a `defaultMemoryIdSupplier` fallback to `PlannerBasedInvocationHandler` in the `langchain4j-agentic` module and I will send a pull request to the upstream langchain4j to implement this fix.

6. Chat Scope context fixes

While working on the bridge, a stale-scope bug in `ChatScopeManagedContext` was discovered: on Vert.x duplicated contexts, `ContextLocal.remove()` is a no-op, so a destroyed scope could linger as a stale reference. The `currentContext()` method now checks for destroyed scopes and treats them as absent. The top-level `end()` now calls `remove()` instead of `set(null)` for consistency.

### DIFF
--- a/agentic/deployment/pom.xml
+++ b/agentic/deployment/pom.xml
@@ -63,6 +63,17 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+<dependency>
+      <groupId>io.quarkiverse.langchain4j</groupId>
+      <artifactId>quarkus-langchain4j-chat-scopes-deployment</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-test-vertx</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
--- a/agentic/deployment/pom.xml
+++ b/agentic/deployment/pom.xml
@@ -63,7 +63,7 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
-<dependency>
+    <dependency>
       <groupId>io.quarkiverse.langchain4j</groupId>
       <artifactId>quarkus-langchain4j-chat-scopes-deployment</artifactId>
       <version>${project.version}</version>
--- a/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
+++ b/agentic/deployment/src/main/java/io/quarkiverse/langchain4j/agentic/deployment/AgenticProcessor.java
@@ -63,6 +63,7 @@ import io.quarkiverse.langchain4j.deployment.SkipToolBoxProcessingBuildItem;
 import io.quarkiverse.langchain4j.runtime.NamedConfigUtil;
 import io.quarkus.arc.deployment.BeanContainerBuildItem;
 import io.quarkus.arc.deployment.BeanDiscoveryFinishedBuildItem;
+import io.quarkus.arc.deployment.CustomScopeAnnotationsBuildItem;
 import io.quarkus.arc.deployment.GeneratedBeanBuildItem;
 import io.quarkus.arc.deployment.GeneratedBeanGizmoAdaptor;
 import io.quarkus.arc.deployment.InterceptorResolverBuildItem;
@@ -692,6 +693,7 @@ public class AgenticProcessor {
             InterceptorResolverBuildItem interceptorResolverBuildItem,
             BeanDiscoveryFinishedBuildItem beanDiscovery,
             CombinedIndexBuildItem indexBuildItem,
+            CustomScopeAnnotationsBuildItem customScopes,
             BuildProducer<SyntheticBeanBuildItem> syntheticBeanProducer,
             BuildProducer<RequestChatModelBeanBuildItem> requestChatModelBeanProducer,
             BuildProducer<UnremovableBeanBuildItem> unremovableProducer) {
@@ -728,6 +730,7 @@ public class AgenticProcessor {
                 beanConfigurator = SyntheticBeanBuildItem
                         .configure(detectedAiAgentBuildItem.getIface().name());
             }
+
             beanConfigurator
                     .forceApplicationClass()
                     .unremovable()
@@ -736,7 +739,8 @@ public class AgenticProcessor {
                                     new AiAgentCreateInfo(detectedAiAgentBuildItem.getIface().toString(), chatModelInfo,
                                             hasInterceptorBindings, hasMcpToolBox)))
                     .setRuntimeInit()
-                    .scope(ApplicationScoped.class);
+                    .scope(findScope(customScopes, detectedAiAgentBuildItem));
+
             if (hasInterceptorBindings) {
                 beanConfigurator.injectInterceptionProxy();
             }
@@ -782,6 +786,21 @@ public class AgenticProcessor {
         }
 
         requestedChatModelNames.forEach(name -> requestChatModelBeanProducer.produce(new RequestChatModelBeanBuildItem(name)));
+    }
+
+    private static Class findScope(CustomScopeAnnotationsBuildItem customScopes,
+            DetectedAiAgentBuildItem detectedAiAgentBuildItem) {
+        return customScopes
+                .getScope(detectedAiAgentBuildItem.getIface().annotations())
+                .map(AnnotationInstance::name)
+                .map(scopeDotName -> {
+                    try {
+                        return Class.forName(scopeDotName.toString(), true, Thread.currentThread().getContextClassLoader());
+                    } catch (ClassNotFoundException e) {
+                        throw new RuntimeException("Unable to load scope annotation: " + scopeDotName, e);
+                    }
+                })
+                .orElse((Class) ApplicationScoped.class);
     }
 
     private static boolean isInterceptorBindingAnnotation(AnnotationInstance ann, IndexView index) {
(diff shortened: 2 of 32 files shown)


patriot1burke:
@mariofusco  Wow.  Cool!  Can we have a teams meeting to discuss?  There's some other things I need to talk about too.

geoand:
I was going to suggest the same

patriot1burke:
@mariofusco You on IBM or RHT slack?  I think Eric's on vacation until tuesday

mariofusco:
> @mariofusco You on IBM or RHT slack? I think Eric's on vacation until tuesday

See my reply on IBM slack.

geoand:
Very interesting. Reading the description and quickly looking at the code, I don't understand where the conversation boundaries are

mariofusco:
> Very interesting. Reading the description and quickly looking at the code, I don't understand where the conversation boundaries are

It is possible that I have completely misunderstood @edeandrea's requirements and for this reason I'd appreciate it if we could discuss this in a meeting. Anyway my idea (maybe too simplistic?) was to just provide an API, represented by those `begin()` and `end()` static methods of the new `ConversationContext`, through which users could unequivocally set those boundaries programmatically. The rest of this pull request is to make sure that this API is coherently integrated with @patriot1burke's `ChatScope` and also that this integration, with both the `ConversationContext` and the `ChatScope`, is extended to the agentic module that was left out from the original implementation.

geoand:
Okay thanks. 

We definitely need to discuss

patriot1burke:
I think it is important to have one thing managing the conversation  and making sure that things are wired together.    

To explain what I mean, consider regular Quarkus and all the protocols we support:  Websockets, HTTP, Kafka, Lambda, Azure Functions, Cloud Events, etc...  _**CDI is core to what we do.**_  Each one of the connectors needs to make sure that the CDI RequestContext is started/ended.  WebSockets needs to make sure that the SessionContext is activated.    With LLM apps, IMO, we need conversational state, something structured that isn't just Chat Memory.  I want CDI to be the glue here, which is one of the reasons Chat Scopes was created.    To provide a CDI scope for conversations.  I imagine agent apps are going to have multiple conversations going on at the same time.  

I think its important to identify the entry points into a Q-LC4J application, agentic or otherwise.  If conversation id and conversation state is being propagated into an entry point in the system, then the appropriate `@ChatScope` needs to be wired up and created/activated.  I don't think the SessionContext is enough, for example, a client (specifically a browser client) may be having multiple conversations through the same shared web socket connection.  I think a conversational CDI scope is crucial to everything that's happening in Q-LC4J:  Chat apps, and agentic flows etc.

If an LC4J component is doing something async, we need to make sure ManagedExecutors are being used so that CDI contexts are propagated to the new thread context.

We need to make sure every LC4J component can be a bean so that it can inject the enterprisey things it needs and also provide it a lightweight event system.  All the advantages of CDI need to be available to LC4J components.

The lifecycle of a conversation is as follows:
* begin
* active
* deactivate
* end
*  Maybe a passivate event too 

These lifecycle events effect the design of a conversationcontext API.

Anyways, these are my thoughts.  Hopefully not too rambling.

PR #2728, opened 2026-08-01
fix: generate JSON schema for collection AI service return types

### LINKED ISSUE #943: handle collections as return types or error
if one use List, Set, Map etc. as return types the ask to genAI is: `You must answer strictly in the following JSON format: {\n}` - it should have the the json structure.

This fails:

```
    @RegisterAiService
    public interface jaidataAI {

        @SystemMessage(
            """
            IDENTITY
            
            You are a superintelligent AI that finds all mentions of data structures within an input and...
        """
        )
        @UserMessage("""
                    {data}
                """)
        List<FileInfo> generate(String data);
    }

    record FileInfo(String filename, String content) {};
```

but if change it to the following it works:

```
     FileInfos generate(String data);
    }
    record FileInfos(List<FileInfo>) {};
    record FileInfo(String filename, String content) {};
```

I would think it should be possible to handle - and if not, throw build error it needs to be a non-collection java type?

arimu1:
## What was wrong

AI service methods that return `List`/`Set` (for example `List<FileInfo>`) produced an empty structured-output schema. The prompt path could also end up with a useless `{}` format, so the model was not told the real shape of the response. Wrapping the list in a record worked as a workaround.

## Change

`AiServicesProcessor` was building the response schema with `JsonSchemas.jsonSchemaFrom()`, which only supports POJOs. Upstream `ServiceOutputParser.jsonSchema()` already knows how to describe collections (same path as vanilla LangChain4j `DefaultAiServices`).

This switches schema generation to `ServiceOutputParser.jsonSchema()`. For types where prompt format instructions are not available but a JSON schema is (e.g. `List` of a POJO), the build no longer fails with a bare `IllegalStateException`; the schema is kept and used when the model supports structured output. If neither schema nor format instructions can be produced, we fail with a clearer configuration error.

## Test

Added `CollectionStructuredOutputResponseTest` next to the existing structured-output OpenAI test. It checks that a `List<FileInfo>` return type:

- boots successfully
- sends `response_format.json_schema` named `List_of_FileInfo` with a non-empty `values` array schema
- parses a `{"values":[...]}` response

Locally: `./mvnw -pl model-providers/openai/openai-vanilla/deployment -am test -Dtest=StructuredOutputResponseTest,CollectionStructuredOutputResponseTest`

---

- Fixes: #943

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/AiServicesProcessor.java
@@ -73,6 +73,7 @@ import org.objectweb.asm.tree.analysis.AnalyzerException;
 
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 
+import dev.langchain4j.exception.UnsupportedFeatureException;
 import dev.langchain4j.guardrail.OutputGuardrail;
 import dev.langchain4j.invocation.InvocationParameters;
 import dev.langchain4j.memory.ChatMemory;
@@ -81,7 +82,6 @@ import dev.langchain4j.model.chat.request.json.JsonSchema;
 import dev.langchain4j.service.IllegalConfigurationException;
 import dev.langchain4j.service.Moderate;
 import dev.langchain4j.service.memory.ChatMemoryAccess;
-import dev.langchain4j.service.output.JsonSchemas;
 import dev.langchain4j.service.output.ServiceOutputParser;
 import dev.langchain4j.service.tool.ToolArgumentsErrorHandler;
 import dev.langchain4j.service.tool.ToolErrorContext;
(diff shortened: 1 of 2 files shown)


geoand:
Are you sure that #943 is still an issue?

arimu1:
@geoand Yes — #943 is still present on current `main`.

`AiServicesProcessor.jsonSchemaFrom` still delegates to `JsonSchemas.jsonSchemaFrom(returnType)` only:

```java
return JsonSchemas.jsonSchemaFrom(returnType);
```

That path supports POJOs, not `List`/`Set` return types, so AI service methods like `List<FileInfo> generate(...)` still get an empty structured-output schema / useless `{}` format instructions (the original report).

This PR switches that helper to `ServiceOutputParser.jsonSchema` (same path vanilla LangChain4j `DefaultAiServices` uses), which describes collections, and adds `CollectionStructuredOutputResponseTest`. Happy to adjust if you prefer a different approach.

geoand:
So does the new `CollectionStructuredOutputResponseTest` fail if the change in `AiServicesProcessor` is not applied?

arimu1:
@geoand good questions — answers against tip `1acbd58ba`:

### Is #943 still an issue?
Yes on current main. `JsonSchemas.jsonSchemaFrom(List<POJO>)` still yields empty/`{}` schema, so collection AI-service return types do not get a usable structured-output schema. The new `CollectionStructuredOutputResponseTest` is the regression for that.

### Does the test fail without the `AiServicesProcessor` change?
Yes. Without routing through `ServiceOutputParser.jsonSchema`, the collection return type does not produce a non-empty schema and the test’s assertion on the generated schema / structured output path fails.

### Agentic/MCP CI
The first push was too aggressive: `ServiceOutputParser.jsonSchema` throws `UnsupportedFeatureException` for non-schemaable types such as `AgenticScope`. Tip now catches that and treats it as “no schema” so agentic/mcp modules keep building while #943 still gets real schemas for `List`/`Set` of POJOs.

geoand:
Please squash the commits

arimu1:
@geoand Squashed to a single commit on tip `fcddbf1de` (force-with-lease). Ready for re-review when convenient.

arimu1:
@geoand Fixed impsort on `AiServicesProcessor.java` (unsorted `UnsupportedFeatureException` import after the UnsupportedFeatureException catch change). Still a single squashed commit — tip `23f81d896`. That was the Quick Build + Documentation check failure.

geoand (approved):

PR #2733, opened 2026-08-04
feat(samples): add Infinispan embedding store example

### LINKED ISSUE #328: Contribute an example with Infinispan
The chatbot example works with Infinispan
https://github.com/karesti/quarkus-langchain4j/tree/chatbot-infinispan/samples/chatbot-infinispan

A diff example might be beneficial to showcase how it works
Any ideas ?

love9rose:
## Description
Add a new sample `infinispan-embedding-store` demonstrating the usage of
`quarkus-langchain4j-infinispan` as a vector store for semantic search.

- Closes: #328


karesti (changes requested):
Thanks for the contribution with the example !

It would be great to cover the example with a test and use a local embedding model to avoid the need of Open AI, and provide the documentation to switch to Open AI

Issue #2734, opened 2026-08-07
Response JSON schema is derived from the Result<T> wrapper instead of its type parameter


sqiao97:
### Describe the bug

When an AI service method returns `Result<T>`, the JSON schema sent to the model describes `dev.langchain4j.service.Result` itself rather than `T`.

The output parser does *not* have this problem — it correctly unwraps `Result<T>` and parses the response body as `T`. So the schema and the parser disagree about what the model is supposed to return:

- The model is told to produce `{ "content": {...}, "tokenUsage": {...}, "sources": [...] }`
- The parser tries to read the response as `T`

A model that faithfully follows the schema therefore returns a shape the parser cannot map, and **every field of `T` silently deserialises to `null`** — no exception, no warning. The same method returning `T` directly produces the correct schema.

### Expected behavior

For `Result<Sanitized> sanitize(...)`, the schema should describe `Sanitized` (the same schema produced when the method returns `Sanitized` directly), because that is what the parser will read.

### Actual behavior

```
name     = Result
props    = [content, tokenUsage, sources, finishReason, toolExecutions, intermediateResponses, finalResponse]
required = []
```

### To Reproduce

```java
@CreatedAware
public interface SanitizerService {

    @SystemMessage("Rewrite the user's question into search queries.")
    Result<Sanitized> sanitize(@UserMessage String question);

    record Sanitized(String keywordQuery, String semanticQuery) {}
}
```

```java
@QuarkusTest
class SchemaReproTest {

    /** Captures the outgoing request; advertises JSON-schema support so a schema is attached. */
    static final class RecordingChatModel implements ChatModel {
        ChatRequest captured;

        @Override
        public Set<Capability> supportedCapabilities() {
            return Set.of(Capability.RESPONSE_FORMAT_JSON_SCHEMA);
        }

        @Override
        public ChatResponse chat(ChatRequest request) {
            this.captured = request;
            return ChatResponse.builder()
                    .aiMessage(AiMessage.from("{\"keywordQuery\":\"k\",\"semanticQuery\":\"s\"}"))
                    .build();
        }
    }

    @Test
    void schemaDescribesResultWrapperInsteadOfPayload() {
        RecordingChatModel model = new RecordingChatModel();

        var result = AiServices.builder(SanitizerService.class)
                .chatModel(model)
                .build()
                .sanitize("how do I reset my password?");

        var schema = model.captured.responseFormat().jsonSchema();
[...]

PR #2737, opened 2026-08-09
Cover JSON schema derivation for Result<T> AI service return types

omatheusmesmo:
## Describe your changes

A `Result<T>` AI service method must advertise the schema of `T`, not of the `Result` wrapper, because `ServiceOutputParser` unwraps `Result` before parsing the response body. When the two disagree, a model that faithfully follows the advertised schema returns a shape the parser cannot map, and every field of `T` deserialises to null with no exception and nothing in the logs.

This is already correct on `main`, but only as a side effect of `AiServicesProcessor#jsonSchemaFrom` moving from `JsonSchemas.jsonSchemaFrom` to `ServiceOutputParser#jsonSchema` in 23f81d89. Nothing pins it: the only existing `Result` coverage is `Result<String>` in `AiServicesTest`, and `String` short-circuits before the unwrapping, so it produces no schema at all. The `Result<POJO>` path, the one that actually exercises the unwrapping, was never asserted.

The test follows `StructuredOutputResponseTest` and `CollectionStructuredOutputResponseTest` in the same package: a `@RegisterAiService` with `Result<Person>`, OpenAI configured with `response-format=json_schema` and `strict-json-schema=true`, and assertions on the `response_format.json_schema` actually sent over the wire. Reverting that one line to `JsonSchemas.jsonSchemaFrom` makes it fail with `["name"="Result" (expected: "Person")]`.

---

- Relates to #2734

### DIFF
--- a/model-providers/openai/openai-vanilla/deployment/src/test/java/org/acme/examples/aiservices/ResultStructuredOutputResponseTest.java
+++ b/model-providers/openai/openai-vanilla/deployment/src/test/java/org/acme/examples/aiservices/ResultStructuredOutputResponseTest.java
@@ -1,0 +1,74 @@
+package org.acme.examples.aiservices;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.InstanceOfAssertFactories.map;
+
+import java.io.IOException;
+import java.time.LocalDate;
+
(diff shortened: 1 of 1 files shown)


geoand (approved):
Thanks!

PR #2741, opened 2026-08-13
Update to langchain4j-watsonx 1.19.0 and add new services

andreadimaio:
Update the `langchain4j-watsonx` upstream dependency to `1.19.0,` which introduces new models and services.

**Chat model backends**
The chat configuration is now split into three dedicated namespaces, only one of which can be active per configuration:

  - **chat-model**  → foundation model API  _(/ml/v1/text/chat)_
  - **deployment-chat-model** → deployed model API  _(/ml/v1/deployments/{id}/text/chat)_
  - **gateway-chat-model**  → Model Gateway API _(/ml/gateway/v1/chat/completions)_

Setting more than one backend (e.g. both `deployment-chat-model.deployment-id` and `gateway-chat-model.model-name`) throws a `ConfigValidationException` at startup. To run multiple backends in the same application, declare each one under a named configuration and select it with `@ModelName`.

**New schema services**
Four new CDI beans are now available to generate and manage document schemas without writing them by hand:

  - `CreateSchemaService`  - generates a schema from a sample document in COS
  - `ImproveSchemaService` - enriches an existing schema with better descriptions
  - `MergeSchemaService` - combines multiple schemas into one
  - `ClusterSchemaService` - groups schemas by semantic similarity

A generated schema can be passed directly to `TextExtractionService` or `TextClassificationService` as the semantic configuration.

**Other changes**
  - `frequencyPenalty`, `presencePenalty`, `topP`, `logprobs` and `stop` are now optional and sent to the model only when explicitly set
  - strict JSON Schema mode is now the default (set `strict-json-schema=false` to opt out)

**NOTE:** This PR can be merged once langchain4j 1.19.0 has been picked up by quarkus-langchain4j.


andreadimaio:
Merged into #2723

PR #2742, opened 2026-08-14
Cover case with conditional system message provider  in samples

fedinskiy:
- Add a new samle for multimodule demonstration
- Lower logging level for debugging in chatbot sample


omatheusmesmo (changes requested):
Hi @fedinskiy , 

It makes more sense to create a new sample to explore multi-model names or providers instead of modifying an existing one that wasn't designed for this purpose. You can refer to how the same provider is handled in #2743 for guidance.

fedinskiy:
@omatheusmesmo fair point. Done.

geoand:
I can't say I understand this one. What prompted it?

fedinskiy:
@geoand I wanted to create more detailed coverage for https://github.com/quarkiverse/quarkus-langchain4j/pull/2563

geoand:
Okay, but samples are meant to help users of Quarkus LangChain4j get going for various scenarios. Do you really think this is something people need?

fedinskiy:
@geoand this patch definitely shows, how can the provider be used. Whether people need this, I can not guess.
If you think, that this PR must be improved — please tell me, what should be done. If you think, that such sample should not be added at all — fell free to close the PR.

geoand:
It's not a matter of improvement, it's a matter of whether users benefit from this. Anyway, let's see what @jmartisk thinks

jmartisk:
Yeah TBH I don't see much value in adding this as a sample project (also with all the maintenance burden that comes along with it...). Besides, generally, most models know what kind of model they are, no? There's no need to tell them this way, is there? If there is, I would make this into a documentation piece instead.

Btw., I think we should stop naming the AI "Bob" in our samples... For obvious reasons.

geoand:
> Btw., I think we should stop naming the AI "Bob" in our samples... For obvious reasons.

So true :)

PR #2746, opened 2026-08-17
Preserve generic tool argument types in generated mappers

filipaggoncalves:
Goal: Preserve concrete generic type metadata on generated tool argument-mapper fields.

This lets Jackson deserialize typed tool arguments correctly, such as `List<MyEnum>`, `List<MyDto>`, and `Map<String, MyEnum>`. 

What is currently happening without these changes is: for a tool argument of `List<SomeEnum>`, the AI Service receives and handles it properly, but the framework loses the typing of `T` in the process and treats it as `List<String>` which then throws when trying to fit it back into the expected argument type of `List<SomeEnum>`.

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/ToolProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/ToolProcessor.java
@@ -32,6 +32,7 @@ import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.FieldInfo;
+import org.jboss.jandex.GenericSignature;
 import org.jboss.jandex.IndexView;
 import org.jboss.jandex.MethodInfo;
 import org.jboss.jandex.MethodParameterInfo;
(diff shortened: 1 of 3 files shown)


core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/ToolProcessor.java
```diff
@@ -1229,4 +1238,46 @@ private static class ObjectMapperHolder {
         private static final TypeReference<Map<String, Object>> MAP_TYPE_REF = new TypeReference<>() {
         };
     }
+
+    private static String genericFieldSignature(Type type) {
+        if (containsUnresolvedTypeVariable(type)) {
+            throw new IllegalArgumentException("Tool parameter type must not contain unresolved type variables: " + type);
```
geoand, line 1244: Why not? What happens if they do?

geoand (commented):
Nice change! I added one comment

core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/ToolProcessor.java
filipaggoncalves: If we define a tool parameter as an unresolved type variable, such as `List<T>`, it will fail even without this condition, when it hits the `toJsonSchemaElement` method as it fails to map the parameter and throws `Unsupported type: T`. I felt that since i was already tweaking the generic mapping, it would make sense to raise this exception clearly here.

geoand: Got it, thanks!

PR #2749, opened 2026-08-18
Document tool idempotency considerations with fault tolerance

kittimzhe:
Discussed with @geoand and @cescoffier: instead of a build-time check, the guidance belongs in the documentation — tools are expected to be idempotent, and that contract should be stated where users read about fault tolerance.

This PR now:

- reverts the build-time fault tolerance check (and its test)
- keeps and expands the `CAUTION` note in `guide-fault-tolerance.adoc`, explaining that retrying an AI service method restarts the agent loop and that tools must be idempotent

### DIFF
--- a/docs/modules/ROOT/pages/guide-fault-tolerance.adoc
+++ b/docs/modules/ROOT/pages/guide-fault-tolerance.adoc
@@ -33,6 +33,8 @@ include::{examples-dir}/io/quarkiverse/langchain4j/samples/AiServiceWithFaultTol
 
 NOTE: Use `@Timeout` carefully when using tools or multi-step function calls, as those may require more time to complete due to multiple background interactions with your application's tools or services.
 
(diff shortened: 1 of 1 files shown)


omatheusmesmo (commented):
LGTM

geoand:
I don't know about this TBH. I believe having a warning in the docs is more appropriate.

@cescoffier WDYT?

cescoffier:
I agree - I would only write a warning in the doc.

cescoffier:
For me it's normal that tools are retried, the developers must implement idempotency correctly.

geoand:
Exactly

kittimzhe:
Thanks both for the feedback! Agreed — tool idempotency is the developer's responsibility, and a documentation note is the right place for this.

I'll strip the build-time check and keep only the documentation warning (expanded with the idempotency guidance).

geoand (approved):
Thanks

geoand:
Please squash the commits

kittimzhe:
Done — squashed into a single commit and rebased on top of current main.

geoand (approved):

PR #2752, opened 2026-08-18
Make Anthropic topK optional for models that do not support top_k

### LINKED ISSUE #2751: Make Anthropic `topK` optional for models that do not support `top_k`
**Describe the bug**

When `quarkus.langchain4j.anthropic.chat-model.top-k` is not configured, the Anthropic recorder sends `top_k=40` by default for both regular and streaming chat models.

Anthropic Sonnet 5 does not support this parameter and rejects the request with an HTTP 400 error. The relevant code is present in `AnthropicRecorder.java`:

```java
builder.topK(chatModelConfig.topK().orElse(40));
```

The streaming builder applies the same default:

```java
.topK(chatModelConfig.topK().orElse(40))
```

**Log and Stack trace**

```text
2026-08-18 10:06:18,794 INFO  [io.quarkiverse.langchain4j.anthropic.QuarkusAnthropicClient$AnthropicClientLogger] (vert.x-eventloop-thread-13) Response:
- status code: 400
- headers: [Date: Tue, 18 Aug 2026 09:06:18 GMT], [Content-Type: application/json], [Transfer-Encoding: chunked], [Connection: keep-alive], [x-should-retry: false], [request-id: req_011Ce9wx6GkgnwtisQbsk1nw], [strict-transport-security: max-age=31536000; includeSubDomains; preload], [anthropic-organization-id: 4e3d74ac-7493-4548-bdd6-04ae1a23bc84], [anthropic-workspace-id: wrkspc_01K1Arc18D87bqA5KLfwAg4L], [traceresponse: 00-34c28c140ac680c72a1bf35e9fc99b18-e5c8a16c3cdb20d6-01], [Server: cloudflare], [vary: Accept-Encoding], [Content-Security-Policy: default-src 'none'; frame-ancestors 'none'], [cf-cache-status: DYNAMIC], [X-Robots-Tag: none], [CF-RAY: a2cfc3020c495bd4-LIS]
- body: {"type":"error","error":{"type":"invalid_request_error","message":"`top_k` is deprecated for this model."},"request_id":"req_011Ce9wx6GkgnwtisQbsk1nw"}
```

**To Reproduce**

1. Enable the Anthropic integration without configuring `top-k`:

   ```properties
   quarkus.langchain4j.anthropic.api-key=${ANTHROPIC_API_KEY}
   quarkus.langchain4j.anthropic.chat-model.model-name=claude-sonnet-5
   ```

2. Create and invoke an Anthropic chat or streaming chat model.
3. The request fails because `top_k=40` is sent implicitly.

**Expected behavior**

The `top_k` parameter should be omitted when `top-k` is not explicitly configured. An explicitly configured `top-k` value should continue to be sent. This behavior should be consistent for regular and streaming Anthropic chat models.

**Please complete the following information:**
- LangChain4j version: 1.17.2
- LLM(s) used: Anthropic Sonnet 5

**Additional context**

The affected file is `model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/runtime/AnthropicRecorder.java`.

The thinking configuration already clears `topK` by calling `builder.topK(null)`, but the default value is applied before that logic. Removing the implicit default would allow models that reject `top_k` to work without requiring model-specific configuration.

pedrolopix:
### What changed

- Updated Anthropic chat and streaming clients to include `top_k` only when explicitly configured.
- Added tests covering configured and unset sampling parameters.
- Updated smoke-test expectations to reflect the omitted default `top_k`.

### Why

Anthropic should not receive an implicit `top_k: 40` when the parameter is unset. This change preserves provider defaults and ensures `top_k`, `top_p`, and `temperature` are sent only when configured.

- Closes: #2751


omatheusmesmo (commented):
Fix looks right. Built the module locally, anthropic deployment suite green (11 tests).

Three small things:

1. `ChatModelConfig.java:51` still has `@ConfigDocDefault("40")` on `topK()`. With the default gone, the generated docs would keep advertising 40.

2. In `AnthropicSamplingParametersUnsetTest`, the three `overrideRuntimeConfigKey(..., null)` calls are no-ops (SmallRye reads a null value as absent). I removed them locally and the test still passes.

3. That test overlaps with the two smoke tests updated here, which already assert the body without `top_k`. `AnthropicSamplingParametersTest` is the one adding coverage, but only for `ChatModel` - injecting `StreamingChatModel` there would cover the streaming change too

geoand:
Can you please squash the commits?

pedrolopix:
The commit https://github.com/quarkiverse/quarkus-langchain4j/commit/5bb0d2e7aedcb26e71e84eaff388261a7ec67fa8 resolves this issue.

Issue #2690, opened 2026-07-22
Multiple named pgvector stores sharing the same datasource causes duplicate bean error


edeandrea:
## Description

When configuring multiple named pgvector embedding stores that share the same default datasource, the application fails to start with a duplicate synthetic bean registration error.

The `PgVectorAgroalPoolInterceptor` bean is registered once per named store with the same bean identifier. When two or more named stores share the same datasource, this causes a collision.

## Error

```
java.lang.IllegalStateException: A synthetic bean with identifier
qRjk8s5CiB1hAOpwW8ZGm_odoWQ is already registered: SYNTHETIC bean
[types=[io.agroal.api.AgroalPoolInterceptor,
io.quarkiverse.langchain4j.pgvector.PgVectorAgroalPoolInterceptor,
java.lang.Object], qualifiers=[@Any, @Default], target=n/a]
```

## Configuration

```properties
quarkus.langchain4j.pgvector.default-store-enabled=false

quarkus.langchain4j.pgvector."store-a".datasource=<default>
quarkus.langchain4j.pgvector."store-a".table=store_a
quarkus.langchain4j.pgvector."store-a".dimension=768

quarkus.langchain4j.pgvector."store-b".datasource=<default>
quarkus.langchain4j.pgvector."store-b".table=store_b
quarkus.langchain4j.pgvector."store-b".dimension=768
```

## Reproducer

https://github.com/edeandrea/pgvector-named-stores-reproducer

Clone and run `./mvnw test` to reproduce.

## Root Cause

PR #2409 implemented named pgvector stores assuming each uses a **distinct** datasource. The `PgVectorAgroalPoolInterceptor` bean registration in `PgVectorEmbeddingStoreProcessor.createBean()` appears to register the interceptor per named store without deduplicating by datasource. When multiple named stores share the same datasource, the interceptor is registered multiple times with the same identifier.

## Expected Behavior

Multiple named stores sharing the same datasource should work. The interceptor should be registered once per unique datasource, not once per named store.

## Environment

- Quarkus: 3.37.3
- quarkus-langchain4j-pgvector: managed by quarkus-langchain4j-bom 3.37.3
- Java: 25

geoand:
Thanks for the reproducer.

I'll check tomorrow

PR #2755, opened 2026-08-19
Discover named embedding stores configured with runtime properties only

### LINKED ISSUE #2754: pgvector named stores not discovered without explicit datasource property
## Description

When configuring multiple named pgvector embedding stores, the build-time processor (`PgVectorEmbeddingStoreProcessor.createBean()`) does not discover named stores unless the `datasource` build-time property is set explicitly for each store — even when other runtime properties (`table`, `dimension`) are configured.

## Expected behavior

Named stores should be discovered from any configured property (build-time or runtime), not just build-time properties. The `datasource` property [defaults to `<default>`](https://docs.quarkiverse.io/quarkus-langchain4j/dev/includes/quarkus-all-config.html#quarkus-langchain4j-pgvector_quarkus-langchain4j-pgvector-store-name-datasource), so it shouldn't need to be set explicitly when using the default datasource.

## Actual behavior

Without explicit `datasource` for each named store, the application fails to start with `UnsatisfiedResolutionException` for all `@EmbeddingStoreName`-qualified `EmbeddingStore` beans.

## Root cause

`PgVectorEmbeddingStoreProcessor.createBean()` iterates `buildTimeConfig.namedConfig().entrySet()`. This map uses `@WithParentName @WithDefaults Map<String, PgVectorNamedStoreBuildTimeConfig>`, where `PgVectorNamedStoreBuildTimeConfig` has only one property: `datasource`. SmallRye Config only materializes map keys for entries with explicitly set properties matching the config group, so runtime-only properties (`table`, `dimension` from `PgVectorStoreRuntimeConfig`) do not trigger key discovery.

## Reproducer

Configuration that **fails**:
```yaml
quarkus:
  langchain4j:
    pgvector:
      default-store-enabled: false
      my-store:
        table: my_table
        dimension: 768
```

Configuration that **works** (workaround):
```yaml
quarkus:
  langchain4j:
    pgvector:
      default-store-enabled: false
      my-store:
        datasource: "<default>"
        table: my_table
        dimension: 768
```

## Notes

- All existing integration tests in the repo set `datasource` explicitly, including `PgVectorMultipleNamedStoresSameDataStoreTest` which uses `datasource: "<default>"`.
- This affects only pgvector; the Qdrant extension's named store build-time config has additional properties (`collectionName`, `payloadTextKey`, `clientName`) that make discovery more likely.

omatheusmesmo:
## Describe your changes

Named stores declare their build-time config as `@WithParentName @WithDefaults Map<String, XNamedStoreBuildTimeConfig>`, and SmallRye Config only materializes a map key for properties that structurally match that group. Those groups hold one to three properties, so a store configured through runtime properties alone (`table`, `dimension`, `url`, ...) never produced a key and its beans were never created, failing with `UnsatisfiedResolutionException`.

Reported for pgvector, but it affects every store whose actual settings live in a `@ConfigRoot(RUN_TIME)`: chroma, milvus, neo4j, oracle, pgvector, pinecone, redis and weaviate. Qdrant is the exception, since all of its store properties are build-time ones.

Changes:

- Add `NamedConfigDiscovery` to `core/deployment`. It scans the configured property names and treats `<prefix>.<name>.<property>` as a named store unless `<name>` belongs to the unnamed store. The unnamed property names come from `ConfigMappings.getProperties()`, so `@WithName`, `@WithParentName`, nested groups and indexed collections stay SmallRye's business instead of being reimplemented here.
- Use it in the eight affected processors, and in the chroma and weaviate Dev Services processors so discovered stores also get their URL injected.
- Normalize the pgvector `PgVectorAgroalPoolInterceptor` dedup key. It was keyed on the raw datasource string, where an unset `datasource` (`null`) and an explicit `"<default>"` resolve to the same `@Default` qualifier, which would reintroduce the duplicate bean from #2690 now that discovery works.

Each affected store gets a regression test configuring named stores without the build-time property, verified to fail without the fix. Qdrant gets one too, with no change to `QdrantProcessor`, pinning down that it is not affected. The unit tests on `NamedConfigDiscovery` cover the two cases where a non-store segment could be mistaken for a store name: the root `devservices` property of chroma, milvus and weaviate, and indexed collections such as the documented `quarkus.langchain4j.oracle.metadata-indexes[0].create-option`.

---

- Closes #2754

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/NamedConfigDiscovery.java
+++ b/core/deployment/src/main/java/io/quarkiverse/langchain4j/deployment/NamedConfigDiscovery.java
@@ -1,0 +1,97 @@
+package io.quarkiverse.langchain4j.deployment;
+
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import org.eclipse.microprofile.config.ConfigProvider;
+
+import io.smallrye.config.ConfigMappings;
+import io.smallrye.config.ConfigMappings.ConfigClass;
+import io.smallrye.config.NameIterator;
+
+/**
+ * Discovers the keys of a named configuration map, such as the named embedding stores of an extension.
+ * <p>
+ * Extensions declare their named configurations as {@code @WithParentName @WithDefaults Map<String, G>}, which SmallRye
+ * Config only materializes keys for when a configured property structurally matches the config group {@code G}. A
+ * build-time group usually holds a couple of properties at most, so a named entry configured through runtime properties
+ * alone never produces a key, and the corresponding beans are never created.
+ * <p>
+ * This scans the configured property names instead, treating any {@code <prefix>.<name>.<property>} property as a named
+ * entry unless {@code <name>} is a property of the unnamed configuration.
+ */
+public final class NamedConfigDiscovery {
+
+    private NamedConfigDiscovery() {
+    }
+
+    /**
+     * Discovers the named configuration keys from the properties of the current {@code Config}.
+     *
+     * @param configPrefix the config prefix the named configurations live under, without a trailing dot,
+     *        for instance {@code quarkus.langchain4j.pgvector}
+     * @param knownNames the names SmallRye Config already materialized, kept first in the returned set
+     * @param mappingClasses the config mapping roots covering the configuration, across all config phases
+     */
+    public static Set<String> discoverNames(String configPrefix, Set<String> knownNames, Class<?>... mappingClasses) {
+        return discoverNames(ConfigProvider.getConfig().getPropertyNames(), configPrefix, knownNames, mappingClasses);
+    }
+
+    /**
+     * Discovers the named configuration keys from the given property names.
+     */
+    public static Set<String> discoverNames(Iterable<String> propertyNames, String configPrefix, Set<String> knownNames,
+            Class<?>... mappingClasses) {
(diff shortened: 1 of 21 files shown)


omatheusmesmo:
Hi @edeandrea , could you please give it a try? Thanks.

edeandrea:
Hi @omatheusmesmo ! I just tried this out and it works perfectly!

geoand:
I don't agree with the premise of the originally reported issue. 
Reading runtime properties at build time (even if they are in the application's static `application.properties`) can lead to multiple problems (which we've had multiple times in the Quarkus repo). 

@radcortez has created [this](https://github.com/quarkusio/quarkus/discussions/51105) discussion, which describes how such cases should be handled.

geoand (changes requested):
See my previous comment. Given that, the issue the PR tries to address does not make sense

radcortez:
Please do not read configuration that is marked as runtime during build time. Yes, it may seem to work programmatically because `application.properties` is present at both build time and runtime, but such properties can be overridden at runtime, by removing / adding names, which would cause issues.

Unfortunately, we relied on this too much for a lot of other stuff, which has been a major pain point. We have been slowly moving away from it.

We plan to eventually lock down access to runtime config even at build time:
https://github.com/quarkusio/quarkus/issues/42985

So, my recommendation: do not do this, because it will stop working in the future.

PR #2759, opened 2026-08-20
Report token usage from the Ollama embedding model

### LINKED ISSUE #2758: Embedding models (Ollama, HuggingFace) do not return token usage information
OllamaEmbeddingModel.embedAll() and HuggingFaceEmbeddingModel do not return token usage information. The
  EmbeddingResponse DTO only deserializes the embeddings field and discards prompt_eval_count from the Ollama /api/embed
  JSON response. The result is wrapped with Response.from(embeddings) (single-arg), which sets tokenUsage to null.

  The upstream LangChain4j core (langchain4j-ollama) handles this correctly — its OllamaEmbeddingModel reads
  promptEvalCount and returns it via TokenUsage.

  For comparison, the Azure OpenAI embedding model in this project also works correctly, using Response.from(embeddings, 
  new TokenUsage(inputTokenCount)).

 ` Expected behavior`

  Embedding models should return Response.from(embeddings, tokenUsage) with TokenUsage populated from the provider's
  response.
  
`  Suggested fix`

  For Ollama:
  1. Add prompt_eval_count field to EmbeddingResponse DTO
  2. Construct TokenUsage from it in OllamaEmbeddingModel.embedAll()
  3. Return Response.from(embeddings, tokenUsage) instead of Response.from(embeddings)

  Similar changes may be needed for HuggingFace.

omatheusmesmo:
## Describe your changes

`OllamaEmbeddingModel.embedAll()` always returned `Response.from(embeddings)`, so `tokenUsage()` was `null` for every embedding call. The `/api/embed` response does carry `prompt_eval_count`, but `EmbeddingResponse` never deserialized it.

`EmbeddingResponse` now maps `prompt_eval_count`, and `embedAll()` accumulates it into a `TokenUsage` across the per-segment requests it issues, returning `Response.from(embeddings, tokenUsage)`. The field is optional in the Ollama API (it is absent from the multi-input example in the API docs), so `tokenUsage` stays `null` when the provider does not report it, rather than reporting zero.

Covered by `OllamaEmbeddingModelTokenUsageTest` in `ollama/deployment`, following the existing WireMock test conventions: a single call, accumulation across segments, and the absent-field case.

The issue also asks about HuggingFace. That one is a provider limitation, not something fixable here: the `feature-extraction` endpoint returns a bare array of vectors with no usage metadata, so `HuggingFaceClient.embed()` returns `List<float[]>` and there is no token count to read. Upstream LangChain4j returns no usage there for the same reason.

---

- Closes #2758

### DIFF
--- a/model-providers/ollama/deployment/src/test/java/io/quarkiverse/langchain4j/ollama/deployment/OllamaEmbeddingModelTokenUsageTest.java
+++ b/model-providers/ollama/deployment/src/test/java/io/quarkiverse/langchain4j/ollama/deployment/OllamaEmbeddingModelTokenUsageTest.java
@@ -1,0 +1,116 @@
+package io.quarkiverse.langchain4j.ollama.deployment;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.equalTo;
+import static com.github.tomakehurst.wiremock.client.WireMock.matchingJsonPath;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+
+import jakarta.inject.Inject;
+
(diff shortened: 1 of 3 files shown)


geoand (approved):
Thanks!

Issue #2750, opened 2026-08-18
OpenAI API key can not be retrieved from environment variable


fedinskiy:
According to both Quarkus manual[1] environment variables can be used instead of maven properties. Readme[2] file for chatbot sample explicitly recommends using `QUARKUS_LANGCHAIN4J_OPENAI_API_KEY` to store API keys. An attempt to do that will lead to the following error:
```
Configuration validation failed:
	SRCFG00014: The config property quarkus.langchain4j.openai.api-key is required but it could not be found in any config source
```

How to reproduce:
1. Clone this repo (`main` branch, `0f4723e6e2e88fd0ee3dbee5ef83feed0c60a9bc` at the time of writing) and enter `samples/chatbot` directory
2. Put the key into environment variable: `QUARKUS_LANGCHAIN4J_OPENAI_API_KEY=<the key>`
3. Start the app in the dev mode: `mvn quarkus:dev `
4. When the same key is used inside maven property, the app works as expected: `mvn quarkus:dev -Dquarkus.langchain4j.openai.api-key=${QUARKUS_LANGCHAIN4J_OPENAI_API_KEY}`

[1] https://quarkus.io/guides/config-reference#environment-variables
[2] https://github.com/quarkiverse/quarkus-langchain4j/tree/main/samples/chatbot#running-the-example

geoand:
I tried `QUARKUS_LANGCHAIN4J_OPENAI_API_KEY=.... mvn quarkus:dev` on the `chatbot` sample and it worked just fine

fedinskiy:
@geoand I checked, and surprisingly this worked as well.
So:
- ` QUARKUS_LANGCHAIN4J_OPENAI_API_KEY=X  mvn quarkus:dev` works
- `mvn quarkus:dev -Dquarkus.langchain4j.openai.api-key=X` works
- `QUARKUS_LANGCHAIN4J_OPENAI_API_KEY=X  && mvn quarkus:dev` or setting the value on a separate lines doesn't

It looks like the code checks the properties incorrectly somehow.

UPD: adding the value to .env file works

fedinskiy:
Ok, it looks like the second step in the reproducer is missing `export` command. Sorry for the inconvenience

geoand:
NP, thanks for checking again

PR #2761, opened 2026-08-21
Improve error message for missing OpenAI API key configuration

kittimzhe:
## What changed

Improved the error message in `OpenAiRecorder.createConfigProblem()` to provide actionable guidance when the OpenAI API key is not configured.

## Why

Issue #2750 reports that setting the API key via the `QUARKUS_LANGCHAIN4J_OPENAI_API_KEY` environment variable does not work, and the resulting error message does not tell the user how to fix the problem.

The improved message now suggests setting the key in `application.properties` or via the `-Dquarkus.langchain4j.openai.api-key=<value>` system property.

**Before:**
```
SRCFG00014: The config property quarkus.langchain4j.openai.api-key is required but it could not be found in any config source
```

**After:**
```
SRCFG00014: The config property quarkus.langchain4j.openai.api-key is required but it could not be found in any config source. Please configure it in application.properties or pass it as a system property: -Dquarkus.langchain4j.openai.api-key=<value>
```

This change is scoped to the OpenAI module. The same pattern exists in other recorders (Anthropic, Cohere, VertexAi, etc.) and can be improved in follow-up PRs.

## How verified

No existing tests assert on the exact error message text. The `OpenAiInvalidMaxRetriesTest` tests the separate `createMaxRetriesConfigProblems` method, which is not affected by this change.

Relates to #2750

### DIFF
--- a/model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/OpenAiRecorder.java
+++ b/model-providers/openai/openai-vanilla/runtime/src/main/java/io/quarkiverse/langchain4j/openai/runtime/OpenAiRecorder.java
@@ -823,9 +823,12 @@ public class OpenAiRecorder {
     }
 
     private ConfigValidationException.Problem createConfigProblem(String key, String configName) {
+        String propertyPath = "quarkus.langchain4j.openai"
+                + (NamedConfigUtil.isDefault(configName) ? "." : ("." + configName + ".")) + key;
         return new ConfigValidationException.Problem(String.format(
-                "SRCFG00014: The config property quarkus.langchain4j.openai%s%s is required but it could not be found in any config source",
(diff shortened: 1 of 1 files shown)


jmartisk:
Thanks for your PR!

However, TBH I'm not really sure about this; if we want to keep things unified and consistent, then we would have to update ALL error messages related to missing properties (in the whole Quarkus ecosystem!) in the same fashion. I think it's beyond the responsibility of these error messages to explain how basic configuration mechanisms in Quarkus work. 

Furthermore, `application.properties` and system properties are NOT the only way to pass configuration - there are many other ways (that's what the "in any config source" part of the message conveys - you can have arbitrary configuration sources that read configuration in different ways). So, if you wanted the error messages to explain the whole picture of how configuration can be declared, they would be ridiculously long.

geoand:
I agree with @jmartisk on this.

PR #2763, opened 2026-08-21
Omit temperature and top-p when they are not configured

### LINKED ISSUE #2762: [Bug] Need to remove WithDefault values to Optional Fields
When using  OpenAI's Responses API (POST /v1/responses) with gpt 5.4mini there is always an error returned regarding top_p and temperature fields not supported even though these fields are not initialized in our config files. 

The root cause was that these fields were changed from primitive Double to Optional<Double> in a recent upstream PR ([[#2370]](https://github.com/quarkiverse/quarkus-langchain4j/pull/2370)) specifically to let people omit them but the @WithDefault("1.0") annotations were left in place. An Optional with a @WithDefault is never actually empty when the property is absent; it always resolves to Optional.of(1.0).

This caused the api to always return an error:
Caused by: dev.langchain4j.exception.HttpException: {
  "error": {
    "message": "Unsupported parameter: 'top_p' is not supported with this model.",
    "type": "invalid_request_error",
    "param": "top_p",
    "code": null
Caused by: dev.langchain4j.exception.HttpException: {

Need help on this, so that we can use reasoning with our mini models.

omatheusmesmo:
## Describe your changes

The `@WithDefault` annotations were left in place when `temperature` and `topP` became `Optional` in #2370, so the `Optional` never resolves to empty and `1.0` is always sent. Models that reject these parameters, such as the reasoning models behind the Responses API, fail the request outright.

Azure applied a second hardcoded `0.7` fallback in `AzureOpenAiChatModel` and `AzureOpenAiStreamingChatModel`, which had to go as well for the empty `Optional` to reach the wire. This changes the effective default for anyone building those models directly without setting `temperature`, since `0.7` becomes the provider default. Through Quarkus config nothing changes, because the config always supplied a value before.

`google-genai` carries the same `@WithDefault`, copied when the extension was introduced. Gemini accepts the parameter so it never errored, but it silently overrode the model's own default. `topP` in that same config group is already optional without a default, so `temperature` is now consistent with it. That part is a separate commit and can be dropped independently.

The global `quarkus.langchain4j.temperature` property is unaffected: the default expression is kept, only its literal fallback is dropped.

Each changed path gets a test verified to fail without the fix, covering the blocking, streaming and Responses API models across the three providers. Tests setting the properties explicitly, and through the global property, pin down that those keep reaching the request.

Same direction as #2752 for Anthropic `top_k`.

---

- Closes #2762

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-all-config.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-all-config.adoc
@@ -4344,6 +4344,8 @@ endif::add-copy-button-to-config-props[]
 --
 What sampling temperature to use, with values between 0 and 2. Higher values means the model will take more risks. A value of 0.9 is good for more creative applications, while 0 (argmax sampling) is good for ones with a well-defined answer. It is recommended to alter this or topP, but not both.
 
+If not set, the parameter is omitted from the request and the model's own default applies.
+
 
 ifdef::add-copy-button-to-env-var[]
 Environment variable: env_var_with_copy_button:+++QUARKUS_LANGCHAIN4J_AZURE_OPENAI_CHAT_MODEL_TEMPERATURE+++[]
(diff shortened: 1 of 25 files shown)


geoand (approved):
I definitely agree with this change, but I believe it should be considered a breaking change.

cc @jmartisk

jmartisk (approved):

jmartisk:
> I definitely agree with this change, but I believe it should be considered a breaking change.

+1

PR #2802, opened 2026-09-02
Fix A2A agent discovery from Apicurio Registry always returning no agents

### LINKED ISSUE #2796: ApicurioAgentsRegistry#allAgents() always returns empty — missing .inputKeys(...) before .build()
### Describe the bug

`ApicurioAgentsRegistry#discoverAgents()` (added in #2501) builds each discovered agent like this:

```java
AgentInstance a2aAgent = (AgentInstance) AgenticServices.a2aBuilder(agentUrl)
        .outputKey(name)
        .build();
```

`AgenticServices.a2aBuilder(String)` returns an `A2AClientBuilder<UntypedAgent>`. `DefaultA2AClientBuilder#build()` throws:

```java
if (agentServiceClass == UntypedAgent.class && inputKeys == null) {
    throw new IllegalArgumentException("Input names must be provided for UntypedAgent.");
}
```

Since `.inputKeys(...)` is never called before `.build()`, this exception is thrown for every agent card, every time — even when the remote agent is perfectly reachable. The exception is swallowed by `discoverAgents()`'s own try/catch:

```java
} catch (Exception e) {
    log.warnf(e, "Failed to create A2A agent from artifact %s/%s",
            artifact.getGroupId(), artifact.getArtifactId());
}
```

so `ApicurioAgentsRegistry#allAgents()` / `#getAgent(String)` currently return **no agents at all**, regardless of whether the A2A servers behind the Agent Cards are reachable.

This is masked in the extension's own test suite: `ApicurioRegistryA2AToolsTest#allAgentsReturnsEmptyWhenA2AServersUnreachable` only proves `allAgents()` is empty when the servers are unreachable (`localhost:9999`, `localhost:9998`), so the `A2AClientError` thrown earlier in `DefaultA2AClientBuilder`'s constructor (fetching the agent card) is what's actually being caught in that test — it never exercises the `.build()` call with a reachable server, so the missing `.inputKeys(...)` call was never caught.

### Expected behavior

`ApicurioAgentsRegistry#allAgents()` should return a usable, invocable `AgentInstance`/`UntypedAgent` for every Agent Card whose URL is reachable.

### How to reproduce

1. Register an `AGENT_CARD` artifact in Apicurio Registry with the `a2a-agent-url` label pointing at a real, reachable A2A server.
2. Inject `ApicurioAgentsRegistry` and call `allAgents()`.
3. Observe the map is empty, and a `WARN` log `Failed to create A2A agent from artifact ...` with a stack trace containing `IllegalArgumentException: Input names must be provided for UntypedAgent.`

### Suggested fix

Add `.inputKeys("input")` (or a configurable equivalent) before `.build()` in `ApicurioAgentsRegistry#discoverAgents()`:

```java
AgentInstance a2aAgent = (AgentInstance) AgenticServices.a2aBuilder(agentUrl)
        .inputKeys("input")
        .outputKey(name)
        .build();
```

I found this while updating a demo project (https://github.com/carlesarnal/agent-discovery-demo) to use `ApicurioAgentsRegistry` for real A2A discovery/delegation instead of a hand-rolled implementation — happy to submit a PR with the one-line fix if that's preferred over someone else picking it up.

JeffersonSantos29:
### Problem

`ApicurioAgentsRegistry#allAgents()` always returns an empty map, regardless of
how many agent cards are published in the registry or whether the corresponding
A2A servers are reachable.

Agents are built through `AgenticServices.a2aBuilder(agentUrl)`, whose
single-argument overload resolves to `UntypedAgent`. `DefaultA2AClientBuilder#build()`
rejects an `UntypedAgent` that has no input keys:

```java
if (agentServiceClass == UntypedAgent.class && inputKeys == null) {
    throw new IllegalArgumentException("Input names must be provided for UntypedAgent.");
}
```

Since .inputKeys(...) was never called, every agent card threw, the exception
was swallowed by the surrounding catch and logged as a warning, and discovery
silently yielded nothing. getAgent(name) consequently always failed with
No agent found with name: ....

Why input keys are required

They are not an optional detail of the builder. For untyped agents they are used
in two places:

- DefaultA2AClientBuilder#invokeAgent reads them from the argument map to build
  the TextParts of the A2A message;
- A2AClientAgentInvoker#arguments turns them into the AgentArguments exposed
  to a supervisor or planner, i.e. they are the discovered agent's signature.

Solution

A2A agent cards carry no input schema, so the keys cannot be derived from the
card itself. They are resolved with a fallback chain:

1. the artifact's a2a-input-keys label (comma separated), symmetric with the
   existing a2a-agent-url and a2a-agent-skills labels written by
   A2AAgentCardPublisher;
2. the new quarkus.langchain4j.a2a.apicurio-registry.default-input-keys
   property, which also covers agents with more than one input;
3. input.

The existing single-argument ApicurioAgentsRegistry constructor keeps input
as its default, so direct instantiation is unaffected.

Note that input matching the @V("input") parameter of UntypedAgent#invoke
is a coincidence of naming: @V("input") is the whole argument map, while an
input key is a key within that map. A comment in the code says so, to keep the
two from being "simplified" into one later.

- Fixes:  #2796

### DIFF
--- a/a2a-apicurio-registry/deployment/src/test/java/io/quarkiverse/langchain4j/a2a/test/apicurio/ApicurioRegistryA2AToolsTest.java
+++ b/a2a-apicurio-registry/deployment/src/test/java/io/quarkiverse/langchain4j/a2a/test/apicurio/ApicurioRegistryA2AToolsTest.java
@@ -36,11 +36,12 @@ import io.quarkus.test.QuarkusUnitTest;
  * Integration test verifying that ApicurioAgentsRegistry can connect to
  * a real Apicurio Registry and search for AGENT_CARD artifacts.
  *
- * Note: allAgents() returns empty because the registered agent card URLs
- * (localhost:9999, localhost:9998) are not reachable — the A2A client builder
- * fails to fetch /.well-known/agent-card.json. This is expected in tests
- * without live A2A servers. The registry connectivity itself is verified
- * by querying the search API directly.
+ * Note: allAgents() returns empty here because the registered agent card URLs
+ * (localhost:9999, localhost:9998) are not reachable, so the A2A client builder
+ * fails to fetch /.well-known/agent-card.json before an agent can be built. That
+ * is the only reason left: the input keys the builder requires are resolved by
+ * ApicurioAgentsRegistryInputKeysTest in the runtime module. The registry
+ * connectivity itself is verified by querying the search API directly.
  */
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 public class ApicurioRegistryA2AToolsTest {
--- a/a2a-apicurio-registry/runtime/pom.xml
+++ b/a2a-apicurio-registry/runtime/pom.xml
@@ -46,6 +46,11 @@
             <artifactId>apicurio-registry-java-sdk-adapter-jdk</artifactId>
             <version>${apicurio-registry-sdk.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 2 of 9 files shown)


geoand:
cc @carlesarnal

carlesarnal (approved):
LGTM

geoand (approved):

PR #2803, opened 2026-09-02
fix:metrics  for error handling in MetricsChatModelListener

### LINKED ISSUE #2801: `MetricsChatModelListener.onError` puts the exception message in `error.type` and omits `gen_ai.provider.name`, which breaks the `gen_ai.client.operation.duration`
## Summary

`MetricsChatModelListener.onError` builds its tag set differently from `onResponse`, in two
ways that each cause a distinct production problem:

1. `error.type` is set to `error.getMessage()`, so it carries unbounded, per-request text.
2. the tag set omits `gen_ai.provider.name`, which `onResponse` includes, so the two paths
   register the same meter name with different tag keys.

## Where

`io/quarkiverse/langchain4j/runtime/listeners/MetricsChatModelListener.java`, lines 154-179
in 1.12.2:

```java
@Override
public void onError(ChatModelErrorContext errorContext) {
    ...
    String errorType = errorContext.error() != null
            ? errorContext.error().getMessage()      // (1)
            : "none";

    Tags tags = Tags.of("gen_ai.request.model", requestModel)
            .and("gen_ai.response.model", "none")
            .and("ai_service.class_name", "none")
            .and("ai_service.method_name", "none")
            .and("error.type", errorType);           // (2) no gen_ai.provider.name

    duration.withTags(tags).record(endTime - startTime, TimeUnit.NANOSECONDS);
}
```

`onResponse` (same file) builds:

```java
Tags tags = Tags.of("gen_ai.request.model", requestModel)
        .and("gen_ai.response.model", responseModel)
        .and("gen_ai.provider.name", responseContext.modelProvider()...)   // present here
        .and("ai_service.class_name", aiServiceClassName)
        .and("ai_service.method_name", aiServiceMethodName)
        .and("error.type", "none");
```

## Defect 1 — unbounded cardinality in `error.type`

The OpenTelemetry GenAI semantic conventions define `error.type` as a low-cardinality
attribute (the error class or a short code). The reference implementation in
`opentelemetry-instrumentation-api-incubator`
(`io.opentelemetry.instrumentation.api.incubator.semconv.genai.GenAiClientMetrics`) follows
that. Using `getMessage()` breaks it, because AWS SDK messages embed request-scoped values.

Actual tag values observed in Datadog for one service over 14 days:

```
the_provided_model_identifier_is_invalid._service:_bedrockruntime_status_code:_400_
  request_id:_c06b0882-2e48-4bcc-a20f-185c73b241a2_sdk_attempt_count:_1
the_provided_model_identifier_is_invalid._service:_bedrockruntime_status_code:_400_
  request_id:_c849b6e2-f976-48b3-b9b1-cf41b5e9c513_sdk_attempt_count:_1
_message_:_the_model_returned_the_following_errors:_prompt_is_too_long:_517709_tokens_
  200000_maximum
the_model_returned_the_following_errors:_thinking.type.enabled_is_not_supported_for_this_
  model._use_thinking.type.adaptive_and_output_config.effort_to_control_thinking_behavior...
```

The first two differ only by Bedrock request id: one time series per failed request. The
third embeds a token count. On a paid backend this is billed cardinality, and any dashboard
that groups by `error.type` renders one line per failure.

## Defect 2 — rejected registration on the error path

Prometheus permits one tag-key set per meter name. Because `onError` omits
`gen_ai.provider.name`, the second registration of `gen_ai_client_operation_duration_seconds`
is rejected. langchain4j's `ChatModelListenerUtils` swallows the exception, so the failure is
silent apart from a WARN:

```
WARN [de.la.mo.ch.ChatModelListenerUtils] An exception occurred during the invocation of the
chat model listener. This exception has been ignored.: java.lang.IllegalArgumentException:
Prometheus requires that all meters with the same name have the same set of tag keys. There
is already an existing meter named 'gen_ai_client_operation_duration_seconds' containing tag
keys [ai_service_class_name, ai_service_method_name, cluster, environment, error_type,
gen_ai_operation_name, gen_ai_provider_name, gen_ai_request_model, gen_ai_response_model,
host, kube_node, kube_replica_set, pod_name, region, role]. The meter you are attempting to
register has keys [ai_service_class_name, ai_service_method_name, cluster, environment,
error_type, gen_ai_operation_name, gen_ai_request_model, gen_ai_response_model, host,
kube_node, kube_replica_set, pod_name, region, role].
```

The key sets differ by exactly `gen_ai_provider_name`. Which path loses depends on which
registers first in a given JVM, so behaviour is inconsistent across pods of the same
deployment. Note the same class also registers `gen_ai.client.token.usage` and
`gen_ai.client.estimated_cost`, and an application listener that follows `onResponse`'s key
set (the documented way to add provider-specific token metrics) collides with `onError`.

## Impact

Latency and error counts are lost for exactly the calls that failed — the ones an operator
most wants to see — and the cardinality of `error.type` grows with traffic.

## Expected

`onError` should build the same tag keys as `onResponse`, and `error.type` should be
low-cardinality.

## Suggested fix

```java
String errorType = errorContext.error() != null
        ? errorContext.error().getClass().getSimpleName()
        : "none";

Tags tags = Tags.of("gen_ai.request.model", requestModel)
        .and("gen_ai.response.model", "none")
        .and("gen_ai.provider.name", providerName(errorContext))   // added
        .and("ai_service.class_name", "none")
        .and("ai_service.method_name", "none")
        .and("error.type", errorType);
```

`ChatModelErrorContext` exposes `modelProvider()`, so the provider tag can be filled the same
way `onResponse` does, falling back to `"none"`.

## Workaround

A `MeterFilter` can repair both locally, since filters run before registration:

```java
MeterFilter.forMeters(
    id -> "gen_ai.client.operation.duration".equals(id.getName()),
    MeterFilter.replaceTagValues("error.type", message -> "error", "none"));
```

plus a `map` that adds `gen_ai.provider.name` when absent. The value-collapsing half is what
we shipped; the key-repair half we skipped, because our Datadog registry accepts the
mismatched key set and only the Prometheus endpoint is affected.

matheusandre1:
Fixes: #2801 

The `MetricsChatModelListener`'s `onError` method recorded `error.type` using the exception's `getMessage()`, resulting in unbounded cardinality and omitting the `gen_ai.provider.name` tag included by `onResponse`. Since Prometheus requires a single set of keys per metric name, the second attempt to register `gen_ai.client.operation.duration` was silently rejected.

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/cost/OperationDurationErrorMetricsTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/cost/OperationDurationErrorMetricsTest.java
@@ -1,0 +1,125 @@
+package io.quarkiverse.langchain4j.test.cost;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import jakarta.inject.Inject;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.data.message.UserMessage;
+import dev.langchain4j.model.ModelProvider;
+import dev.langchain4j.model.chat.listener.ChatModelErrorContext;
+import dev.langchain4j.model.chat.listener.ChatModelRequestContext;
+import dev.langchain4j.model.chat.listener.ChatModelResponseContext;
+import dev.langchain4j.model.chat.request.ChatRequest;
+import dev.langchain4j.model.chat.request.DefaultChatRequestParameters;
+import dev.langchain4j.model.chat.response.ChatResponse;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import io.quarkiverse.langchain4j.runtime.listeners.MetricsChatModelListener;
+import io.quarkus.test.QuarkusUnitTest;
+
+/**
+ * Verifies that {@link MetricsChatModelListener#onError} records {@code gen_ai.client.operation.duration} with a
+ * low-cardinality {@code error.type} and with exactly the same tag keys as
+ * {@link MetricsChatModelListener#onResponse}, so registries that require a single tag-key set per meter name
+ * (such as Prometheus) accept both the success and the error path.
+ */
+class OperationDurationErrorMetricsTest {
+
+    private static final String DURATION_METER = "gen_ai.client.operation.duration";
+
+    @RegisterExtension
+    static final QuarkusUnitTest unitTest = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class));
+
+    @Inject
+    MetricsChatModelListener listener;
+
+    @Inject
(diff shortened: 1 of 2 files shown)


omatheusmesmo (approved):
LGTM!

jmartisk (approved):

PR #2805, opened 2026-09-02
Return String tool results as-is instead of JSON-encoding them

### LINKED ISSUE #2799: `QuarkusToolExecutor` JSON-encodes a `String` tool result, where the `DefaultToolExecutor` it replaces returns it as-is
**Extension version:** 1.13.1
**Quarkus version:** 3.39.1
**LangChain4j version:** 1.19.0

A `@Tool` method returning `String` delivers its text to the model JSON-quoted with `\n` escaped, instead of raw. `QuarkusAiServicesFactory$QuarkusAiServices` overrides `tools(Collection<Object>)` — which `tools(Object...)` delegates to — and installs `QuarkusToolExecutor` in place of langchain4j's `DefaultToolExecutor`. The two disagree on how a result becomes text:

| return type | `DefaultToolExecutor.toText` | `QuarkusToolExecutor.handleResult` |
|---|---|---|
| `void` | `"Success"` | `"Success"` |
| `String` | **returned as-is** | `Json.toJson` |
| anything else | `Json.toJson` | `Json.toJson` |

`QuarkusToolExecutor.handleResult` has no `returnType == String.class` branch, so the encoding is unconditional. Because `AiServices.builder(…)` resolves through the `dev.langchain4j.spi.services.AiServicesFactory` SPI, this applies to every AI service inside Quarkus, including hand-built ones.

Still an unconditional `Json.toJson` on `main` (`QuarkusToolExecutor.java:147-152`).

### Reproducer

A tool returning markdown, registered through the object overload (`@RegisterAiService(tools = DocTool.class)`, or `AiServices.builder(…).tools(new DocTool())`):

```java
static class DocTool {
    @Tool("Reads a document")
    String readSource() {
        return "# Title\nKind: file\n\nBody.";
    }
}
```

### Actual behaviour

The `ToolExecutionResultMessage` the model receives carries:

```
"# Title\nKind: file\n\nBody."
```

### Expected behaviour

The raw text, as `DefaultToolExecutor` returns it for a `String` result:

```
# Title
Kind: file

Body.
```

Passing the same tool as `tools(Map<ToolSpecification, ToolExecutor>)` — an overload the extension does not override — produces the expected text, which isolates the difference to the executor.

omatheusmesmo:
`QuarkusToolExecutor.handleResult` applied `Json.toJson` unconditionally, so a `@Tool` method returning `String` reached the model JSON-quoted with escaped newlines. `DefaultToolExecutor`, which this executor replaces through the `AiServicesFactory` SPI, returns a `String` result unchanged, so the text differed for every AI service built inside Quarkus.

A `String` result now passes through untouched. `Uni<String>` is covered as well, since it is resolved to its item before the result is converted. The check is `instanceof` rather than the declared return type, both to stay reflection-free and because a `Uni<String>` tool would otherwise still be encoded.

Existing tests that asserted the JSON-quoted form are updated.

- Closes: #2799

### DIFF
--- a/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ToolExecutorTest.java
+++ b/core/deployment/src/test/java/io/quarkiverse/langchain4j/test/ToolExecutorTest.java
@@ -28,6 +28,7 @@ import io.quarkiverse.langchain4j.runtime.ToolsRecorder;
 import io.quarkiverse.langchain4j.runtime.tool.QuarkusToolExecutor;
 import io.quarkiverse.langchain4j.runtime.tool.ToolMethodCreateInfo;
 import io.quarkus.test.QuarkusUnitTest;
+import io.smallrye.mutiny.Uni;
 
 class ToolExecutorTest {
 
@@ -141,6 +142,30 @@ class ToolExecutorTest {
         @Tool
         String mappedEnum(Map<String, SortBy> arg0) {
             return arg0.get("first").name();
+        }
+
+        @Tool
+        String markdownDocument() {
+            return "# Title\nKind: file\n\nBody.";
+        }
+
+        @Tool
+        String nullDocument() {
+            return null;
+        }
+
+        @Tool
+        Uni<String> markdownDocumentUni() {
+            return Uni.createFrom().item("# Title\nKind: file\n\nBody.");
+        }
+
+        @Tool
+        ArgObject objectResult() {
+            return new ArgObject("ABC", SortBy.DATE);
+        }
+
+        @Tool
+        void voidResult() {
         }
     }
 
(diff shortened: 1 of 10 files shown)


geoand:
The test failures seem related

omatheusmesmo:
> The test failures seem related

I am so sorry @geoand , I missed the CI results, now it's fixed, thanks for letting me know.

geoand:
Thanks. 

Is this behavior in line with what upstream LangChain4j does?

omatheusmesmo:
Yes, upstream does the same in `DefaultToolExecutor.toText()`: https://github.com/langchain4j/langchain4j/blob/cbbd8e2ec79863710301f179a07e8e7c4a181c2b/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java#L305-L316

PR #2809, opened 2026-09-04
Bump LangChain4j to 1.20.0 version

mariofusco:


geoand:
cc @andreadimaio

geoand (approved):

geoand:
Let's get the blessing from @jmartisk before merging

mariofusco:
If it is ok for you, can we please have this merged and if possible also have another patch release of the extension with this bump? I pushed in LangChain4j a fix for the `AgenticScope` serialization that is required by QuarkusFlow integration and without it we are also blocked on moving the workshop forward.

geoand:
@jmartisk can you PTAL?

As for the release, we can of course do one but it will have to be a new minor version as the existing release will be going into Quarkus 3.40 (LTS) so we can't bump LangChain4j in it

jmartisk (approved):

PR #2800, opened 2026-09-02
fix(anthropic): default client timeout when unset

CryoThrust:
`QuarkusAnthropicClient` dereferenced `builder.timeout` unconditionally, so programmatically built Anthropic models failed with a `NullPointerException` when no timeout was configured. The LangChain4j Anthropic client contract defines a 15-second default when no timeout is provided; this change applies the same fallback before configuring the Quarkus REST client.

A focused runtime test verifies that the client builder can be constructed without an explicit timeout.

Local Maven verification is currently blocked because the configured internal repository resets while resolving the parent artifact `io.quarkiverse:quarkiverse-parent:21`.

- Fixes: #2798.

PR #2822, opened 2026-09-08
Fix NPE when Anthropic client timeout is not configured

### LINKED ISSUE #2798: `QuarkusAnthropicClient` NPEs when the builder sets no `.timeout(…)`
**Extension version:** 1.13.1
**Quarkus version:** 3.39.1
**LangChain4j version:** 1.19.0
**Provider:** Anthropic (`quarkus-langchain4j-anthropic`)

`QuarkusAnthropicClient` dereferences `builder.timeout` without a null check, so an `AnthropicChatModel` or `AnthropicStreamingChatModel` built in code without `.timeout(…)` throws at construction. The same builder code works against plain `langchain4j-anthropic`, whose `DefaultAnthropicClient` treats the timeout as optional: `getOrDefault(getOrDefault(builder.timeout, httpClientBuilder.connectTimeout()), Duration.ofSeconds(15))`.

Models built from config are unaffected, because `AnthropicRecorder` passes `anthropicConfig.timeout().orElse(Duration.ofSeconds(10))` — this only hits a programmatically built model.

### Reproducer

Any Quarkus application with `quarkus-langchain4j-anthropic` on the classpath — the extension installs its client through the `AnthropicClientBuilderFactory` SPI, so no API key and no network call are involved:

```java
AnthropicChatModel.builder()
        .apiKey("dummy")
        .modelName("claude-sonnet-4-5")
        .build();
```

### Actual behaviour

```
java.lang.NullPointerException: Cannot invoke "java.time.Duration.toSeconds()" because "builder.timeout" is null
        at io.quarkiverse.langchain4j.anthropic.QuarkusAnthropicClient.<init>(QuarkusAnthropicClient.java:79)
```

### Expected behaviour

The client is constructed, with the timeout defaulted when unset, as `DefaultAnthropicClient` does.

wzj1228516103:
## What does this PR do?

Fixes a NullPointerException in `QuarkusAnthropicClient` when the builder does not configure a timeout. The client now uses a 15-second default timeout, while preserving explicitly configured timeout values.

## Why is this needed?

`AnthropicClient.Builder.timeout` is optional. The Quarkus client previously dereferenced it unconditionally while configuring the REST client.

## How was this tested?

- `QuarkusAnthropicClientTest` verifies the default and explicit timeout paths.
- `AnthropicCacheTokenUsageExtractorTest` continues to pass.

Fixes #2798

### DIFF
--- a/model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/QuarkusAnthropicClient.java
+++ b/model-providers/anthropic/runtime/src/main/java/io/quarkiverse/langchain4j/anthropic/QuarkusAnthropicClient.java
@@ -13,6 +13,7 @@ import static java.util.stream.StreamSupport.stream;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


geoand:
We actually already have https://github.com/quarkiverse/quarkus-langchain4j/pull/2800 which fixes the issue.

PR #2844, opened 2026-09-10
feat: Add Apicurio Registry integration for prompt templates

### LINKED ISSUE #2658: feat: Apicurio Registry integration for prompt template governance
## Context

With the Apicurio Registry MCP discovery integration (`mcp-apicurio-registry` module, commit b33ff14), Quarkus LangChain4j can already discover and connect to MCP servers at runtime via the registry. This issue proposes extending that integration to cover **prompt template governance** — the third pillar alongside agent discovery (A2A) and tool discovery (MCP).

Since Apicurio Registry 3.1, the registry supports a first-class `PROMPT_TEMPLATE` artifact type (aligned with the Microsoft Prompty spec) with versioning, variable validation, compatibility rules, and MCP prompt exposure.

## Proposal

Allow `@SystemMessage` / `@UserMessage` to load prompt templates directly from Apicurio Registry at runtime, with governance checks enforced by the registry:

```java
@AiService
public interface Assistant {
    @SystemMessage(fromRegistry = "customer-support-prompt-v2")
    String chat(String userMessage);
}
```

### Key capabilities

- **Runtime resolution**: resolve prompt templates from Apicurio Registry by artifact ID (and optionally version/group), with caching and fallback behavior
- **Variable validation**: the registry's `PROMPT_TEMPLATE` schema defines expected variables — mismatches between the template and the AI service method parameters can be caught at build time or startup
- **Compatibility enforcement**: if an ML engineer publishes a breaking prompt change (e.g. renames a variable), the registry's backward compatibility rules reject it before it reaches the Quarkus application
- **Versioning**: pin to a specific version, a branch, or follow `latest` — same semantics as other Apicurio artifact references
- **Build-time validation** (optional): a Quarkus build step could pull the template and verify variable bindings, failing the build on incompatibilities

### Configuration sketch

```properties
quarkus.langchain4j.prompt.apicurio-registry.url=http://localhost:8080/apis/registry/v3
quarkus.langchain4j.prompt.apicurio-registry.cache-ttl=60s
quarkus.langchain4j.prompt.apicurio-registry.compatibility-check=true
```

## Why this matters

Without centralized prompt governance, prompt changes are a common source of silent runtime failures in AI applications — an ML engineer modifies a template variable, and downstream agents break with no clear signal. The registry's compatibility checks prevent this the same way schema registries prevent breaking API changes.

## Prior art

- `mcp-apicurio-registry` module — same pattern of registry-backed discovery, applied to MCP servers
- `a2a-apicurio-registry` — agent card discovery via the registry
- Apicurio Registry `PROMPT_TEMPLATE` artifact type — [docs](https://www.apicur.io/registry/docs/apicurio-registry/3.2.x/getting-started/assembly-llm-artifact-types-implementation.html)

matheusandre1:
Adds prompt template governance backed by [Apicurio Registry](https://www.apicur.io/registry/), the third pillar next to A2A agent discovery (`a2a-apicurio-registry`) and MCP tool discovery (`mcp-apicurio-registry`).

AI services can now load their system/user message templates from `PROMPT_TEMPLATE` artifacts at runtime:

```java
@RegisterAiService
public interface Assistant {

    @SystemMessageFromRegistry("customer-support")
    String chat(@V("product") String product, @UserMessage String question);

    @SystemMessageFromRegistry(value = "customer-support", groupId = "support-team", version = "2.1.0")
    String chatPinned(@V("product") String product, @UserMessage String question);

    @UserMessageFromRegistry("summarize")
    String summarize(String text);
}
```

```properties
quarkus.langchain4j.prompt.apicurio-registry.url=http://localhost:8080/apis/registry/v3
quarkus.langchain4j.prompt.apicurio-registry.cache-ttl=60s
quarkus.langchain4j.prompt.apicurio-registry.validate-on-startup=true
```

> `@SystemMessage` / `@UserMessage` are LangChain4j upstream annotations, so a `fromRegistry` attribute (as sketched in the issue) cannot be added to them. Two Quarkus-side annotations, `@SystemMessageFromRegistry` and `@UserMessageFromRegistry`, are introduced instead, mirroring `fromResource`.

### Core (`core/runtime`, `core/deployment`)

- New annotations `io.quarkiverse.langchain4j.SystemMessageFromRegistry` (method or interface) and `UserMessageFromRegistry` (method), with `value` (artifact id), optional `groupId` and `version`.
- Registry-agnostic SPI in `io.quarkiverse.langchain4j.prompt`: `PromptTemplateRegistry`, `PromptTemplateReference`, `ResolvedPromptTemplate`, `PromptVariable`, `PromptTemplateUsage`, plus `PromptTemplateResolutionException` / `PromptTemplateValidationException`.
- `AiServiceMethodCreateInfo.TemplateInfo` carries an optional registry reference. At runtime, `PromptTemplateRegistrySupport` resolves the template through the `PromptTemplateRegistry` bean, applies declared default values, and fails with a clear message when a required variable is not bound by the method.
- Build-time checks: combining `@SystemMessage` with `@SystemMessageFromRegistry` (or the user message equivalents) on the same element is rejected. Method-level annotations take precedence over class-level ones, as today.
- New `PromptTemplateUsageBuildItem` exposes every registry-backed message (reference, bound variable names, location) so registry extensions can validate eagerly.

### New extension `prompt-apicurio-registry` (`quarkus-langchain4j-prompt-apicurio-registry`)

- `ApicurioPromptTemplateRegistry`, registered as a synthetic bean implementing `PromptTemplateRegistry`, using the Apicurio Java SDK (JDK HTTP adapter, like `a2a-apicurio-registry`).
- **Runtime resolution with caching**: stale-while-revalidate. Once `cache-ttl` expires the cached template keeps being served while a refresh runs in the background, so AI service invocations are never blocked by the registry after the first resolution. Failed refreshes keep the last good version and log a warning. `cache-ttl=0` fetches once.
- **Variable validation**: required variables declared by the artifact must be bound by the method (parameter names, `@V`, or the implicit `chat_memory` / `response_schema`). Declared defaults are applied for unbound variables.
- **Startup validation** (`validate-on-startup=true`, default): every referenced template is fetched and validated when the application starts; the app fails fast on unresolvable templates or unbound required variables. With `false`, resolution is lazy and the same checks run on first use.
- **Versioning**: `version` accepts any Apicurio version expression (`1.0.0`, `branch=latest`); `latest` is an alias for `branch=latest`. Defaults come from `default-group` / `default-version`.
- **Content formats**: JSON and YAML documents (`template` + `variables` / Prompty `inputs`), Prompty-style front matter + body (`text/x-prompt-template`), and plain text. Variables may be declared as a name-to-schema map or as a JSON Schema object.
- Authentication: HTTP basic (`basic-auth.*`) or OAuth2 client credentials (`oauth2.*`).
- `ApicurioPromptTemplateRegistryBuilder` for programmatic use; `invalidate()` / `refresh()` to bypass the cache.
- Compatibility enforcement is left to the registry's own `COMPATIBILITY` rules (server side); the docs show how to enable them so breaking prompt changes are rejected before they reach the application.

- Since the template content is only known at runtime, methods with more than one parameter must annotate the user message parameter with `@UserMessage` (the build-time inference cannot see the template variables). Documented this

----

- Resolves: #2658


matheusandre1:
cc: @carlesarnal

jmartisk:
I'm starting to think that these apicurio-related modules should be in an extra directory under the root, now that there will be 3 of them :)

geoand:
💯

carlesarnal:
> I'm starting to think that these apicurio-related modules should be in an extra directory under the root, now that there will be 3 of them :)

Looks like a good idea to me. Generally speaking we're expanding our scope to have a registry of AI artifacts, for now, that's agents/MCPs/prompts, ideally, they will converge at some point and we can have just one but, for now, it does not seem feasible.

geoand (commented):
Thanks for this.

I need to both think about the implications of this and go through its implementation

geoand:
I actually think this is not the right approach because we already have a similar (but different) mechanism in `SystemMessageProvider`

matheusandre1:
> I actually think this is not the right approach because we already have a similar (but different) mechanism in `SystemMessageProvider`

Right, I understand. You want to keep the core intact and build something based on the existing provider?

geoand:
I would like us to come up with some unified (and pluggable) way to achieve this

jmartisk:
Well this takes us back to the (unresolved) Context Engineering  discussion (https://github.com/quarkiverse/quarkus-langchain4j/discussions/2060) which suggests some unified approaches to providing system messages and Apicurio could be just a specific context provider if we implemented it. But we have never made too much progress with it.

jmartisk:
I don't expect that to be done very soon though, so... I'm afraid we may need to go with the `@SystemMessageFromRegistry` or something like that

geoand:
Dunno, I don't want to make things harder for us when we eventtually do get around to having a context engineering story

jmartisk:
Yeah I very much agree, but on the other hand, I don't want to considerably delay a new feature because of that :/ 
But if we want to avoid introducing a completely new mechanism, then yeah, let's somehow wrap it around [SystemMessageProvider](https://github.com/quarkiverse/quarkus-langchain4j/blob/main/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/SystemMessageProvider.java)

geoand:
> I don't want to considerably delay a new feature because of that :/

Sure, but is this something that is really that important?

Issue #2873, opened 2026-09-18
quarkus-agent-mcp's quarkus_searchDocs returns no results for quarkus-langchain4j


kdubois:
When using the Quarkus Agent MCP (https://github.com/quarkusio/quarkus-agent-mcp) server, the quarkus_searchDocs tool is unable to find any documentation for quarkus-langchain4j concepts.  

Developers using the Quarkus AI coding assistant get helpful, context-aware answers for core Quarkus features but hit a dead end the moment they ask about anything specific to this extension: AI services, model providers, RAG pipelines, MCP integration, guardrails, etc. This is precisely the area where documentation lookup matters most, given how much configuration and concept knowledge is involved.

geoand:
cc @mariofusco

PR #2875, opened 2026-09-19
Add separate RAG artifact for documentation search

omatheusmesmo:
Supersedes #2629, rebased onto current `main`, with one change on top: `maven_args: -Prag` in the release workflow. The original commit is preserved with @phillip-kruger as author, and `docs-rag/pom.xml` is byte-identical to his.

The Quarkiverse shared release workflow deploys with `mvn -B deploy -DperformRelease ...` and activates no RAG profile on its own, so `skipRagGeneration` stayed `true` and the published artifact carried no `META-INF/quarkus-rag.sql`. 11 of the 12 RAG artifacts currently on Maven Central are empty 2 KB jars for this reason. `quarkus-json-rpc` is the exception, and it passes `maven_args: -Prag` exactly like this.

Verified locally with the profile the workflow now passes:

```
./mvnw -pl docs-rag package -Prag
→ Wrote 971 chunks to .../META-INF/quarkus-rag.sql (5.5 MB)
```

Discovery for a typical application needs one more thing: `RagSqlLoader` resolves direct dependencies only, and applications declare a model provider rather than `quarkus-langchain4j-core`, which is where the pointer file lives. So the artifact would be published and then not found. That is reported with a reproducer in quarkusio/quarkus-agent-mcp#280 and fixed in quarkusio/quarkus-agent-mcp#281, which switches that scan to the transitive overload already used by `SkillReader` there.

`quarkus-json-rpc` does not hit this because it is a single-module extension, so the artifact carrying its pointer is the one applications declare. If quarkusio/quarkus-agent-mcp#281 is not accepted, the alternative is shipping the pointer file in more deployment modules here.

Relates to #2873.


phillip-kruger:
Thanks for picking this up and rebasing it. The diagnosis is right: the shared workflow only passes `-DperformRelease`, so the `release` profile never runs and the jar goes out empty. I'd like two changes before we merge:

**1. `-Prag` will clash with quarkiverse-parent 22.** We're on parent 21. Parent 22 is already on Central and adds its own `rag` profile, which runs `generate-rag` (execution id `generate-extension-rag`) in every module that inherits the parent. Once Renovate bumps the parent, `-Prag` will turn that profile on for the whole reactor, not just `docs-rag`. Every extension module would run the plugin with default settings, and `docs-rag` would run it twice. There are two ways to fix it:

- Rename this module's profile, e.g. `<id>docs-rag</id>`, and pass `maven_args: -Pdocs-rag`, or
- (my preference) move to parent 22 in this PR and use the parent's mechanism:
  - set `guidesDirectory` on the `generate-extension-rag` execution in `docs-rag/pom.xml`
  - drop the hard-coded `0.0.7` so the version comes from the parent's pluginManagement
  - pass `-Drag` in the workflow

  That keeps us in line with the other Quarkiverse repos.

**2. Please remove the `release` profile from `docs-rag/pom.xml`.** It never runs on the real release path. It makes the pom look as if releases already produce the RAG data, which is how this went unnoticed.

About discovery: agreed that this only helps once quarkusio/quarkus-agent-mcp#281 lands, since the pointer file lives in `core/deployment`. That's fine to merge in either order. If #281 doesn't go in, we can talk about putting the pointer file in more modules then.

I'll close #2629 once this is in.

omatheusmesmo:
Hi @phillip-kruger , thanks for reviewing, I addressed the issues mentioned by you. Please review again whenever you have some time! :smile:

phillip-kruger:
Thanks @omatheusmesmo, both points are addressed and this looks good to me. quarkusio/quarkus-agent-mcp#281 is merged too, so discovery through the model-provider dependencies works now.

One follow-up that doesn't need anything from this PR: parent 22 pins `quarkus-documentation-rag-maven-plugin` to 0.0.5, not 0.0.7, so for now the generated data will have the old `quarkus_version` metadata key and `quarkus.io/guides` URLs. `RagSqlLoader` already handles both. I've opened quarkiverse/quarkiverse-parent#262 to release parent 23 with 0.0.7. The version isn't pinned here anymore, so Renovate's bump to 23 will pick it up.

PR #2876, opened 2026-09-21
Make @CdiBean implicit when a CDI qualifier is present on a supplier parameter

mariofusco:
Currently `@ChatModelSupplier` methods that need a CDI-managed model must use both `@CdiBean` and `@ModelName`

In this case the `@CdiBean` annotation is redundant here since `@ModelName` is already a @Qualifier-annotated CDI annotation, so the user's intent is unambiguous. 

 This pull request allows to avoid the `@CdiBean` annotation in this cases, and applies to any CDI qualifier (not just `@ModelName`), so it also benefits `@McpClientName` and any custom @Qualifier-annotated annotation.


geoand (approved):

mariofusco:
It seems that all the build failures are not related with my change, but something caused by a problem while downloading the latest JBang release.

